Issue #47, opened 2021-11-02
Postgres Source fails when handling bytea primary keys


dylanlott:
# Steps to Reproduce
- Have a Postgres Source configured to read a table that has a primary key column of type `bytea`.
- Configure a pipeline to run that plugin.

# Expected Behavior
The Postgres Source should handle this functionality in the same way it would handle any other key column.

# Actual Behavior
The Postgres connector cannot handle this query correctly and instead of returning the correct result or a descriptive error it returns an `ErrEndData` which implies that the plugin operated correctly and is now at its end.

neovintage:
This should be fixed. @uchennakevinm1 can you link to the appropriate PR / Issue where we did this.

uchennakevinm1:
@neovintage https://github.com/ConduitIO/conduit-connector-postgres/pull/14

PR #72, opened 2022-01-13
Makes snapshot default behavior and refactors snapshot transition to cdc mode [1]

### LINKED ISSUE #49: Postgres source: records duplicated in certain cases
If the last saved position in a Pg source is _M_, and a newly inserted row has ID _N_, then the newly created record will be returned _N-M_ times by the source. For example, if the last row returned by a Pg source had ID 10, and then we insert a row with the ID 30, then the new row will be returned 20 times.

**Build** 0ec0aa6

**Steps to reproduce**
1. Setup a table in Pg, with one row, where ID is 1 (for example):
```
insert into persons (firstname, personid) values ('foo', 100);
```
2. Create a pipeline with a Pg source and any destination (can be a file destination). Pg source configuration I used:
```
"settings": {
	"table": "persons",
	"url": "postgresql://meroxauser:meroxapass@localhost/conduit-test-db?sslmode=disable"
}
```
3. Run the pipeline.
4. Expected (and actual behavior): you see following in the file destination:
```
{"firstname":"foo","personid":100}
```
5. Insert following row into Pg:
```
insert into persons (address, city, firstname, lastname, personid) values ('wall street', 'new york', 'foo', 'bar', 105);
```
Run the pipeline again.
6. Expected: you see following in the file destination:
```
{"firstname":"foo","personid":105}
```
Actual behavior:
```
{"firstname":"foo","personid":105}
{"firstname":"foo","personid":105}
{"firstname":"foo","personid":105}
{"firstname":"foo","personid":105}
{"firstname":"foo","personid":105}
```

dylanlott:
### Description

This PR makes it so that the Postgres plugin attempts a snapshot by default on start and then switches over to CDC mode if it's configured. If not, it will simply return `ErrEndData` as all records have thus been read out.

It is important to note that this PR makes it so that Read's `Position` is not respected, as both Snapshot and CDC mode do not need it to return their next value. Position should eventually be set at `Open` instead and respected by the CDC iterator.

Fixes #49 & #47

### Quick checks:


pkg/plugins/pg/source/source.go
```diff
@@ -166,74 +165,25 @@ func (s *Source) Read(ctx context.Context, pos record.Position) (record.Record,
...
+	// check cdc buffer
+	if s.cdc != nil {
+		if s.cdc.HasNext() {
+			next, err := s.cdc.Next()
+			return next, err
 		}
-		return record.Record{}, cerrors.Errorf("failed to build payload: %w", err)
 	}
-	return rec, nil
+	// TODO: implement long polling reader here
```
dylanlott, line 178: For postgres users who can't uses CDC due to lack of configuration options or version issues, we can support a long polling approach to table snapshotting.

pkg/plugins/pg/source/cdc.go
```diff
@@ -132,6 +147,8 @@ func (s *Source) withCDC(ctx context.Context, cfg plugins.Config) error {
 	go func() {
```
hariso, line 147: I noticed one thing in the code which has been merged previously:
```go
if err != nil {
	if !strings.Contains(err.Error(), "SQLSTATE 42710") {
		return cerrors.Errorf("failed to create publication %s: %w", slotName, err)
	}
	log.Printf("publication %s already exists; continuing to slot setup", publication)
}
```
My understanding is that we want all DML operations to be published by this this publication. However, if the publication already exists, it _may_ be set up such that only some DML operations are being published. E.g. the publication might already exist, but only `insert`s are published.

pkg/plugins/pg/source/cdc.go
```diff
@@ -28,9 +28,24 @@ import (
...
 // error.
-func (s *Source) withCDC(ctx context.Context, cfg plugins.Config) error {
+func (s *Source) withCDC(cfg plugins.Config) error {
+	c := context.Background()
+	ctx, cancel := context.WithCancel(c)
+	s.killswitch = cancel
+	// early return if cdc is disabled
+	v, ok := cfg.Settings["cdc"]
+	// if cdc is set, check the value for falsy values.
+	if ok {
```
hariso, line 42: i.e. the default behavior is that it is turned on?

pkg/plugins/pg/source/cdc.go
```diff
@@ -28,9 +28,24 @@ import (
...
+		switch v {
+		case "disabled", "false", "off", "0":
```
hariso, line 44: Do we care about case sensitivity here?

pkg/plugins/pg/source/snapshot.go
```diff
@@ -149,53 +149,47 @@ func (s *Snapshotter) Teardown() error {
...
-		Isolation: sql.LevelReadCommitted,
-	})
-	if err != nil {
-		return cerrors.Errorf("failed to start transaction: %w", err)
-	}
+// loadRows loads the rows returned from the database onto the snapshotter
+// or returns an error.
+// * It returns nil if no error was detected.
+// * rows.Close and rows.Err are called at Teardown.
+func (s *Snapshotter) loadRows(db *sql.DB) error {
```
hariso, line 159: Nitpick: `db` unused.

pkg/plugins/pg/source/source.go
```diff
@@ -40,20 +39,17 @@ var (
 	// ErrInvalidURL is returned when the DB can't be connected to with the
 	// provided URL
 	ErrInvalidURL = cerrors.New("incorrect url")
+	// ErrEmptyConfig is returned when no config is provided
+	ErrEmptyConfig = cerrors.Errorf("must provide a plugin config")
```
hariso, line 43: Nitpick, with a disclaimer of me not being a native English speaker :D When I hear "empty config" I think of a config that actually exists but has nothing in it. The comment and the error text though indicate there's no config at all.

hariso (commented):
Thanks for the PR Dylan! I left a few comments, mostly nitpick, to buy me some time for another pass.:D

pkg/plugins/pg/source/source.go
```diff
@@ -68,17 +64,19 @@ type Source struct {
 	sub *pgoutput.Subscription
 	// subErr is the return value of the `sub.Start` function in withCDC
 	subErr error
-	// buffer holds a reference to our CDCIterator so that we can read
-	// from the buffer when rows are no longer available.
-	buffer *CDCIterator
+	// cdc holds a reference to our CDCIterator so that we can read
+	// from the cdc when rows are no longer available.
+	cdc Iterator
```
hariso, line 69: If this should always be a CDCIterator, should we declare it as such?

pkg/plugins/pg/source/source.go
```diff
@@ -98,60 +96,61 @@ func (s *Source) Open(ctx context.Context, cfg plugins.Config) error {
...
-// * Read takes the _current_ position of the connector, and returns the row
-// at the _next_ position.
-func (s *Source) Read(ctx context.Context, pos record.Position) (record.Record, error) {
+// Read takes a context and a Position and returns a Record or an error.
+// * This currently doesn't respect which Position it is fed.
+// * Instead Read will send the next record in the Snapshot over until the
+// initial table snapshot is finished.
+// * Once the table snapshot is finished, it checks for a CDC buffer and if
+// one exists it reads off that buffer.
+// * If not CDC buffer exists it will return ErrEndData and switch to long
```
hariso, line 150: Nitpick: If not CDC buffer exists -> If no CDC buffer exists.

pkg/plugins/pg/source/cdc.go
dylanlott: We probably do, but I don't know at what level to enforce that at. I would think we should address that in a different PR.

pkg/plugins/pg/source/cdc.go
```diff
@@ -28,9 +28,24 @@ import (
 	"github.com/jackc/pgx"
 )
 
+var bufferSize = 1000
+
 // withCDC sets up change data capture for the Postgres Source or returns an
 // error.
-func (s *Source) withCDC(ctx context.Context, cfg plugins.Config) error {
+func (s *Source) withCDC(cfg plugins.Config) error {
+	c := context.Background()
```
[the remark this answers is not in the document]
dylanlott: Good catch, I'll remove that.

pkg/plugins/pg/source/cdc.go
dylanlott: This is probably something we need to point out in the documentation. It can also exist if we've created it before, which is why we have to detect that error manually like this. I agree, though, that could be a foot-gun to the end user that we can't really stop. Our best advice is probably to recommend a unique name for the connector's slot in all cases unless you specifically know what you're doing reusing an established slot.

hariso: So, I haven't used Pg publications before, but if we start using an existing publication, will we get all operations from it, or just those from the moment we subscribe to it?

One thing we can do, to ensure that the publication we use is publishing all the operations, is to read its metadata. If all the operations are there, we should be fine. If not, we may even want to fail or create a new one (if it won't consume too many resources).

pkg/plugins/pg/source/source.go
dylanlott: I don't like the stutter that introduces, unless you mean making it a *CDCIterator instead of an Iterator? If it's the latter, I would rather keep it as the interface.

pkg/plugins/pg/source/cdc.go
dylanlott: When you start using an existing replication slot, you specify in the pglogrepl subscription where you want to start in the WAL log. Currently it always starts at zero, meaning it will read all of the contents of the log out. This will be changed to be passed to a specific LSN soon. We get access to a set of [pglogrepl message types](https://github.com/jackc/pglogrepl/blob/master/message.go), but I'm not sure how we could reliably differentiate between the absence of an event type and regular operation if we encounter no other errors.

pkg/plugins/pg/source/source.go
dylanlott: I think you're right, I updated it to `ErrNilConfig` which I think is more descriptive.

pkg/plugins/pg/source/cdc.go
hariso: Yup, that makes sense. Would be nice if we capture this in an issue, so we don't lose track of it.

pkg/plugins/pg/source/source.go
hariso: Yup, I meant making it a *CDCIterator instead of an Iterator, since it will always be that, and actually has to be that. But I don't feel too strongly on that.

pkg/plugins/pg/source/cdc.go
hariso: For the rest: we talked about this privately. Checking how an existing publication has been setup can be done in a separate PR.

pkg/plugins/pg/source/source.go
hariso: 👍

hariso (approved):

pkg/plugins/pg/source/cdc.go
lovromazgon: > Currently it always starts at zero, meaning it will read all of the contents of the log out. This will be changed to be passed to a specific LSN soon.

@dylanlott Is there a follow-up issue for that? If not we should create one.

pkg/plugins/pg/source/cdc.go
lovromazgon: Hmm, I'd rather use sort of an enum here and allow only two values (e.g. `true`/`false` or `enabled`/`disabled` or `on`/`off`), if the field contains anything else we return an error, the config is invalid, I wouldn't want to silently fall back if the user specified an invalid config option.

We should standardize this by providing some helper methods in the SDK for parsing the config and ensuring we use consistent values and error messages across plugins. No need to act on this now, just writing down my thoughts, I'll create an issue for this when I get the time.

dylanlott:
@lovromazgon I created #156 to track that bug.

PR #73, opened 2022-01-13
adds idempotent teardown test [2]

dylanlott:
### Description
Adds a test for idempotent calls to `Teardown`.

Relies on #72 

### Quick checks:


dylanlott:
Tests on this are expected to fail until #72 is merged in.

pkg/plugins/pg/source/source_integration_test.go
```diff
@@ -517,6 +517,49 @@ func TestRead_Iterator(t *testing.T) {
 	assert.True(t, src.sub == nil, "failed to cleanup subscription")
 }
 
+// TestSnapshotIdempotentTeardown asserts that Teardown can be called
+// idempotently in snapshot mode. Since the snapshot should not be finished,
+// we call it immediately and thus the snapshot interrupt error is expected.
```
hariso, line 522: I might be missing something, but why does Teardown() have to be idempotent? I do, however, see that it should return the correct error.

dylanlott: There was a discussion @lovromazgon and I had about making Teardown calls idempotent. It will be important when we go to cluster conduit as well.

hariso (approved):

Issue #65, opened 2022-01-03
Use conduit-plugin for connector plugin interface


lovromazgon:
The Conduit plugin gRPC interface was refactored and moved to https://github.com/conduitio/conduit-plugin. We should change Conduit to use the new interface and adjust the existing plugins to adhere to the new interface.

PR #85, opened 2022-01-18
Conduit-plugin [2]: Plugin registry

lovromazgon:
### Description

This creates the plugin registry interface that knows how to spin up a plugin. There are two registry implementations:
- standalone registry - is using go-plugin to spin up a plugin in a separate process and communicates with it via gRPC (same as the old approach)
- builtin registry - adds support for running a plugin in the same process as Conduit by bypassing the go-plugin library entirely

Related to #65

### Quick checks:


pkg/plugin/builtin/v1/destination.go
```diff
@@ -0,0 +1,333 @@
...
+// 	s.stopResp()
+// }
+//
+// func (s *destinationRunStream) stopReq() {
+// 	s.reqCancel()
+// }
+//
+// func (s *destinationRunStream) stopResp() {
+// 	s.respCancel()
+// }
```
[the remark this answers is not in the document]
lovromazgon: Whoops, seems like a leftover, need to remove it!

pkg/plugin/builtin/v1/destination.go
```diff
@@ -0,0 +1,333 @@
...
+	"github.com/conduitio/conduit/pkg/foundation/cerrors"
+	"github.com/conduitio/conduit/pkg/foundation/log"
+	"github.com/conduitio/conduit/pkg/plugin"
+	"github.com/conduitio/conduit/pkg/plugin/builtin/v1/internal/fromplugin"
+	"github.com/conduitio/conduit/pkg/plugin/builtin/v1/internal/toplugin"
+	"github.com/conduitio/conduit/pkg/record"
+	"github.com/rs/zerolog"
+)
+
+// TODO make sure a panic in a plugin doesn't crash Conduit
```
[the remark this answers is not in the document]
lovromazgon: I'll create an issue and update the comment 👍

pkg/plugin/registry.go
```diff
@@ -0,0 +1,52 @@
...
+	"strings"
+
+	"github.com/conduitio/conduit/pkg/foundation/log"
+)
+
+const builtinPluginPrefix = "builtin:"
+
+// registry is an object that can create new plugin dispensers. We need to use
+// an interface to prevent a cyclic dependency between the plugin package and
+// builtin and standalone packages.
```
dylanlott, line 27: Consider adding to this comment a summary of what we are defining as builtin and standalone packages. As someone coming into the code, it's not immediately apparent what the distinction is.

pkg/plugin/registry.go
```diff
@@ -0,0 +1,52 @@
...
+}
+
+func (r *Registry) New(logger log.CtxLogger, name string) (Dispenser, error) {
+	logger = logger.WithComponent("plugin")
+
+	if strings.HasPrefix(name, builtinPluginPrefix) {
+		return r.builtin.New(logger, strings.TrimPrefix(name, builtinPluginPrefix))
+	}
+
+	return r.standalone.New(logger, name)
```
hariso, line 51: Now that we have the option to register and dispense builtin plugins, it looks like we should also change the UI (replace the plugin paths with `builtin:plugin_name`), same for API (relevant for Lyubo?), and probably update the Docker images not to build the plugins (not needed anymore). Does that should right?

pkg/plugin/builtin/v1/destination.go
```diff
@@ -0,0 +1,231 @@
...
+	req, err := toplugin.DestinationConfigureRequest(cfg)
+	if err != nil {
+		return err
+	}
+	s.logger.Trace(ctx).Msg("calling Configure")
+	resp, err := s.impl.Configure(s.withLogger(ctx), req)
+	if err != nil {
+		// TODO create new errors as strings, this happens when they are
+		//  transmitted through gRPC and we don't want to falsely rely on types
+		//  in built-in plugins
```
hariso, line 66: Do you mind elaborating a bit more? I'm probably missing something, but I can't see how gRPC relates to the built-in plugins?

pkg/plugin/standalone/v1/destination.go
```diff
@@ -0,0 +1,166 @@
...
+		return plugin.ErrStreamNotOpen
+	}
+
+	protoReq, err := toproto.DestinationRunRequest(r)
+	if err != nil {
+		return err
+	}
+
+	err = s.stream.Send(protoReq)
+	if err != nil {
```
hariso, line 95: We could reduce the indentation and make it slightly more readable by inverting this condition.

pkg/plugin/standalone/v1/destination.go
```diff
@@ -0,0 +1,166 @@
...
+		if err == io.EOF {
+			// actual error can be discovered through Recv, let's do it
+			_, err = s.stream.Recv()
+			if err == nil {
+				// Recv did not return an error, we just read an ack, that's a huge bug!
+				panic(err)
```
hariso, line 101: I believe that the comment above could also be a good part of the panic message.

pkg/plugin/standalone/v1/dispenser.go
```diff
@@ -0,0 +1,191 @@
...
+
+	d.client = c
+	return nil
+}
+
+func (d *Dispenser) increaseDispensed() error {
+	d.m.Lock()
+	defer d.m.Unlock()
+
+	d.dispensed++
```
hariso, line 66: In which cases is it expected that this will be > 1, i.e. that we will use the same plugin client (and process) for let's say, `specify` and `source`?

pkg/plugin/standalone/v1/internal/fromproto/record.go
```diff
@@ -0,0 +1,64 @@
...
+	d := in.GetData()
+	if d == nil {
+		return nil, nil
+	}
+
+	switch v := d.(type) {
+	case *connectorv1.Data_RawData:
+		return record.RawData{
+			Raw:    v.RawData,
+			Schema: nil,
```
hariso, line 57: Is there any way we can get the schema?

pkg/plugin/standalone/v1/internal/fromproto/specifier.go
```diff
@@ -0,0 +1,66 @@
...
+		Description:       in.Description,
+		Version:           in.Version,
+		Author:            in.Author,
+		DestinationParams: destinationParams,
+		SourceParams:      sourceParams,
+	}
+	return out, nil
+}
+
+func SpecifierParameter(in *connectorv1.Specifier_Parameter) (plugin.Parameter, error) {
```
hariso, line 59: Maybe make it unexported?

pkg/plugin/standalone/v1/internal/toproto/record.go
```diff
@@ -0,0 +1,70 @@
...
+		if err != nil {
+			return nil, err
+		}
+		return &connectorv1.Data{
+			Data: &connectorv1.Data_StructuredData{
+				StructuredData: data,
+			},
+		}, nil
+	default:
+		return nil, cerrors.New("invalid Data type")
```
hariso, line 68: Can we also return the type name itself? I believe that would help troubleshooting.

pkg/plugin/standalone/v1/logger.go
```diff
@@ -0,0 +1,161 @@
...
+	return h.logger
+}
+
+func (h *hcLogger) applyArgs(e *zerolog.Event, args []interface{}) *zerolog.Event {
+	if h.name != "" {
+		// add plugin path name
+		e.Str(log.PluginNameField, h.name)
+	}
+	for i := 0; i < len(args)-1; i += 2 {
+		key := args[i].(string) // keys must be strings
```
hariso, line 144: If they aren't, trying to log something will actually panic, right?

pkg/plugin/standalone/v1/logger.go
```diff
@@ -0,0 +1,161 @@
...
+		switch key {
+		case "timestamp":
+			// skip timestamps, they are added by our internal logger
+			continue
+		case "@module":
+			key = log.ComponentField
+		}
+
+		val := args[i+1]
```
hariso, line 153: Similar to above: if we have an odd number of params, this would panic?

pkg/plugin/registry.go
lovromazgon: That's exactly how I imagine the next steps, yes!

pkg/plugin/builtin/v1/destination.go
lovromazgon: We should strive to make built-in plugins work exactly the same way as standalone plugins, so given that error types are lost in gRPC communication we should actively drop the type here as well. The idea is to not start relying on some specific error type in a test, and then find out that the flow breaks if the plugin is used in standalone mode.

pkg/plugin/standalone/v1/destination.go
lovromazgon: I mean, yes we could, but then the happy path wouldn't be on the left :/ If anything, I would say we could extract the logic inside of this `if` into a separate function.

pkg/plugin/standalone/v1/dispenser.go
lovromazgon: This is probably overengineered, I don't expect this to ever be over 1 tbh. And it shouldn't, because otherwise the plugin developer needs to care about it in case they are using some shared state.

We can simplify this!

pkg/plugin/standalone/v1/internal/fromproto/specifier.go
lovromazgon: I feel like I saw this comment before 😅   I see this (and other `fromX` and `toX` packages) as a utility package, which doesn't care about how and where it's used, it just provides the functionality, so I wouldn't hide this function as it's just another conversion function.

pkg/plugin/standalone/v1/internal/toproto/record.go
lovromazgon: Good point, will add that info!

pkg/plugin/standalone/v1/logger.go
lovromazgon: Yes, good point, we might want to make this safer so a simple log message doesn't kill the whole plugin.

pkg/plugin/standalone/v1/logger.go
lovromazgon: In this case it wouldn't panic, because of how the for loop condition is set up, but the last parameter would get lost in the matrix :neckbeard:

pkg/plugin/registry.go
hariso: Thanks for the confirmation. If not already, it would be good to have those post-merge steps written down somewhere.

pkg/plugin/builtin/v1/destination.go
hariso: Sounds great to me, and I think your comment here would be a good one to be added to the source code too.

pkg/plugin/standalone/v1/destination.go
hariso: I think the latter would improve the readability too. 👍

pkg/plugin/standalone/v1/internal/fromproto/specifier.go
hariso: Yeah, I'm sure I asked it before somewhere, but, you know "better ask, then be sorry". You're our internal Google.:D

pkg/plugin/standalone/v1/internal/toproto/record.go
lovromazgon: Done.

pkg/plugin/standalone/v1/logger.go
lovromazgon: Changed it so we log all params, even if it's an odd number.

pkg/plugin/standalone/v1/client.go
```diff
@@ -0,0 +1,145 @@
...
+			"exec", pluginPath,
+		}
+
+		in.Logger.Info("DELVE: starting plugin", "port", port)
+		return nil
+	})
+}
+
+func getFreePort() int {
+	l, err := net.Listen("tcp", "localhost:0")
```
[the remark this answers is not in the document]
lovromazgon: Added a comment.

pkg/plugin/registry.go
lovromazgon: Done.

pkg/plugin/builtin/v1/destination.go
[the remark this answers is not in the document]
lovromazgon: Removed!

pkg/plugin/standalone/v1/internal/fromproto/record.go
hariso: 👍

hariso (approved):
Great work, Lovro! There's only this non-blocker left I believe: https://github.com/ConduitIO/conduit/pull/85/files#r802448561

lovromazgon:
Thanks @hariso, I will create issues later today, among them the followup you mention in your comment!

PR #88, opened 2022-01-18
Conduit-plugin [5]: File plugin

lovromazgon:
### Description

This adjusts the file plugin to use the new plugin SDK.

Related to #65

### Quick checks:


pkg/plugins/file/destination.go
```diff
@@ -95,3 +94,20 @@ func openOrCreate(path string) (*os.File, error) {
...
+		return requiredConfigErr(ConfigPath)
+	}
+
+	// make sure we can stat the file, we don't care if it doesn't exist though
+	_, err := os.Stat(cfg[ConfigPath])
+	if err != nil && !os.IsNotExist(err) {
+		return cerrors.Errorf(
+			"%q config value does not contain a valid path: %w",
+			ConfigPath, err,
+		)
```
hariso, line 109: It might be nice to include the path value itself too.

hariso (approved):

pkg/plugins/file/source.go
```diff
@@ -15,49 +15,67 @@
...
+
+func (s *Source) Read(ctx context.Context) (sdk.Record, error) {
+	select {
+	case line := <-s.tail.Lines:
+		return sdk.Record{
+			Position:  sdk.Position(strconv.FormatInt(line.SeekInfo.Offset, 10)),
+			CreatedAt: line.Time,
+			Payload:   sdk.RawData(line.Text),
+		}, nil
+	case <-ctx.Done():
```
maha-hajja, line 62: not needed anymore :D

lovromazgon: Not really, it is still needed here! We are not using `sdk.ErrBackoffRetry` in this plugin, instead, we are blocking the call until we get a new line _or_ the context gets canceled.

PR #123, opened 2022-01-25
Conduit-plugin [6]: S3 plugin

maha-hajja:
### Description

This adjusts the S3 plugins to use the new plugin SDK.

Related to #65 

### Quick checks:


pkg/plugins/s3/destination/destination.go
```diff
@@ -18,96 +18,111 @@ import (
...
-}
-
-// Write writes a record into a Destination. Typically Destination maintains an in-memory
+// WriteAsync writes a record into a Destination. Typically Destination maintains an in-memory
 // buffer and doesn't actually perform a write until the buffer has enough
 // records in it. This is done for performance reasons.
-func (s *Destination) Write(ctx context.Context, r record.Record) (record.Position, error) {
-	// If either Destination or Writer have encountered an error, there's no point in
+func (d *Destination) WriteAsync(ctx context.Context, r sdk.Record, ackFunc sdk.AckFunc) error {
+	// If either Destination or Writer have encountered an error, there'd no point in
```
lovromazgon, line 85: I don't think you wanted to replace this `s` with a `d` 🙂

pkg/plugins/s3/destination/destination.go
```diff
@@ -18,96 +18,111 @@ import (
...
+
+		// call all the written records' ackFunctions
+		for _, ack := range d.AckFuncCache {
+			err := ack(d.Error)
+			if err != nil {
+				return err
+			}
 		}
+		// clear ackFunc cache
+		d.AckFuncCache = make([]sdk.AckFunc, 0, d.Config.BufferSize)
```
lovromazgon, line 119: Sidenote: we could make this more performant by not allocating a new slice but rather reusing the old one (same goes for `d.Buffer`.
```suggestion
		d.AckFuncCache = d.AckFuncCache[:0]
```

pkg/plugins/s3/spec.go
```diff
@@ -14,17 +14,85 @@
...
+			},
+			destination.ConfigKeyBufferSize: {
+				Default:     "1000",
+				Required:    false,
+				Description: "the buffer size {when full, the files will be written to destination}, max is \"100000\".",
+			},
+			destination.ConfigKeyFormat: {
+				Default:     "",
+				Required:    false,
+				Description: "the destination format, either \"json\" or \"parquet\".",
```
lovromazgon, line 62: Protip: use backticks when you need to use quotes in a string, then you don't have to escape them:
```suggestion
				Description: `the destination format, either "json" or "parquet".`,
```

pkg/plugins/s3/source/iterator/snapshot_iterator.go
```diff
@@ -134,19 +128,18 @@ func (w *SnapshotIterator) Next(ctx context.Context) (record.Record, error) {
...
+		Payload:   sdk.RawData(rawBody),
+		Key:       sdk.RawData(*key),
 		CreatedAt: *object.LastModified,
 	}
 
 	return output, nil
 }
+func (w *SnapshotIterator) Stop() {
+	// nothing to stop
+}
```
lovromazgon, line 145: Do we need this function?

lovromazgon (changes requested):
We need to implement the `Flush` function in `Destination` which will write out any records that are still in the buffer.

pkg/plugins/s3/source/iterator/snapshot_iterator.go
maha-hajja: not really.. just added it so it would implement the Iterator interface

pkg/plugins/s3/destination/destination.go
hariso: If you're using GoLand, then you can put your cursor on the `d`, and then press shift+F6. GoLand will rename all the occurrences of `d`, but it will also ask you to rename all the other receivers.

pkg/plugins/s3/destination/destination.go
```diff
@@ -18,96 +18,135 @@ import (
...
 
-	s.Buffer = append(s.Buffer, r)
+	d.Buffer = append(d.Buffer, r)
+	d.AckFuncCache = append(d.AckFuncCache, ackFunc)
 
-	if len(s.Buffer) >= int(s.Config.BufferSize) {
-		bufferedRecords := s.Buffer
-		s.Buffer = make([]record.Record, 0, s.Config.BufferSize)
+	if len(d.Buffer) >= int(d.Config.BufferSize) {
+		bufferedRecords := d.Buffer
```
lovromazgon, line 99: We could simply call `Flush` here.

lovromazgon (commented):
One last comment that simplifies the code, otherwise we just need to resolve the conflicts and this should be ready for a merge!

pkg/plugin/builtin/registry.go
```diff
@@ -20,12 +20,18 @@ import (
...
+	"github.com/conduitio/conduit/pkg/plugins/file"
 	"github.com/conduitio/conduit/pkg/plugins/generator"
+	"github.com/conduitio/conduit/pkg/plugins/s3"
+	s3destination "github.com/conduitio/conduit/pkg/plugins/s3/destination"
+	s3source "github.com/conduitio/conduit/pkg/plugins/s3/source"
 )
 
 var (
 	DefaultDispenserFactories = []DispenserFactory{
+		sdkDispenserFactory(file.Specification, file.NewSource, file.NewDestination),
```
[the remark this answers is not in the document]
lovromazgon: That's strange, the code is definitely in `lovro/conduit-plugin`, not sure how the diff is calculated here. I hope this will go away after conflicts are resolved.

lovromazgon (approved):
✅

Issue #32, opened 2021-06-21
Pipeline file


lovromazgon:
Collects work needed to implement pipeline config file support in Conduit.
- [x] #489
- [x] #490
- [x] #491
- [x] #492
- [x] #493
- [x] #33

lovromazgon:
Make sure values can be injected through environment variables.

PR #126, opened 2022-01-25
[UI] Add runtime API URL host

jmar910:
### Description
Generally, API hosts (and other environment variables) are written at build time if not using the same host. Since we build the conduit UI into the conduit binary, we do not know at build time what the API host is, so we fallback to ember-data and axios with no host. 

There is some further work slated to further define and optimize our configurations for enabling/disabling the UI (and other configurations). See #32 

### Quick checks:

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -7,8 +7,8 @@ RUN apt-get update &&\
     apt-get install -y build-essential &&\
     apt-get install -y git
 
-# Install Node@v12
-RUN curl -sL https://deb.nodesource.com/setup_12.x | bash - &&\
+# Install Node@v16
+RUN curl -sL https://deb.nodesource.com/setup_16.x | bash - &&\
     apt-get install -y nodejs &&\
     npm update &&\
     npm i -g yarn@1.22.17
--- a/ui/app/models/pipeline.js
+++ b/ui/app/models/pipeline.js
@@ -10,6 +10,8 @@ const STATUS_MAP = {
   STATUS_RUNNING: 'running',
 };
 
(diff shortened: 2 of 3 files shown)


ui/app/adapters/application.js
```diff
@@ -10,7 +10,9 @@ export default class ApplicationAdapter extends RESTAdapter {
   flashMessages;
 
   namespace = 'v1';
-  host = config.conduitAPIURL;
+  host = config.conduitAPIURL
+    ? config.conduitAPIURL
+    : `${window.location.protocol}//${window.location.host}`;
```
jayjayjpg, line 15: Do we want to use the `window.location.host` here as well or is `window.location.protocol` already returning the expected value?

jayjayjpg: I think it may just be the way the code format is rendered here that makes `//${window.location.host}` look as if it was a comment instead of part of the string -- nevermind then!

jayjayjpg (approved):
This reads great! :sparkles:

ui/app/adapters/application.js
jmar910: Ah yeah!! looks like theres a highlighting issue here with github, I can confirm this works via @anaptfox & @devarismeroxa . I built them a Docker image using this change, and they were successfully able to deploy on DO and built pipelines start to finish using the UI remotely.

jayjayjpg: That sounds perfect then, thank you for the outline!

dylanlott (approved):
LGTM!

PR #138, opened 2022-01-28
Make builds for more platforms

### LINKED ISSUE #50: Provide pre-built binaries for Windows and Macs with M1 chips
This is a follow-up on ConduitIO/conduit-old#438.

Due to reasons mentioned in the PR, we currently do not have pre-built binaries of Conduit and plugins for Windows and also Macs with M1 chips. 

We should investigate what are our options and what needs to be done so we get them as well.

hariso:
### Description

Provides builds for more platforms (macOS M1, Windows). Depends on https://github.com/ConduitIO/conduit/pull/105.

Fixes #50 

### Quick checks:

- [ ] I have written unit tests. -- Tested on a fork of the repo: https://github.com/hariso/conduit


.goreleaser.yml
```diff
@@ -59,15 +71,14 @@ archives:
...
       - generator
       - kafka
-    allow_different_binary_count: true
     replacements:
-      darwin: Darwin
+      darwin: macOS
       linux: Linux
       windows: Windows
       386: i386
       amd64: x86_64
```
lovromazgon, line 79: Any specific reason we have these replacements? Why not just use the defaults?

lovromazgon (commented):
Just a comment regarding the name of the produced files, otherwise looks like it should work 👍

.goreleaser.yml
hariso: I've got them from a template. The OS names appear a bit prettier with the capitalization. `i386` and `x86_64` _I think_ are closer to being official names for the respective architectures.

lovromazgon: I guess it's a matter of preference, but specifically the capitalization is what bothers me the most 😄  The assets that we produce have all lowercase letters except the OS, it really caught my eye. Same with `macOS` vs `darwin`, as a developer I'm used to seeing `darwin`, and since Conduit is a tool for developers I feel like we should stick to that. All in all, I feel like the replacements cause more confusion than benefits, I'd rather defer to the opinion of the go-releaser creators, who probably chose the names for a good reason.

hariso: I completely agree that it's a matter of preference. I've seen both, darwin and macOS, being used in file names, but now that I think of it, I do think darwin is being used more frequently, and I can change that.

However, with or without that, and also with or without capitalization, I don't think it really matters a lot. It's easy to spot the binary you need, which is I believe the most important thing here. And regarding GoReleaser, they do capitalization too.

lovromazgon: Fair enough. I now see the value of these replacements (as explained [here](https://github.com/goreleaser/goreleaser/issues/236#issuecomment-302376484) it's easy to build the download link), although I personally am still inclined to use the defaults, because the values would correspond to the values of [GOOS and GOARCH](https://go.dev/doc/install/source#environment).

Anyway, I'm fine with how it is now 👍

lovromazgon (approved):

PR #164, opened 2022-02-03
Nightly builds

### LINKED ISSUE #117: Nightly Builds
### Feature description

This will help Meroxa internally to get critical bugs fixed and then have them into the platform ASAP.

hariso:
### Description

This PR adds a GH action which triggers nightly builds and cleans up older ones. I've added a bit of docs to explain how it works.

This setup has been running every 15 minutes in [my fork of this repo](https://github.com/hariso/conduit) since yesterday afternoon, where I've verified that the cleanup of releases and images works. (**edit**: I've just deleted older releases, since I replaced 'snapshot' with 'nightly')

A few additional notes:
1. The next minor version is hard-coded. I plan improving that separately, but we should be fine for at least a few months. Also, it won't help when bumping up the major version, but that's a rare event, so it shouldn't be a big thing.
2. Also, as you'll notice in the fork, the nightly build will run for the same commit (if there were no changes since the last run). I've intentionally left that check out since I think it's not relevant for us. We have new commits every day, and most of time those are not dependency upgrades.
3. It's possible to manually trigger a nightly build.
4. ~I'm looking into an issue where the `v` is left out in Docker image tags.~

Resolves #117 

### Quick checks:

-  [x] I have made sure that the PR is of reasonable size and can be easily reviewed.


docs/releases.md
```diff
@@ -5,9 +5,30 @@ A Conduit release has the following parts:
...
 
-### How to release a new version?
+### How to release a new version
 A release is triggered by pushing a new tag which starts with `v` (for example `v1.2.3`). Everything else is then handled by 
-GoReleaser and GitHub actions. To push a new tag, please use the script [tag.sh](https://github.com/ConduitIO/conduit/blob/main/scripts/tag.sh), 
-which also checks if the version conforms to SemVer.
+GoReleaser and GitHub actions. To push a new tag, please use the script [scripts/tag.sh](https://github.com/ConduitIO/conduit/blob/main/scripts/tag.sh), 
+which also checks if the version conforms to SemVer. Example:
+```
+scripts/tag.sh 1.2.3
```
lovromazgon, line 16: Shouldn't this be `scripts/tag.sh v1.2.3`?

.github/workflows/trigger-nightly.yml
```diff
@@ -0,0 +1,46 @@
...
+    runs-on: ubuntu-latest
+    steps:
+      -
+        name: 'Checkout'
+        uses: actions/checkout@v2
+        with:
+          token: ${{ secrets.NIGHTLY_BUILD_GH_TOKEN }}
+      -
+        name: 'Push new tag'
+        # TODO determine nightly version based on latest minor version
```
lovromazgon, line 21: This is something we have to do before releasing v0.2.0. Do you want to add it to this PR or create a follow-up?

.github/workflows/trigger-nightly.yml
```diff
@@ -0,0 +1,46 @@
...
+        run: |
+          git config user.name "${GITHUB_ACTOR}"
+          git config user.email "${GITHUB_ACTOR}@users.noreply.github.com"
+
+          TAG=v0.2.0-nightly.$(date +'%s')
+          git tag -a $TAG -m "$TAG: nightly build"
+          git push origin $TAG
```
lovromazgon, line 28: So I guess if we didn't commit anything the tag will be added to the same commit and it is as if nothing happened - we don't trigger another nightly build, right?

.github/workflows/trigger-nightly.yml
```diff
@@ -0,0 +1,46 @@
...
+      - name: 'Clean up nightly releases'
```
lovromazgon, line 29: What if we didn't commit anything for 7 days? We will be left without a nightly release, right? 😕  I would rather make sure we always have at least 1 nightly release - is there an easy way to configure this (i.e. we don't delete it if it's the last one)?

.github/workflows/trigger-nightly.yml
```diff
@@ -0,0 +1,46 @@
...
+        env:
+          GITHUB_TOKEN: ${{ secrets.NIGHTLY_BUILD_GH_TOKEN }}
+      - name: 'Delete nightly containers older than a week'
+        uses: snok/container-retention-policy@v1
+        with:
+          image-names: conduit
+          cut-off: 1 week ago UTC
+          account-type: org
+          org-name: ConduitIO
+          keep-at-least: 7
```
lovromazgon, line 44: This configures how many docker images to keep, this relates to my previous comment and I think it should be set to 1.

.github/workflows/trigger-nightly.yml
hariso: It will actually trigger another build, since every tag push will trigger a release (regular or nightly).

.github/workflows/trigger-nightly.yml
hariso: It will be the same as above: we will always keep 7 images.

docs/releases.md
hariso: It's actually 1.2.3 and then the script adds the prefix.

.github/workflows/trigger-nightly.yml
hariso: I think it would be good to see if there's much work left in the other comments/feedback. If not, I'll try adding it. If yes, I think it would be good to merge sooner rather than later, so we can start building the nigthlies.

.github/workflows/trigger-nightly.yml
lovromazgon: I see, makes sense. Last thing here - would it make sense to change the date format to `YYYYMMDD`? Then the tag would be readable, we would quickly have an idea when the nightly was created.

.github/workflows/trigger-nightly.yml
lovromazgon: If we change the cron schedule to weekdays then this should be 5, right?

.github/workflows/trigger-nightly.yml
```diff
@@ -0,0 +1,46 @@
+name: Trigger nightly build
+
+on:
+  schedule:
+    # * is a special character in YAML so you have to quote this string
+    - cron:  '0 * * * *'
```
lovromazgon, line 6: As mentioned in the meeting - we should change this to weekdays only.

.github/workflows/trigger-nightly.yml
lovromazgon: Good point. It's your call how you want to proceed, as long as you pick one of those options 👍

.github/workflows/trigger-nightly.yml
hariso: That's a good point. I actually had it like that initially, but changed to the timestamp, because I was running the builds every 15 minutes in my fork.: )

lovromazgon: Yeah, it should be the last change you do, once you're sure it works 😄

.github/workflows/trigger-nightly.yml
hariso: Yup! Done.

.github/workflows/trigger-nightly.yml
hariso: Done.

.github/workflows/trigger-nightly.yml
hariso: Opened a new issue: https://github.com/ConduitIO/conduit/issues/186.

.github/workflows/trigger-nightly.yml
hariso: Changed date format.

PR #170, opened 2022-02-03
Conduit-plugin [7]: Kafka plugin

hariso:
### Description

This adjusts the Kafka plugin to use the new plugin SDK.

Related to https://github.com/ConduitIO/conduit/issues/65

### Quick checks:


pkg/plugins/kafka/source.go
```diff
@@ -21,67 +21,65 @@ import (
...
+
+func (s *Source) Read(ctx context.Context) (sdk.Record, error) {
 	message, kafkaPos, err := s.Consumer.Get(ctx)
 	if err != nil {
-		return record.Record{}, cerrors.Errorf("failed getting a message %w", err)
+		return sdk.Record{}, cerrors.Errorf("failed getting a message %w", err)
 	}
 	if message == nil {
-		return record.Record{}, plugins.ErrEndData
+		return sdk.Record{}, plugins.ErrEndData
```
lovromazgon, line 72: This error should now be `sdk.ErrBackoffRetry`. In the future we should rewrite `s.Consumer.Get` to block until either the context is cancelled or a message is read.

pkg/plugins/kafka/source.go
```diff
@@ -95,16 +93,21 @@ func (s *Source) startFrom(position record.Position) error {
 	return nil
 }
 
-func toRecord(message *kafka.Message, position string) (record.Record, error) {
-	return record.Record{
+func toRecord(message *kafka.Message, position string) (sdk.Record, error) {
+	return sdk.Record{
 		Position:  []byte(position),
 		CreatedAt: time.Time{},
```
lovromazgon, line 99: > `CreatedAt` represents the time when the change occurred in the source system. If that's impossible to find out, then it should be the time the change was detected by the connector.

Not sure if `*kafka.Message` contains a timestamp of when it was created, otherwise we should use `time.Now()`.

lovromazgon (changes requested):
Generally looks good, we still need to adjust the tests in [pkg/plugins/kafka/destination_integration_test.go](https://github.com/ConduitIO/conduit/pull/170/files#diff-47b1720c6724a2257661a210b28b28dfc7f4aa780ec8a5e038779da12cb99cc3).

pkg/plugins/kafka/source.go
hariso: 👍 Done.

pkg/plugins/kafka/source.go
hariso: Good point 👍 Done.

pkg/plugins/kafka/source.go
```diff
@@ -95,16 +91,21 @@ func (s *Source) startFrom(position record.Position) error {
...
 }
 
-func (s *Source) Ack(context.Context, record.Position) error {
+func (s *Source) Ack(ctx context.Context, position sdk.Position) error {
 	return s.Consumer.Ack()
 }
+
+func (s *Source) Teardown(context.Context) error {
+	fmt.Println("Tearing down a Kafka Source...")
+	s.Consumer.Close()
```
lovromazgon, line 109: Teardown can and will be called even if the plugin didn't start consuming, for example when the connector is created Conduit calls `Configure` and then `Teardown`. This line panics in that case.

The other issue is that Conduit logs the error but then happily continues without failing the creation of the connector. Another followup, yay!

pkg/plugins/kafka/destination.go
```diff
@@ -19,52 +19,58 @@ import (
...
 }
 
-// Validate takes config and returns an error if some values are missing or incorrect.
-func (s *Destination) Validate(cfg plugins.Config) error {
-	_, err := Parse(cfg.Settings)
-	return err
+// Teardown shuts down the Kafka client.
+func (d *Destination) Teardown(context.Context) error {
+	fmt.Println("Tearing down a Kafka Destination...")
+	d.Client.Close()
```
lovromazgon, line 74: Same here.

pkg/plugins/kafka/destination.go
```diff
@@ -19,52 +19,58 @@ import (
...
 	return nil
 }
 
-// Validate takes config and returns an error if some values are missing or incorrect.
-func (s *Destination) Validate(cfg plugins.Config) error {
-	_, err := Parse(cfg.Settings)
-	return err
+// Teardown shuts down the Kafka client.
+func (d *Destination) Teardown(context.Context) error {
+	fmt.Println("Tearing down a Kafka Destination...")
```
lovromazgon, line 73: For logging we can and should now use `sdk.Logger(ctx)`, this way we can produce leveled and structured logs!

lovromazgon (changes requested):
I tried the plugin out and noticed a couple of things that need to be addressed, otherwise works like a charm 👍

pkg/plugins/kafka/source.go
lovromazgon: Followup: https://github.com/ConduitIO/conduit/issues/224

pkg/plugins/kafka/destination.go
hariso: 👍 Done.

pkg/plugins/kafka/destination.go
hariso: 👍 Done.

pkg/plugins/kafka/source.go
hariso: 👍 Done.

pkg/plugins/kafka/consumer.go
```diff
@@ -115,7 +115,7 @@ func (c *segmentConsumer) Close() {
 	// this will also make the loops in the reader goroutines stop
 	err := c.reader.Close()
```
lovromazgon, line 116: It would make sense to return the error here and then return the error in `Teardown`. We could sneak it into a PR once we touch this code again!

lovromazgon (approved):
LGTM 👍  :shipit:

pkg/plugins/kafka/consumer.go
hariso: 👍 Sounds good!

PR #191, opened 2022-02-07
Removes key from payload to prevent duplicate keys in query

### LINKED ISSUE #175: Bug: keys are being duplicated in column upsert query causing invalid query errors
### Bug description

There is an issue where the PostgreSQL destination adds the key twice to the list of columns to be queried which results in an incorrect query error.

### Steps to reproduce

1. Configure a pipeline with a Postgres Source on a table of schema `id serial, data varchar` using the default key and column settings that feeds into a Postgres Desination that is using default settings. 
2. Start the pipeline and the Destination will error with an incorrect query error with `id` being added twice.

### Version

v0.1.0

dylanlott:
### Description

If the record's `Key` was in the `Payload`, it caused column values to be listed twice in the upsert query, causing an error.

This PR skips adding a `Payload` field to the upsert query if the field's name is the same as the record's `Key` name.

Fixes #175 

### Quick checks:

### DIFF
--- a/pkg/plugins/pg/destination/destination.go
+++ b/pkg/plugins/pg/destination/destination.go
@@ -99,13 +99,15 @@ func (s *Destination) Write(ctx context.Context, r record.Record) (record.Positi
 		}
 
 		// add key data to the query
-		for k, v := range d {
-			colArgs = append(colArgs, k)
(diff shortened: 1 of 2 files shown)


lyuboxa (approved):
sweet!

PR #235, opened 2022-02-16
Support for passing flags to go test via make

lovromazgon:
### Description

This change removes the `-v` flag from go test by default, which makes for a nicer output in the CLI that makes it easier to spot which test failed. It also allows you to set the variable `GOTEST_FLAGS` when calling `make test` or `make test-integration` and supplying additional flags to `go test`. The Github action is adjusted to pass `-v -count=1` which enables verbose logging (same as it was before) and disables test caching (caching is not happening anyway on the CI, but just to be sure).

### DIFF
--- a/pkg/plugins/kafka/destination_integration_test.go
+++ b/pkg/plugins/kafka/destination_integration_test.go
@@ -48,7 +48,7 @@ func TestDestination_Write_Simple(t *testing.T) {
 	assert.Ok(t, writeErr)
 	assert.Equal(t, record.Position, result)
 
-	message, err := waitForReaderMessage(cfg.Settings[kafka.Topic], 10*time.Second)
+	message, err := waitForReaderMessage(cfg.Settings[kafka.Topic], 15*time.Second)
 	assert.Ok(t, err)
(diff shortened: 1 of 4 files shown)


maha-hajja (approved):
LGTM

pkg/plugins/kafka/destination_integration_test.go
```diff
@@ -48,7 +48,7 @@ func TestDestination_Write_Simple(t *testing.T) {
 	assert.Ok(t, writeErr)
 	assert.Equal(t, record.Position, result)
 
-	message, err := waitForReaderMessage(cfg.Settings[kafka.Topic], 10*time.Second)
+	message, err := waitForReaderMessage(cfg.Settings[kafka.Topic], 15*time.Second)
```
lovromazgon, line 51: I increased this as I was hitting the timeout when running the test locally.

lovromazgon (commented):
I added a healthcheck for the kafka container and configured the docker compose command to wait until the containers are healthy before running integration tests.

This change requires docker compose v2.1.1 or higher, the CI image currently includes v2.2.3 so we're safe there.

hariso (approved):

PR #265, opened 2022-02-28
Allow usage of inmemory db

lovromazgon:
### Description

Allow users to use flag `--db.type=inmemory` to store pipeline configurations in memory. The caveat is that changes are not persisted when Conduit restarts so a warning is printed if Conduit starts with this option:

```
WRN Using in-memory store, all pipeline configurations will be lost when Conduit stops.
```

### Quick checks:


maha-hajja (approved):
LGTM

PR #230, opened 2022-02-15
New connector plugin interface

lovromazgon:
### Description

This PR combines the work of 11 different PRs that were separately reviewed:
- https://github.com/ConduitIO/conduit/pull/84
- https://github.com/ConduitIO/conduit/pull/85
- https://github.com/ConduitIO/conduit/pull/86
- https://github.com/ConduitIO/conduit/pull/87
- https://github.com/ConduitIO/conduit/pull/88
- https://github.com/ConduitIO/conduit/pull/123
- https://github.com/ConduitIO/conduit/pull/170
- https://github.com/ConduitIO/conduit/pull/262
- https://github.com/ConduitIO/conduit/pull/279
- https://github.com/ConduitIO/conduit/pull/231
- https://github.com/ConduitIO/conduit/pull/239

Closes #65, closes #218, closes #219, closes #174, closes #156.

Issue #118, opened 2022-01-24
Describe Connector Plugin API


neovintage:
### Feature description

Need an API endpoint that will allow the developer to know what connectors and capabilities of the connector exist.

- [x] proto design https://github.com/ConduitIO/conduit/pull/268
- [x] orchestration
- [x] implementation https://github.com/ConduitIO/conduit/pull/311

Issue #180, opened 2022-02-04
Feature: Endpoint for plugin config validation


lovromazgon:
### Feature description

We should provide an endpoint that accepts a plugin name and config and either returns an empty response or an error describing why the config is invalid.

PR #268, opened 2022-03-01
create a plugin service API proto

maha-hajja:
### Description

This PR describes the Plugin Service APIs. the implementation will be delayed until #230 is merged

Related to #118 

### Quick checks:


proto/api/v1/api.proto
```diff
@@ -799,3 +800,61 @@ option (grpc.gateway.protoc_gen_openapiv2.options.openapiv2_swagger) = {
...
+    TYPE_BOOL = 4;
+    TYPE_ENUM = 5;
+  }
+
+  message Parameter {
+    string default = 1;
+    bool required = 2;
+    string description = 3;
+    Type type = 4;
+    repeated string enum = 5; // specify the enum expected values
```
maha-hajja, line 849: any other suggestions for this param name? enum-values or values maybe?

hariso: What's this enum about btw?

proto/api/v1/api.proto
```diff
@@ -799,3 +800,61 @@ option (grpc.gateway.protoc_gen_openapiv2.options.openapiv2_swagger) = {
...
+service PluginService {
+  rpc ListPlugins(ListPluginsRequest) returns (ListPluginsResponse) {
+    option (google.api.http) = {
+      get: "/v1/plugins"
+      response_body: "plugins"
+    };
+  };
+  rpc GetPlugin(GetPluginRequest) returns (GetPluginResponse) {
+    option (google.api.http) = {
+      get: "/v1/plugin/{name}"
```
lovromazgon, line 813: I'm thinking about the name, it can either be `builtin:[plugin-name]` or the path to the plugin. The path might be absolute (i.e. start with a slash), I doubt the grpc-gateway would even parse this correctly. Sidenote - this also made me think about #180, where we need an endpoint for validating a plugin config, in that case something like `/v1/plugin/{name}/validate` could make sense, but it wouldn't be possible to determine if the `/validate` is part of the name or the action. We might have to rethink how we identify plugins.

Anyway, we could remove this endpoint and instead add a name filter in the `ListPlugins` rpc, same as we do in `ListPipelines`, I think that makes the most sense right now.

proto/api/v1/api.proto
```diff
@@ -799,3 +800,61 @@ option (grpc.gateway.protoc_gen_openapiv2.options.openapiv2_swagger) = {
...
+message GetPluginRequest {
+  string name = 1;
+}
+
+message GetPluginResponse {
+  PluginSpecifications plugin = 1;
+}
+
+message PluginSpecifications{
+  enum Type {
```
lovromazgon, line 835: I would move this `Type` inside of the `Parameter` message, to make it clear that this is a parameter type.

proto/api/v1/api.proto
lovromazgon: It's meant to hold the enum values in case `type` is `TYPE_ENUM`. This way the UI could display a dropdown instead of a freeform field, a great example would be the [acks](https://github.com/ConduitIO/conduit-plugin-kafka/blob/361df379fa98cd9cfcda439120e9a0752f6230b4/config.go#L31) option in the kafka connector, which could define the allowed fields `none`, `one` and `all`.

That said, I've been thinking of another possible approach - defining an enum could be compared with setting up a validation. To define an enum the field would still have the type `TYPE_STRING` but a validation would be added on top which defines what are the allowed values. If we have a way to add validations to a field it would add more flexibility, the UI and Conduit could validate the config (to some degree) before it even gets to the plugin.

While it adds flexibility, it also adds complexity. The question is if this is something we want or not.

lovromazgon (changes requested):

proto/api/v1/api.proto
```diff
@@ -799,3 +800,78 @@ option (grpc.gateway.protoc_gen_openapiv2.options.openapiv2_swagger) = {
...
+
+message GetPluginRequest {
+  string name = 1;
+}
+
+message GetPluginResponse {
+  PluginSpecifications plugin = 1;
+}
+
+message PluginSpecifications{
```
maha-hajja, line 830: should I rename this to `ConnectorSpecifications` ?

lovromazgon: No, internally we still have to make a difference between plugins and connectors, they are distinct entities in Conduit.

proto/api/v1/api.proto
```diff
@@ -828,34 +828,42 @@ message GetPluginResponse {
...
-  message Validation{
-    enum Type {
-      TYPE_UNSPECIFIED = 0;
-      TYPE_REQUIRED = 1;
-      TYPE_INCLUSION = 2;
-      TYPE_EXCLUSION = 3;
-      TYPE_GREATER_THAN = 4;
-      TYPE_LESS_THAN = 5;
-      TYPE_REGEX = 6;
+      Type validation =1;
```
lovromazgon, line 849: Nitpick: I'd name this `type` so we have a `Validation.type` instead of `Validation.validation`.
```suggestion
      Type type = 1;
```

lovromazgon (approved):
Looking good ✨

PR #292, opened 2022-03-08
Fix container cleanup

### LINKED ISSUE #291: Delete old nightly docker images
Old nightly docker images are not cleaned up right now (e.g. see [this run](https://github.com/ConduitIO/conduit/runs/5457318263?check_suite_focus=true#step:6:15)), the error is:
```
Couldn't delete conduit:14831694.
Status code: 404
Response: {'message': 'Package not found.', 'documentation_url': 'https://docs.github.com/rest/reference/packages#delete-a-package-version-for-an-organization'}
```

lovromazgon:
### Description

This hopefully fixes the issue with the container cleanup. In the [package](https://github.com/ConduitIO/conduit/pkgs/container/conduit) the install command looks like this:
```
docker pull ghcr.io/conduitio/conduit:latest-nightly
```

Notice that the organization is all lower case. I have a _hunch_ this might be the reason for the cleanup job to fail.

This PR also changes the schedule of when the nightly job runs (Tue-Sat), I think it makes more sense since it now happens after each workday.

Fixes #291

### DIFF
--- a/.github/workflows/trigger-nightly.yml
+++ b/.github/workflows/trigger-nightly.yml
@@ -41,7 +41,7 @@ jobs:
           image-names: conduit
           cut-off: 1 week ago UTC
           account-type: org
-          org-name: conduitio
+          org-name: ConduitIO
           keep-at-least: 5
           token: ${{ secrets.NIGHTLY_BUILD_GH_TOKEN }}
           filter-tags: '*-nightly*'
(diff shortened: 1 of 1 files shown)


lovromazgon:
The change didn't fix the issue 😕  https://github.com/ConduitIO/conduit/runs/5464034337?check_suite_focus=true#step:6:15

lovromazgon:
This is fixed now, I created the @ConduitIO/automation team, added @meroxa-machine to the team and gave it admin permissions to the repo.

I still think it would make sense to change the schedule, that's the only change left in this PR now.

hariso (approved):

PR #298, opened 2022-03-14
Clean up code related to old plugin interface

### LINKED ISSUE #221: Plugin: Clean up code related to old plugin interface
### Feature description

In the process of moving to the new connector plugin interface, we left some leftovers in the code, we need to clean it up. Namely the code in `pkg/plugins`, proto files, make target for generating proto code.

maha-mstfa:
### Description

Clean up conduit code related to old plugin interface

Fixes #221 

### Quick checks:

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -70,7 +70,7 @@ jobs:
             org.opencontainers.image.vendor=ConduitIO
 
       - name: Build and push Docker image
-        uses: docker/build-push-action@v2.9.0
+        uses: docker/build-push-action@v2.10.0
         with:
           context: .
           push: true


lovromazgon (changes requested):
Three more things that fell to mind:
- After this is merged we need to remove the AWS secrets from the repo (I'm sure you're aware, this is more a note for me).
- Since we're cleaning up we should also remove [`pkg/foundation/analytics`](https://github.com/ConduitIO/conduit/tree/maha/cleanup-old-plugins/pkg/foundation/analytics), this package isn't being used.
- We should change these [two](https://github.com/ConduitIO/conduit/blob/d4d4b4e61e4cd17b02fc517860555c32aaa69b17/pkg/connector/service_test.go#L136) [lines](https://github.com/ConduitIO/conduit/blob/d4d4b4e61e4cd17b02fc517860555c32aaa69b17/pkg/connector/service_test.go#L172) to `builtin:file` instead, this shouldn't affect the test.

Otherwise looks great, I love this PR ❤️

PR #326, opened 2022-03-25
Design document: OpenCDC

lovromazgon:
This PR contains the [design document](https://github.com/ConduitIO/conduit/blob/lovro/dd-opencdc/docs/design-documents/20220309-opencdc.md) for the OpenCDC format.

You can see it [here](https://github.com/ConduitIO/conduit/blob/lovro/dd-opencdc/docs/design-documents/20220309-opencdc.md).


docs/design-documents/20220309-opencdc.md
```diff
@@ -0,0 +1,349 @@
...
+- How transforms change the associated schema?
+- How can a schema be attached to a raw record?
+- How can a schema be extracted to a schema registry?
+
+### Questions
+
+- **Should we have a separate field announcing the record was created as part of a snapshot?**
+  - No, this information will be stored in the action (insert, update, delete, snapshot).
+- **Should the record support sending a schema only (e.g. we transmit a CREATE TABLE statement)? This might be useful for setting up the target structure even if the source is still empty.**
+  - No, for now we ignore this case.
```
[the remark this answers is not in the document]
dylanlott: After some more thought, I think it boils down to how we want to send schema changes over the wire. Would they get their own message followed by records that adhere to that schema update, or would schema changes be sent along with the record that adheres to them and the reader would have to detect and apply the schema changes?

docs/design-documents/20220309-opencdc.md
```diff
@@ -0,0 +1,349 @@
...
+- We are Debezium compatible out of the box.
+- When people work with our internal Record it makes more sense because it is actually the same as an OpenCDC record.
+- If our record is marshalled into JSON (or avro or proto) it's OpenCDC compatible by default.
+
+#### Cons
+
+- Breaking change for connectors, the record will change.
+- We are tied to the Debezium format (less flexibility).
+- The Debezium format contains a schema describing the record itself (i.e. fields that are part of the format itself), it is redundant.
+- The Debezium format is taylored to databases, we could be dealing with any source that is not a database (schemaless raw data, schemaless structured data etc.).
```
dylanlott, line 330: nit: `taylored` -> `tailored`

docs/design-documents/20220309-opencdc.md
```diff
@@ -0,0 +1,349 @@
...
+#### Pros
+
+- We are Debezium compatible out of the box.
+- When people work with our internal Record it makes more sense because it is actually the same as an OpenCDC record.
+- If our record is marshalled into JSON (or avro or proto) it's OpenCDC compatible by default.
+
+#### Cons
+
+- Breaking change for connectors, the record will change.
+- We are tied to the Debezium format (less flexibility).
```
dylanlott, line 328: This feels like a pretty huge con to me. Again, though, I think this comes down to our philosophy for transformers. Are they meant to be minimal data alteration tools, or are they meant to be very capable data massaging and sculpting tools?

docs/design-documents/20220309-opencdc.md
```diff
@@ -0,0 +1,349 @@
...
+### Goals
+
+The main goal is to ensure compatibility between any combination of source/destination connectors.
+
+Define a format that fulfills the following criteria:
+- A record can be associated to a schema that describes the payload or key
+- The paylod/key schema can be included in the record or pointing to a schema stored in a schema registry
+- Be able to represent an inserted, updated and deleted record
+- Be able to represent a record created in a snapshot
+- Support for splitting big records into multiple parts
```
dylanlott, line 14: I don't see this mentioned again in this document. Is there more to this or do you consider this achievable with just metadata and record payloads?

docs/design-documents/20220309-opencdc.md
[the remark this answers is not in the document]
ahmeroxa: Emitting schemas in the form of DDL statements (e.g. a `CREATE TABLE` statement) only makes sense in the case where you're doing straight replication between identical resources. 

As soon as you mix connectors and/or introduce transforms this functionality becomes useless.

I think what we actually want is the ability to detect a schema from a given Source and _register_ it with a schema registry somewhere. This would then allow the destination to preemptively query that registry and create the table structure in advance, however this also falls apart if a transform is introduced.

That is to say, all of the above leads me to think that actually the lazy schema lookup/integration is probably the most sensible approach. E.g. when a destination sees a record with a schema (ID or version) that differs from what it currently has set, then it can _infer_ the DDL to support the new schema.

docs/design-documents/20220309-opencdc.md
lovromazgon: I forgot to mention it in the standard metadata chapter, I was thinking of having some standard metadata field to denote multipart records. I'll add this!

docs/design-documents/20220309-opencdc.md
lovromazgon: Depends on what you mean by "very capable" 😄  I imagine transforms should let you alter the OpenCDC record in any way you want, the only constraint is that the output of the transform is still an OpenCDC record.

docs/design-documents/20220309-opencdc.md
```diff
@@ -0,0 +1,349 @@
...
+
+- We have full flexibility in defining a record that matches our needs.
+- When people work with our internal Record it makes more sense because it is actually the same as an OpenCDC record.
+- If our record is marshalled into JSON (or avro or proto) it's OpenCDC compatible by default.
+
+#### Cons
+
+- We define a new format that nobody is using but us. https://xkcd.com/927/
+- Breaking change for connectors, the record will change.
+- Debezium compatibility has to be achieved through transforms.
```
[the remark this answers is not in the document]
lovromazgon: I'm not sure how we'd do that in Conduit, Conduit only knows how to work with our internal record (which will match the OpenCDC structure), it's the responsibility of the connector to format the record in a way that makes sense for the output. Outputting a Debezium record sounds like a formatting issue, so if a connector needs this functionality it needs to provide it. If we see that this is something a lot of connectors need, we could provide a utility in the SDK for transforming an `sdk.Record` into a Debezium record.

docs/design-documents/20220309-opencdc.md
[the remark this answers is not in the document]
lovromazgon: > when a destination sees a record with a schema (ID or version) that differs from what it currently has set, then it can infer the DDL to support the new schema.

The record design includes `before` and `after` data, possibly even schemas for both, so this should be doable. We may want to provide utilities for detecting schema changes in the SDK to make it easier for the connector developer.

docs/design-documents/20220309-opencdc.md
```diff
@@ -0,0 +1,350 @@
...
+
+## Introduction
+
+### Goals
+
+The main goal is to ensure compatibility between any combination of source/destination connectors.
+
+Define a format that fulfills the following criteria:
+- A record can be associated to a schema that describes the payload or key
+- The paylod/key schema can be included in the record or pointing to a schema stored in a schema registry
```
maha-hajja, line 11: typo: paylod>payload

docs/design-documents/20220309-opencdc.md
```diff
@@ -0,0 +1,350 @@
...
+
+### Background
+
+Here are a couple of scenarios that we want to support. Records in these examples are formatted in the format defined in implementation option 1.
+
+#### Scenario 1: Connector ships raw data without schema
+
+In this scenario we have a pipeline that reads from a schemaless source (for example Kafka). The key and payload that is read is raw, the connector isn't aware of what the structure is so it creates an OpenCDC record with the raw key, raw payload and some metadata and sends it to Conduit for further processing.
+
+The record that Conduit receives could look somthing like this (note that the record is an internal Go struct, the JSON representation is strictly for illustration purposes):
```
maha-hajja, line 41: typo: somthing>something

docs/design-documents/20220309-opencdc.md
```diff
@@ -0,0 +1,350 @@
...
+  }
+}
+```
+
+#### Scenario 3: Connector ships raw data with schema
+
+This scenario assumes the connector provides raw data but also provides a schema for parsing the raw data and using it as structured data (for example a connector dealing with protobuf data). The reason behind this might be that the raw data can be compressed better and would improve the performance of the pipeline.
+
+
+Example record that Conduit recieves in this case:
```
maha-hajja, line 106:
```suggestion
Example record that Conduit receives in this case:
```

docs/design-documents/20220309-opencdc.md
```diff
@@ -0,0 +1,350 @@
...
+    }
+  }
+}
+```
+
+#### Scenario 4: Connector ships structured data with schema
+
+In this scenario we have a pipeline that reads from a source that contains structured data (for example Postgres). The connector is aware of the structure so it extracts the schema and structures the data into an OpenCDC record that gets sent to Conduit for further processing.
+
+Example record that Conduit recieves in this case:
```
maha-hajja, line 134:
```suggestion
Example record that Conduit receives in this case:
```

docs/design-documents/20220309-opencdc.md
```diff
@@ -0,0 +1,350 @@
...
+    }
+  }
+}
+```
+
+#### Scenario 2: Connector ships structured data without schema
+
+A connector could read from a source that supplies structured data, but has no idea about the schema of that structured data (for example a connector exposing a webhook accepting any JSON request). The connector supplies an OpenCDC record with a structured payload and/or key and sends it to Conduit for further processing. Note that the connector could infer the schema from the structured payload, but it does not need to do that since Conduit could do the same thing.
+
+Example record that Conduit recieves in this case:
```
maha-hajja, line 73: typo: receives

docs/design-documents/20220309-opencdc.md
```diff
@@ -0,0 +1,350 @@
...
+- The Debezium format is tailored to databases, we could be dealing with any source that is not a database (schemaless raw data, schemaless structured data etc.).
+
+### Option 3
+
+We define OpenCDC by starting with our current record structure and extending it with additional fields to keep it backwards compatible.
+
+#### Pros
+
+- Backwards compatibility.
+- We are somewhat flexibile in defining something that works for us.
```
maha-hajja, line 340:
```suggestion
- We are somewhat flexible in defining something that works for us.
```

maha-hajja (commented):
only nitpicks

docs/design-documents/20220309-opencdc.md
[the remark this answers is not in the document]
dylanlott: Since keys and payloads in both `before` and `after` have schemas, we have a complete picture of a record's change at any point in time. This leads to the question of how diffs should be represented and handled but I think that's out of the scope of this document.

I do agree that the lazy lookup / registration approach maximizes flexibility, and I think that's what we want for Conduit's many-to-many ecosystem.

>  Emitting schemas in the form of DDL statements (e.g. a CREATE TABLE statement) only makes sense in the case where you're doing straight replication between identical resources. 

Additionally, adding command language or anything similar in records will spiral out of control in complexity. I would say we should avoid any approach that would have us sending DDL statements over the wire.

docs/design-documents/20220309-opencdc.md
```diff
@@ -0,0 +1,350 @@
...
+-	// a destination Payload will never be null.
+-	Payload Data
++	// Before holds the key and payload data that was valid before the change.
++	// This field is only populated if the operation is update or delete.
++	Before struct {
++		// Key contains the data and schema of the key before the change.
++		Key DataWithSchema
++		// Payload contains the data and schema of the payload before the change.
++		Payload DataWithSchema
++	}
```
hariso, line 253: I'm not myself sure about this, but I'll throw it to hear opinions: do we need metadata in the before and after structs? Metadata about the payload that is. For example, if we have an S3 source, I imagine the payload will be an S3 file itself. However, the S3 object also has metadata, which can change and in which we might be interested.

lovromazgon: Interesting thought, but I would argue that this kind of metadata is part of the payload. Given that we don't prescribe any specific structure for the payload itself the connector is welcome to create a metadata field in the payload itself, then changes to that metadata can be tracked if needed.

hariso: Yup, that's true, keeping in the payload is definitely an option and has its pros for sure. A con though, is that it requires a "mini schema" in the connector to package the data and metadata togeher. It also adds to processing time for, for example, binary there, since it needs to be encoded.

hariso: Before I say anything, this is the only thing (and it's minor) which I had for a comment. I like the rest of the OpenCDC, and AFAIC we can merge it. :muscle:  I don't think this is even close to being a blocker.

I too don't see many use cases either, which is a strong argument for not including it, and which is why I don't feel strongly about this. On the other hand, it's easier to remove an unused field, than to have client code work around the limitation of not having it.

What we would get with metadata in `before` and `after` is basically support for "context" changes, and not only changes in data objects themselves. 

> but any code dealing with the payload or key would need to somehow deal with the metadata as well

Hmm, why so? We may not be thinking about the same thing, but I'm thinking about a specific connector's code which is free to use or just ignore this field (in the same way it will ignore the `before` field in a number of cases). And when it comes to the code responsible for serialization/deserialization, then it would be just one more field.

> It would also create strange expectations 

Do we want to create expectations here about that all though? This document is detailed when it comes to how changes are described (kudos for that!), but we may not have to go into what triggers an update. That sounds like something source connectors should decide.

lovromazgon: I won't see this comment as a blocker then and merge the PR, but I don't want to leave you hanging so I will at least respond before doing so 🙂 

---

First, seems like we agree that such a field would have few use cases. Where we might have different opinions is if having such a field would create more problems or solve more problems. I'm thinking it would create more problems. Why? Complexity and expectations.

Complexity - here I mean mostly higher cognitive load for the developer. Having a rarely used field doesn't mean it can be ignored, it probably exists for a reason, some connectors will use it and will expect that metadata to have an effect. A developer will have to think about what to do with that field, decide if it's safe to ignore it, figure out what the difference is between record metadata and payload metadata. Unless the field would have a clear use case and a clearly defined purpose separating it from the record metadata, it could create more confusion.

Expectations - besides the expectations I mentioned in my previous comment I want to point out that having a field that can be safely ignored could create false expectations in the behavior across connectors. Say a connector puts some data in this metadata field while another connector ignores this field, this could mean they aren't as compatible as we would expect. If the metadata can be safely ignored it can be put into the record metadata. If it shouldn't be ignored it can be part of the payload. That are clear expectations. On the other hand, wondering if a field will be present in the record metadata or in the payload metadata and wondering how behavior will change in either case, that aren't clear expectations.

To sum up, I'd say the main reason for my opinion is the lack of a clear enough distinction between "record metadata", "payload metadata" and "payload data". The first one is metadata, that can safely be ignored, there is no expectation of the connector using these fields, though it can use it to alter its behavior (I would compare it to HTTP headers or gRPC metadata). The payload on the other hand is the "meat of the record", that's the content that caused for the record to be created in the first place. Adding "payload metadata" would put it somewhere in between and I just don't see how it would improve the user experience.

simonl2002 (approved):
lgtm

Issue #145, opened 2022-01-28
UI Improvement: do not allow a customer to create transform/edit pipeline name while it is running


brettgoulder:
### Feature description

We should prompt the customer earlier when they cannot edit a pipeline due to it running.

At the moment, if a pipeline is running, you can click the button to add a transform, or edit the pipeline name, but it will throw an error when you actually try to finalize it. We should disable the button and notify the user sooner.

PR #338, opened 2022-03-29
Add pipeline running validations

jmar910:
### Description
Add a pipeline running validation to prevent the user from updating transforms or editing pipeline settings while the pipeline is running

Fixes https://github.com/ConduitIO/conduit/issues/145

### Screenshots
Screen Shot 2022-04-01 at 1 58 31 PM
Screen Shot 2022-04-01 at 1 58 35 PM
Screen Shot 2022-04-01 at 1 58 39 PM

### Quick checks:

### DIFF
--- a/ui/config/environment.js
+++ b/ui/config/environment.js
@@ -31,7 +31,7 @@ module.exports = function (environment) {
 
   if (environment === 'development') {
     ENV['ember-cli-mirage'] = {
-      enabled: true,
+      enabled: ENV.isDevMirageEnabled,
     };
 
     // ENV.APP.LOG_RESOLVER = true;
--- a/ui/tests/acceptance/pipeline/index-test.js
+++ b/ui/tests/acceptance/pipeline/index-test.js
@@ -21,6 +21,8 @@ const page = {
     '[data-test-connector-column="destination"] [data-test-connector-node]',
 
   pipelineAddNewNodeButton: '[data-test-pipeline-editor-add-node]',
+  pipelineAddNewNodeDisabledButton:
+    '[data-test-pipeline-editor-add-node-disabled]',
   pipelineAddNewSourceButton: '[data-test-pipeline-editor-add-node-source]',
 
   pipelineStatus: '[data-test-pipeline-status-label]',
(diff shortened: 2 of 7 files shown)


lovromazgon (commented):
Looks good, how hard would it be adding the same check for editing a connector? Right now I can get to the edit connector popup and press "update" but nothing happens.

EDIT: Actually I see a popup warning that the pipeline is running. Something strange happens if I try to delete the connector first though, then the update doesn't do anything anymore and looks like it's broken.

jmar910:
@lovromazgon resolved!

ui/config/environment.js
```diff
@@ -31,7 +31,7 @@ module.exports = function (environment) {
 
   if (environment === 'development') {
     ENV['ember-cli-mirage'] = {
-      enabled: true,
+      enabled: ENV.isDevMirageEnabled,
```
jayjayjpg, line 34: Does this require any documentation or do most users run the app directly against the conduit API anyways?

ui/tests/acceptance/pipeline/index-test.js
```diff
@@ -327,12 +339,34 @@ module('Acceptance | pipeline/index', function (hooks) {
...
-    test('does not show a degraded button', async function () {
+    test('does not show a degraded button', function () {
       assert.dom(page.pipelineDegradedButton).doesNotExist();
     });
+
+    test('it disables adding connectors', function () {
+      assert.dom(page.pipelineAddNewNodeDisabledButton).exists();
+    });
+
+    test('it disables updating transforms', async function () {
```
jayjayjpg, line 358: Neat! :sparkles:

jayjayjpg (approved):
Awesome work as always, this reads great! :sparkles:

ui/config/environment.js
jayjayjpg: Not a blocker, just curious by the way! :-)

jmar910: No documentation required. We originally ran it in passthru mode anyway by default, so this is the same thing, but just fully disables mirage. Most if not all users should be running the app directly against the conduit API

jayjayjpg: Neat and thank you for the outline!

PR #349, opened 2022-04-04
add more swagger documentation

### LINKED ISSUE #312: Documentation: Add request body descriptions to Swagger API definition.
### Feature description

Adding more descriptions would help users understand the requirements and constraints of the API request body. Here is the information that's currently provided:

```json
{
  "type": "TYPE_UNSPECIFIED",
  "plugin": "string",
  "pipelineId": "string",
  "config": {
    "name": "string",
    "settings": {
      "additionalProp1": "string",
      "additionalProp2": "string",
      "additionalProp3": "string"
    }
  }
}
```

Swagger Docs: https://swagger.io/specification/?sbsearch=Description%20Property

maha-hajja:
### Description

Fixes #312 

### Quick checks:


proto/api/v1/api.proto
```diff
@@ -543,7 +559,9 @@ service ConnectorService {
 
 message CreateConnectorRequest {
   Connector.Type type = 1;
+  // Plugin name
```
lovromazgon, line 562: We need more info here about builtin vs standalone plugins - this can be either an absolute path to a standalone plugin or the name of the builtin plugin.

proto/api/v1/api.proto
```diff
@@ -543,7 +559,9 @@ service ConnectorService {
...
   string plugin = 2;
+  // Pipeline id of which the connector will be in
```
lovromazgon, line 564:
```suggestion
  // ID of the pipeline to which the connector will get attached.
```

lovromazgon (commented):
Can you please go through all the existing comments and make sure they start with a capital letter and end with a dot? 🙏

proto/api/v1/api.proto
maha-hajja: should talk about the difference between them? or just what the name would be in each case?

lovromazgon (approved):
LGTM, we need to add even more docs in the future but this is already a step in the right direction :+1:

PR #358, opened 2022-04-06
Ensure acker node stops correctly

lovromazgon:
### Description

This fixes the issue of stopping an empty pipeline that didn't process any messages.

### Quick checks:

### DIFF
--- a/pkg/pipeline/stream/acker.go
+++ b/pkg/pipeline/stream/acker.go
@@ -76,7 +76,7 @@ func (n *AckerNode) Run(ctx context.Context) (err error) {
 
 	select {
 	case <-ctx.Done():
-		return nil
+		return ctx.Err()
 	case <-n.start:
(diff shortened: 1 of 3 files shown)


pkg/pipeline/stream/acker.go
```diff
@@ -188,6 +188,12 @@ func (n *AckerNode) ForgetAndDrop(msg *Message) {
 // Wait can be used to wait for the count of outstanding acks to drop to 0 or
 // the context gets canceled.
 func (n *AckerNode) Wait(ctx context.Context) {
+	// in case ExpectAck wasn't called we call it here
+	n.startOnce.Do(func() {
+		n.init()
+		close(n.start)
+	})
```
lovromazgon, line 195: The issue was the node was stuck in the select statement in line 77, now we ensure it continues on there even if no messages flowed through the pipeline.

PR #360, opened 2022-04-07
warn user if Go version is less than required

### LINKED ISSUE #151: warn user if min version of Go required is not installed
### Feature description

As someone building Go using `make`, I'd like `make` (i.e. all build targets) to warn me if the version of Go I have installed is less than the one required by Conduit.

The ReadMe section about building from source is pretty clear about the requirements, but I believe this feature request still have some value because:
1. some people never the read the manual
2. some people work may work with multiple versions of Go

maha-hajja:
### Description

* added a script to check if go version is less than required

Fixes #151 

### Quick checks:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,9 +1,11 @@
 .PHONY: test test-integration build run proto-api proto-lint clean download install-tools generate
 
 VERSION=`./scripts/get-tag.sh`
+GO_VERSION_CHECK=`./scripts/check-go-version.sh`
 
 # The build target should stay at the top since we want it to be the default target.
 build: pkg/web/ui/dist
+	@echo ${GO_VERSION_CHECK}
 	go build -ldflags "-X 'github.com/conduitio/conduit/pkg/conduit.version=${VERSION}'" -o conduit -tags ui ./cmd/conduit/main.go
 	@echo "\nBuild complete. Enjoy using Conduit!"
 	@echo "Get started by running:"
@@ -20,6 +22,7 @@ test-integration:
 		exit $$ret
 
 build-server:
+	@echo ${GO_VERSION_CHECK}
 	go build -ldflags "-X 'github.com/conduitio/conduit/pkg/conduit.version=${VERSION}'" -o conduit ./cmd/conduit/main.go
 	@echo "build version: ${VERSION}"
 
--- a/Makefile
+++ b/Makefile
@@ -62,6 +62,6 @@ ui-%:
 
 check-go-version:
 	@if [ "${GO_VERSION_CHECK}" != "" ]; then\
-        echo "${GO_VERSION_CHECK}";\
-        exit 1;\
-    fi
+		echo "${GO_VERSION_CHECK}";\
+		exit 1;\
+	fi
--- a/Makefile
+++ b/Makefile
@@ -1,11 +1,10 @@
-.PHONY: test test-integration build run proto-api proto-lint clean download install-tools generate
+.PHONY: test test-integration build run proto-api proto-lint clean download install-tools generate check-go-version
 
 VERSION=`./scripts/get-tag.sh`
 GO_VERSION_CHECK=`./scripts/check-go-version.sh`
 
 # The build target should stay at the top since we want it to be the default target.
-build: pkg/web/ui/dist
-	@echo ${GO_VERSION_CHECK}
+build: check-go-version pkg/web/ui/dist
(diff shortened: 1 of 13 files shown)


hariso:
This looks good, and provides additional useful info. However, #151 was originally about the `make` command, because if the Go version is not supported, than most likely we won't even be able to run Conduit (so it can print the warning).

maha-hajja:
@hariso oh makes sense! I'll mark this as a draft and check another solution then. thank you

hariso:
> @hariso oh makes sense! I'll mark this as a draft and check another solution then. thank you

You're welcome. :+1:

scripts/check-go-version.sh
```diff
@@ -0,0 +1,16 @@
+#!/bin/bash
+
+# checks if the installed Go version is less than the required version.
+# returns a warning message if it's less than required. Otherwise, returns "Building Conduit..."
+
+function version_lt() { test "$(echo "$@" | tr " " "\n" | sort -rV | head -n 1)" != "$1"; }
```
hariso, line 6: :+1:

Makefile
```diff
@@ -1,9 +1,11 @@
 .PHONY: test test-integration build run proto-api proto-lint clean download install-tools generate
 
 VERSION=`./scripts/get-tag.sh`
+GO_VERSION_CHECK=`./scripts/check-go-version.sh`
 
 # The build target should stay at the top since we want it to be the default target.
 build: pkg/web/ui/dist
+	@echo ${GO_VERSION_CHECK}
```
hariso, line 8: Reporting the version is great, but it would be good if we could fail the build.

maha-hajja: should we? conduit can be built using go1.17 now because we're not using generics right?

hariso: That's a good point. In that case, though, we shouldn't say that 1.18 is the minimum required version, if things work with 1.17?

scripts/check-go-version.sh
```diff
@@ -0,0 +1,16 @@
...
+
+GO_VERSION=$(go version | { read -r _ _ v _; echo "${v#go}"; })
+# needs to be changed each time conduit Go version is upgraded
+REQUIRED_GO_VERSION="1.18"
+
+if version_lt "$GO_VERSION" "$REQUIRED_GO_VERSION"; then
+    echo "WARNING: min Go version required is go$REQUIRED_GO_VERSION, version installed is go$GO_VERSION"
+else
+    echo "Building Conduit..."
```
lovromazgon, line 15: I would skip the `else` part to make the script a bit more general (we might use it in other places than building Conduit in the future).

maha-hajja: I added it so that it wouldn't just print an empty line before the build command.. it looked weird

maha-hajja: but yeah didn't mind it that much.. will take it out

Makefile
hariso: That sounds good to me. :+1:

lovromazgon: Wait a sec, even though it can be built with 1.17 right now that might break the first time we use a 1.18 feature. Our local versions are all 1.18 and our CI jobs use 1.18, we can only "guarantee" the build will work if you use 1.18, so I would side towards being strict and enforce 1.18.

hariso: Personally, I don't mind the strictness if it's generally easy to switch between Go versions (I always use only one version). Otherwise, given that 1.18 is pretty new, many projects are still likely not using, and so likely many developers are not using it either.

lovromazgon: I'd have the same opinion if we were writing a library and the main use of the code would be for other people to use it as a dependency. But since we are writing a tool where the main use case is for a person to download a pre-built binary or the Docker image I see little harm in using the latest version that might not be widely adopted yet, at least from the end-user's point of view. And again, since we can only guarantee that Conduit works as expected if the developer uses the expected tooling, I think we should be strict about it.

hariso: I'm also thinking that the dev community (i.e. our contributors) could have troubles with it, but on one hand there aren't many right now, and together with the point about the main use case (prebuilt binary), being strict here is fine.

Makefile
```diff
@@ -62,3 +60,8 @@ pkg/web/ui/dist:
 ui-%:
 	@cd ui && make $*
 
+check-go-version:
+	@if [ "${GO_VERSION_CHECK}" != "" ]; then\
+        echo "${GO_VERSION_CHECK}";\
+        exit 1;\
+    fi
```
lovromazgon, line 67: 🤖  we should use tabs instead of spaces in the Makefile

lovromazgon (approved):
LGTM 👍  Just a linter comment 🤖

PR #361, opened 2022-04-07
docs: Instructions to get config. params

hariso:
### Description
This PR adds instructions to get configuration parameters for Conduit.

### Quick checks:

### DIFF
--- a/README.md
+++ b/README.md
@@ -38,7 +38,7 @@ Conduit was created and open-sourced by [Meroxa](https://meroxa.io).
 
 ## Installation guide
 
-### Download release
+### Download and run release
 
 Download a pre-built binary from the [latest release](https://github.com/conduitio/conduit/releases/latest) and simply
 run it!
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -33,6 +33,7 @@ Conduit was created and open-sourced by [Meroxa](https://meroxa.io).
 - [API](#api)
 - [UI](#ui)
 - [Documentation](#documentation)
+- [Configuration](#configuration)
```
lovromazgon, line 36: This is misplaced, it should be above `Documentation`.

Thinking out loud here - does this deserve its own top-level header? Especially since it's only 1 sentence right now, but also because this list is getting a bit crowded IMO. What if we add that sentence in the `Download release` chapter where we explain how to run Conduit? We can hint that they can check the help to see config options.

hariso: Agreed. :+1:

lovromazgon (approved):

PR #363, opened 2022-04-07
Fix acker node test and nil pointer

lovromazgon:
### Description

This [failed run](https://github.com/ConduitIO/conduit/actions/runs/2110239882/attempts/1) exposed a race condition in the test as well as an interface conversion error if the message is somehow not in the acker cache.

### Quick checks:


dylanlott (approved):
That's a subtle change in LoadAndDelete, good catch!

hariso (approved):
Good stuff. :+1:

PR #366, opened 2022-04-08
Catch panic in built-in connectors

### LINKED ISSUE #220: Plugin: Catch panics in built-in connectors
### Feature description

Built-in connector plugins run in the same process as Conduit, a badly implemented plugin could potentially bring down the whole process with a panic. We need to make sure built-in plugins run in some kind of a sandbox where we catch all panics and transform them into an error instead.

Standalone plugins should not be affected by this since panics should be handled by the framework, but we need to test this out to be sure.

lovromazgon:
### Description

This PR makes sure explicit calls to builtin plugins can't crash Conduit. If the plugin spawns a goroutine that panics it can still bring down Conduit, there's no way to catch those panics (if there is, let me know!).

First use of generics in Conduit! 😁 

Fixes #220

### Quick checks:

### DIFF
--- a/pkg/plugin/builtin/v1/source.go
+++ b/pkg/plugin/builtin/v1/source.go
@@ -65,7 +65,7 @@ func (s *sourcePluginAdapter) Configure(ctx context.Context, cfg map[string]stri
 		return err
 	}
 	s.logger.Trace(ctx).Msg("calling Configure")
-	resp, err := s.impl.Configure(s.withLogger(ctx), req)
+	resp, err := runSandbox(s.impl.Configure, s.withLogger(ctx), req)
 	if err != nil {
 		// TODO create new errors as strings, this happens when they are
 		//  transmitted through gRPC and we don't want to falsely rely on types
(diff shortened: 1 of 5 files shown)


pkg/plugin/builtin/v1/source.go
```diff
@@ -96,7 +96,7 @@ func (s *sourcePluginAdapter) Start(ctx context.Context, p record.Position) erro
 	s.stream = newSourceRunStream(ctx)
 	go func() {
 		s.logger.Trace(ctx).Msg("calling Run")
-		err := s.impl.Run(s.withLogger(ctx), s.stream)
+		err := runSandboxNoResp(s.impl.Run, s.withLogger(ctx), cpluginv1.SourceRunStream(s.stream))
```
dylanlott, line 99: Do we need to sandbox the `SourceRunStream` too, like the comment for `DestinationRunStream` notes?

dylanlott (approved):
I was wondering when we'd see generics first used. Looks good to me. ✨

pkg/plugin/builtin/v1/source.go
lovromazgon: Which comment do you mean? I don't think we need to sandbox anything else, actually, I don't think we _can_ sandbox anything else. Conduit calls 5 functions in the plugin and those functions are all sandboxed now.

dylanlott: I thought there was a TODO comment about sandboxing the SourceRunStream, but I don't see it anymore so disregard 👍

PR #371, opened 2022-04-15
create and implement a 'validate' endpoint

### LINKED ISSUE #180: Feature: Endpoint for plugin config validation

maha-hajja:
### Description

added an endpoint that validates the connector configurations and returns an error with the problem or an empty response

Fixes #180 

### Quick checks:


pkg/web/api/connector_v1.go
```diff
@@ -145,3 +146,21 @@ func (c *ConnectorAPIv1) DeleteConnector(ctx context.Context, req *apiv1.DeleteC
 
 	return &apiv1.DeleteConnectorResponse{}, nil
 }
+
+// ValidateConnector hi
```
hariso, line 150: Hello, ValidateConnector!

maha-hajja: lol forgot about this one

pkg/orchestrator/connectors.go
```diff
@@ -163,3 +165,42 @@ func (c *ConnectorOrchestrator) Update(ctx context.Context, id string, config co
...
+	case connector.TypeSource:
+		src, _ := d.DispenseSource()
+		err := src.Configure(ctx, config.Settings)
+		if err != nil {
+			return cerrors.Errorf("configurations failed: %w", err)
+		}
+		err = src.Teardown(ctx)
+		if err != nil {
+			return cerrors.Errorf("couldn't teardown the connector: %w", err)
+		}
```
hariso, line 190: I'm wondering, does a client care if the teardown didn't succeed? It's more like a cleanup type of thing, which we can turn a blind eye to.

pkg/web/openapi/swagger-ui/api/v1/api.swagger.json
```diff
@@ -155,6 +155,45 @@
...
+      "post": {
+        "operationId": "ConnectorService_ValidateConnector",
+        "responses": {
+          "200": {
+            "description": "A successful response.",
+            "schema": {
+              "$ref": "#/definitions/v1ValidateConnectorResponse"
+            }
+          },
+          "500": {
```
hariso, line 168: WDYT about returning something other than 500? This feels like a client error, not a server-side one.

proto/api/v1/api.proto
```diff
@@ -570,6 +576,17 @@ message CreateConnectorResponse {
...
+message ValidateConnectorRequest {
+  Connector.Type type = 1;
+  // Plugin name is the name of the builtin plugin (builtin:name), or the absolute path of a standalone plugin.
+  string plugin = 2;
+  // Configurations for the connector to get validated
+  Connector.Config config = 4;
+}
+
+message ValidateConnectorResponse {
+}
```
hariso, line 588: Do we need an error here?

pkg/web/api/connector_v1_test.go
```diff
@@ -284,6 +284,39 @@ func TestConnectorAPIv1_DeleteConnector(t *testing.T) {
...
+	ctrl := gomock.NewController(t)
+	csMock := apimock.NewConnectorOrchestrator(ctrl)
+	api := NewConnectorAPIv1(csMock)
+	connBuilder := connmock.Builder{Ctrl: ctrl}
+
+	source := connBuilder.NewSourceMock(uuid.NewString(), connector.Config{
+		Name:     "A source connector",
+		Settings: map[string]string{"path": "path/to"},
+		Plugin:   "builtin:file",
+	})
```
hariso, line 298: Do we actually need to build the source? It would feel a bit more natural without building, since we will always first validate a config, and then build a source. And also, it looks like it acts more like a container: we only use it to get the type and config. In other words, it might make sense to just declare the name, settings and plugin variables and use them directly.

pkg/web/api/connector_v1_test.go
```diff
@@ -284,6 +284,39 @@ func TestConnectorAPIv1_DeleteConnector(t *testing.T) {
 	assert.Equal(t, want, got)
 }
 
+func TestConnectorAPIv1_ValidateConnector(t *testing.T) {
```
hariso, line 287: It would be nice to have a test case for a validation error.

proto/api/v1/api.proto
maha-hajja: at first, I had an error here, but then decided to go with returning the error, which would look like:
`{
    "code": 13,
    "message": "configurations failed: \"path\" config value must be set",
    "details": []
}` 
but If having the error as a string in the response makes more sense, then I can go with that.. WDYT?

pkg/web/openapi/swagger-ui/api/v1/api.swagger.json
lovromazgon: That would make sense, the problem is we can't differentiate a validation error from any other error right now, the plugin `Configure` method either succeeds or fails. We need to first generally define the mechanism how we could make Conduit aware of specific plugin errors (e.g. marshal errors into a certain format and include an error code or something).

lovromazgon: What we could do is just return 400 (and the corresponding gRPC error code) in case the validation failed.

proto/api/v1/api.proto
hariso: Given that, it's fine IMO. :+1:

pkg/orchestrator/connectors.go
```diff
@@ -163,3 +166,42 @@ func (c *ConnectorOrchestrator) Update(ctx context.Context, id string, config co
 	r.Skip()
 	return conn, nil
 }
+
+func (c *ConnectorOrchestrator) Validate(
```
lovromazgon, line 170: I think it would make sense to move the logic in this function into a function in the plugin service. The orchestration layer should ideally be as dumb as possible and just deal with transactions and orchestration, the logic right now is too aware of implementation details for my taste.

Or maybe we can have two functions in the plugin service (e.g. along the lines of `ValidateSourceConfig` and `ValidateDestinationConfig`), then the plugin service doesn't have to be aware of `connector.Type`.

pkg/plugin/errors.go
```diff
@@ -14,10 +14,39 @@
...
+)
 
 var (
 	ErrStreamNotOpen    = cerrors.New("stream not open")
 	ErrPluginRunning    = cerrors.New("plugin is running")
 	ErrPluginNotRunning = cerrors.New("plugin is not running")
 )
+
+type ValidationError struct {
+	Err error
```
lovromazgon, line 30: Nitpick: I wouldn't expose a field unless there's a reason, and here I don't see one. The underlying error can be accessed by `Unwrap` already.

pkg/web/api/connector_v1.go
```diff
@@ -145,3 +146,22 @@ func (c *ConnectorAPIv1) DeleteConnector(ctx context.Context, req *apiv1.DeleteC
...
+	req *apiv1.ValidateConnectorRequest,
+) (*apiv1.ValidateConnectorResponse, error) {
+	err := c.cs.Validate(
+		ctx,
+		fromproto.ConnectorType(req.Type),
+		fromproto.ConnectorConfig(req.Config, req.Plugin, "", nil),
+	)
+
+	if err != nil {
+		return nil, status.PluginError(err)
```
lovromazgon, line 163: This should be `status.ConnectorError`, it is the connector API after all.

proto/api/v1/api.proto
```diff
@@ -481,6 +481,12 @@ service ConnectorService {
       }
     };
   };
+  rpc ValidateConnector(ValidateConnectorRequest) returns (ValidateConnectorResponse) {
+    option (google.api.http) = {
```
lovromazgon, line 485: We need to add a `grpc.gateway.protoc_gen_openapiv2.options.openapiv2_operation` option and specify that the response can contain the http status 400.

pkg/plugin/errors.go
maha-hajja: I needed to expose it because I needed to assign the error when creating it in the "Validate" method.

> if err != nil {
> 			return &plugin.ValidationError{Err: err}
> 		}

pkg/orchestrator/connectors.go
maha-hajja: I tried adding this logic in the plugin service at first, but we would have to import the connector package, which would cause a circular dependency.
So I'll check the second approach

pkg/plugin/errors.go
lovromazgon: If we manage to pull the logic into the plugin service this shouldn't be an issue anymore.

pkg/orchestrator/connectors.go
```diff
@@ -163,3 +165,30 @@ func (c *ConnectorOrchestrator) Update(ctx context.Context, id string, config co
...
+	ctx context.Context,
+	t connector.Type,
+	config connector.Config,
+) error {
+	logger := log.CtxLogger{Logger: zerolog.Nop()}
+	d, err := c.plugins.NewDispenser(logger, config.Plugin)
```
maha-hajja, line 175: is it better to get the dispenser here, or in the plugin service?

lovromazgon: I'd say plugin service.

lovromazgon: I just realized you already did it this way, let's leave it like this.

maha-hajja: I did this way to minimize duplicated code

pkg/orchestrator/connectors.go
```diff
@@ -163,3 +165,30 @@ func (c *ConnectorOrchestrator) Update(ctx context.Context, id string, config co
 	r.Skip()
 	return conn, nil
 }
+
+func (c *ConnectorOrchestrator) Validate(
+	ctx context.Context,
+	t connector.Type,
+	config connector.Config,
+) error {
+	logger := log.CtxLogger{Logger: zerolog.Nop()}
```
lovromazgon, line 174: The last thing we need to do is add a logger to `orchestrator.base` so that orchestrators can log stuff, then we can also pass that logger to `NewDispenser`.

pkg/plugin/service.go
```diff
@@ -70,3 +70,45 @@ func (r *Service) List(ctx context.Context) (map[string]Specification, error) {
...
+	if err != nil {
+		return cerrors.Errorf("could not dispense source: %w", err)
+	}
+
+	defer func(src SourcePlugin, ctx context.Context) {
+		err := src.Teardown(ctx)
+		if err != nil {
+			return
+		}
+	}(src, ctx)
```
lovromazgon, line 85: We lose the error returned by `Teardown`. A way to solve it would be to create a named return and overwrite the returned error if we aren't returning an error already ([example](https://go.dev/play/p/LIg9tLSjPnB)).

Nitpick: we don't need to specify params for this func, it can access `src` and `ctx` already.
```suggestion
	defer func() {
		terr := src.Teardown(ctx)
		if err == nil {
			err = terr // only overwrite if we are not returning an error
		}
	}()
```

pkg/orchestrator/orchestrator.go
```diff
@@ -36,13 +36,15 @@ type Orchestrator struct {
...
 	db database.DB,
+	logger log.CtxLogger,
 	pipelines PipelineService,
 	connectors ConnectorService,
 	processors ProcessorService,
 	plugins PluginService,
 ) *Orchestrator {
 	b := base{
 		db:         db,
+		logger:     logger,
```
lovromazgon, line 47: We should set the component in the logger (e.g. `logger.WithComponent("orchestrator")`).

pkg/orchestrator/pipelines_test.go
```diff
@@ -35,7 +37,8 @@ func TestPipelineOrchestrator_Start_Success(t *testing.T) {
 		ID:     uuid.NewString(),
 		Status: pipeline.StatusSystemStopped,
 	}
-	orc := NewOrchestrator(db, plsMock, consMock, procsMock, pluginMock)
+	logger := log.InitLogger(zerolog.InfoLevel, log.FormatCLI)
```
lovromazgon, line 40: Tip: in tests we can use `log.Nop()`, since we don't care about the log output.

pkg/plugin/service.go
```diff
@@ -77,12 +77,12 @@ func (r *Service) ValidateSourceConfig(ctx context.Context, d Dispenser, setting
...
 	}
 
-	defer func(src SourcePlugin, ctx context.Context) {
-		err := src.Teardown(ctx)
-		if err != nil {
-			return
+	defer func() {
+		terr := src.Teardown(ctx)
+		if err == nil {
+			err = terr // only overwrite error if it's nil
```
lovromazgon, line 83: For this to work we need to make `err` a [named return value](https://go.dev/tour/basics/7).

lovromazgon (commented):
Last change request, I promise 🙂

PR #381, opened 2022-04-26
Add creation and modification timestamp to objects

### LINKED ISSUE #202: Feature: Add creation and modification timestamp to objects
### Feature description

This potentially collapses multiple feature requests together, but I'd like to see creation and modification timestamps
on the first class objects, like connectors, pipelines, processors etc.

This will help out-of-bound changes be detected from automation systems that persist some kind of state.

maha-hajja:
### Description

added CreatedAt and UpdatedAt timestamps for pipelines, connectors, and processors
* still working on the tests

Fixes #202 

### Quick checks:


proto/api/v1/api.proto
```diff
@@ -35,8 +36,11 @@ message Pipeline {
 
   // -- children connections ---------------------------------------------------
 
-  repeated string connector_ids = 5 [(google.api.field_behavior) = OUTPUT_ONLY];
-  repeated string processor_ids = 6 [(google.api.field_behavior) = OUTPUT_ONLY];
+  repeated string connector_ids = 4 [(google.api.field_behavior) = OUTPUT_ONLY];
+  repeated string processor_ids = 5 [(google.api.field_behavior) = OUTPUT_ONLY];
```
lovromazgon, line 40: As much as I'd also like to have numbers aligned, this would create a breaking change and anybody using the gRPC API would need to generate their client again. So for that reason, I think we shouldn't change the field numbers of existing fields, at least not now. Maybe we can clean them up before releasing v1.

pkg/connector/source.go
```diff
@@ -36,6 +36,10 @@ type source struct {
 	// logger is used for logging and is set when source is created.
 	logger log.CtxLogger
 
+	// timestamps
+	XCreatedAt time.Time
+	XUpdatedAt time.Time
```
lovromazgon, line 41: Nitpick - I'd define this above logger to group X prefixed fields.

pkg/connector/builder.go
```diff
@@ -72,10 +73,13 @@ func (b *DefaultBuilder) Init(c Connector, id string, config Config) error {
 		return cerrors.Errorf("could not create plugin %q: %w", config.Plugin, err)
 	}
 
+	tn := time.Now()
```
lovromazgon, line 76: This has to be moved to `Build`, otherwise we'll have issues when retrieving connectors from the store ([this function](https://github.com/ConduitIO/conduit/blob/7c852bff2541a1faee9b688e75f40d72c74e3a35/pkg/connector/store.go#L166)).

The store first calls `Build` to get the connector of the right type, then unmarshals the stored value and then calls `Init`. The stored value will contain `UpdatedAt` and `CreatedAt` so we should use those values and not create new ones in `Init`.

lovromazgon (commented):
I think we should also update `UpdatedAt` in the following functions, since they change the entity:
- `pipeline.Service.AddProcessor`
- `pipeline.Service.RemoveProcessor`
- `pipeline.Service.AddConnector`
- `pipeline.Service.RemoveConnector`
- `connector.Service.AddProcessor`
- `connector.Service.RemoveProcessor`

I'm aware they don't change the user-defined configuration, but they change the structure of the pipeline, reason enough to update the timestamp IMO.

pkg/connector/connector.go
```diff
@@ -42,6 +43,10 @@ type Connector interface {
 	Config() Config
 	SetConfig(Config)
 
+	CreatedAt() time.Time
+	UpdatedAt() time.Time
+	SetUpdatedAt(time.Time)
```
hariso, line 48: Getters and setters... Java flashbacks.: )

pkg/connector/service.go
```diff
@@ -147,6 +148,7 @@ func (s *Service) Update(ctx context.Context, id string, data Config) (Connector
 	}
 
 	conn.SetConfig(data)
+	conn.SetUpdatedAt(time.Now())
```
hariso, line 151: Maybe it's easier if we pull SetUpdatedAt into SetConfig.

lovromazgon: I almost posted the same comment in the previous review, but after giving it a second thought decided against it. My reasoning - given that the decision to update the `UpdatedAt` field is made in `pipeline.Service` and `processor.Service`, I would rather let `connector.Service` explicitly set it as well, to have some consistency between services.

hariso: Consistency is a good point. In this case, it appears to be consistency in implementation details of independent services, not about behavior or user experience, so at least IMHO it's still more valuable to pull it into SetConfig, because that has the value of making sure it always gets updated (and we're already calling it multiple times).

lovromazgon: I'm not convinced tbh, it's a question of "whose responsibility is it to detect an update" and I think it should be the services (at least that's what we do in other services). The fact that we have functions on the connector that could take over this responsibility is only a consequence of the design, which I already [mentioned](https://github.com/ConduitIO/conduit/pull/381#discussion_r864842440) is bad and should change.

hariso: At least to me, it's the object itself which knows most about itself, including these changes and which should track its own metadata. It might help with race conditions too, and making sure invalid timestamps are not used. That would of course mandate using functions to update fields, so if that's not Go-idiomatic, it makes it less desirable as a solution. However, I certainly don't think this is a blocker.

lovromazgon: I understand your view and if we were doing OOP I would be on your side! I think the difference in opinions stems from how we think about the connector - you just called it an "object", while I think of it as more of a "dumb struct" that carries data and can be manipulated by other "objects". Of course, my view of the connector breaks down once the pipeline starts running because the "dumb struct" then manages its own state 🙃  So my view depends on the context, and once again it boils down to bad design (I am to blame for that).

Maybe we could make a better distinction between the connector "data struct" (only carries fields that are persisted) and a "running connector" (contains logic for communicating with the plugin), which would make the responsibilities clearer. I really hope we will have some time before the end of the milestone to refactor!

lovromazgon (approved):
LGTM

pkg/web/api/connector_v1_test.go
```diff
@@ -45,6 +47,7 @@ func TestConnectorAPIv1_ListConnectors(t *testing.T) {
 		Return(map[string]connector.Connector{source.ID(): source, destination.ID(): destination}).
 		Times(1)
 
+	tn := time.Now()
```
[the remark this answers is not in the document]
maha-hajja: done

PR #383, opened 2022-04-26
docs: Pipeline semantics

lovromazgon:
### Description

This PR adds a doc describing pipeline semantics. Note that some sections in the document do not explain the behavior Conduit implements right now but rather how it _should_ work after we fix the problematic behavior explained below.

### Problems in current semantics / implementation

While writing the document I noticed a couple of problems in our current implementation:
- **Stopping a pipeline is controlled by the connector**
  - **Current behavior**: Source connectors are allowed to cache records. When a pipeline gets a signal to stop, Conduit forwards the signal to the source connector which returns a message essentially saying "ok, I will stop when I can". The connector is allowed to drain its cache and produce any number of records after the stop signal, so Conduit relies on the source connector to do the right thing. Since Conduit doesn't know how many more records are coming it lets the connector determine when the stream can be stopped. On the other hand, the stream needs to stay open for acks to flow back from Conduit to the source connector. When something goes wrong and some messages won't get acknowledged we are unable to determine when the stream can be closed - Conduit can't decide to close the stream in case the connector tries to send more messages while the connector can't decide to close the stream in case Conduit will send back acknowledgments. Right now we avoid this by tracking the number of outstanding acks in the connector SDK and waiting for all acks to be received, but again, that will not work correctly when some acks are not sent back.
  - **New behavior**: we will change the response of function `Source.Stop` in the connector protocol and make it return the position of the last record that was produced. The connector is supposed to block the call to `Source.Stop` until the cache is drained (if any) and the record reading routine is stopped completely. This gives Conduit the ability to control the closing of the stream, since it will know when it received the last record from the source and it knows if there are any more open messages in the pipeline that might still produce an acknowledgment. After all messages reached an end state (i.e. are acked or nacked) and Conduit received the message with the position returned from `Source.Stop`, it knows it can safely close the stream and call `Teardown` in the connector.
- **Order of acks is not guaranteed**
  - **Current behavior**: Acks are delivered to the source as they happen and not necessarily in the same order of produced messages. For example a message could take a lot of time to be written to the destination while a message produced after it could be filtered out and acknowledged in the meantime. Some connectors might depend on the correct order as they can't acknowledge arbitrary records (e.g. kafka).
  - **New behavior**: Introduce a node that ensures that acks/nacks are forwarded in the correct order. It should also take care of forwarding nacked messages to a DLQ handler in the correct order (see next problematic behavior).
- **Message end states are needlessly complex**
  - **Current behavior**: A message can be open (i.e. in processing), acked (successfully processed), nacked (processing failed) or dropped (processing failed and node stopped running). The status "dropped" was introduced to be able to control if a message should be sent to a DLQ or not (e.g. in case a fatal error occurred and the pipeline stops asap nodes drop messages instead of nacking them). The distinction between nacked and dropped is subtle and a source of confusion.
  - **New behavior**: Remove the "dropped" status from a message. This can be done by centralizing the decision if a message should be sent to the DLQ or not instead of letting the nodes make the decision. A new ack/nack handler node should take care of keeping track which messages are still in the pipeline, make sure that acks/nacks of those messages are processed in the correct order, forward nacks to a DLQ handler and forward acks to the source connector. This node should be added after each source node.
- **Failed messages stop the pipeline**
  - **Current behavior**: When a message isn't processed successfully it causes the node to stop running and consequently stops the whole pipeline.
  - **New behavior**: Make sure users are able to specify a DLQ handler that decides what happens with failed messages. Three options should be available:
    - Stop the pipeline (default) - this is the current behavior, it treats failed messages as fatal errors that stop the pipeline.
    - Ignore failures - this handler should only log a message and let the pipeline continue running. This would allow a user to change the delivery semantics to "at most once" in case the data flowing through the pipeline is not mission critical and continuous operation is more important than delivering all messages.
    - Reroute to DLQ - this would reroute unsuccessfully processed messages to another destination connector that stores the message in a dead letter queue. If the message fails to be stored in the DLQ the pipeline stops.
  - Note: if we need more advanced behavior regarding DLQs we should add a DLQ handler that reroutes the message to another Conduit pipeline, this would give the user a lot of flexibility. This is not something we want to tackle now though.

### Tasks

Based on the problems identified we need to implement these tasks (issues will be created once the PR is approved and we agree this is what we want):
- Change plugin protocol to return position of last produced position in Source.Stop.
- Change SDK to return last produced position on stop and wait for read stream to be closed.
- Change SDK to not expect all acknowledgments.
- Introduce node to handle acks/nacks
  - Each source node should get a node right after it that registers an ack/nack handler on the message
    that sends acks back to the source connector, also it should reroute nacked messages to the DLQ handler
    and, if they are processed correctly, send acks back to the source connector.
  - Acks/nacks should be processed in the order they were registered (guarantee order of acks sent to
    source connector and order of messages in DLQ).
- Remove status "dropped" from message.
- Introduce DLQ handler that sends messages to a destination.

Here is a list of created issues:
- #389
  - #390
  - #391
  - #392
  - #393
  - #394


docs/pipeline_semantics.md
```diff
@@ -0,0 +1,170 @@
...
+
+A pipeline node can register state change handlers on a message that will be called when the message state changes. This
+is used for example to register handlers that reroute nacked messages to a dead letter queue or to update metrics when a
+message reaches the end of the pipeline.
+
+## Semantics
+
+### Messages are delivered in order
+
+Since messages are passed between nodes in channels and a node only processes one message at a time it is guaranteed
```
maha-hajja, line 92:
```suggestion
Since messages are passed between nodes in channels and a node only processes one message at a time, it is guaranteed
```

docs/pipeline_semantics.md
```diff
@@ -0,0 +1,170 @@
...
+that messages will flow through the Conduit pipeline in the same order as they are received by the source.
```
maha-hajja, line 93:
```suggestion
that messages will flow through the Conduit pipeline in the same order that they were received at the source.
```

docs/pipeline_semantics.md
```diff
@@ -0,0 +1,170 @@
...
+
+There are two caveats:
+
+- If a pipeline contains multiple source connectors then the order of two messages coming from different connectors is
+  nondeterministic. Messages coming from the same source connector are still guaranteed to retain their order.
+- If a dead letter queue is configured then negatively acknowledged messages will be removed from the stream while the
```
maha-hajja, line 99:
```suggestion
- If a dead letter queue is configured then negatively acknowledged, messages will be removed from the stream while the
```

docs/pipeline_semantics.md
```diff
@@ -0,0 +1,170 @@
...
+  pipeline will be kept running, thus impacting the order of messages.
```
maha-hajja, line 100:
```suggestion
  pipeline is still running, thus impacting the order of messages.
```

docs/pipeline_semantics.md
```diff
@@ -0,0 +1,170 @@
...
+
+The order guarantee only holds inside of Conduit. Once a message reaches a destination connector it is allowed to buffer
```
maha-hajja, line 102:
```suggestion
The order guarantee only holds inside of Conduit. Once a message reaches a destination connector, it is allowed to buffer
```

docs/pipeline_semantics.md
```diff
@@ -0,0 +1,170 @@
...
+messages and batch write them to 3rd party systems. Normally the connector would retain the order, although we can't
+vouch for badly written connectors that don't follow this behavior.
+
+### Messages are delivered at least once
+
+Between pipeline restarts, it is guaranteed that any message that is processed successfully by all nodes and not
+filtered out will be delivered to a destination connector at least once. Multiple deliveries can occur in pipelines with
+multiple destinations that stopped because of a negatively acknowledged record, or pipelines, where a destination
```
maha-hajja, line 110: this time it's an extra comma :D 
```suggestion
multiple destinations that stopped because of a negatively acknowledged record, or pipelines where a destination
```

docs/pipeline_semantics.md
```diff
@@ -0,0 +1,170 @@
...
+implementing the write operation of a destination connector in an idempotent way (if possible).
+
+The delivery guarantee can be changed to "at most once" by adding a [dead letter queue](#Dead-letter-queue) handler that
+drops unsuccessfully processed messages.
+
+### Acks are delivered in order
+
+Conduit ensures that acknowledgments are sent to the source connector in the exact same order as records produced by the
+connector. This guarantee still holds, even if a destination connector acknowledges records in a different order or if a
+later record gets filtered out faster than an earlier record managed to get written to all destinations.
```
maha-hajja, line 121: I don't get this sentence.. could you elaborate more, please?

docs/pipeline_semantics.md
```diff
@@ -0,0 +1,170 @@
...
+that simply logs a warning and drops messages to achieve "at most once" delivery guarantees.
+
+### Pipeline stop
+
+A pipeline can be stopped in two ways - either it's stopped gracefully or forcefully.
+
+- A graceful stop is initiated either by Conduit shutting down or by the user requesting the pipeline to stop. Only the
+  source connector nodes will receive the signal to stop running. The source nodes will stop running and close their
+  outgoing channels, notifying the downstream nodes that there will be no more messages. This behavior propagates down
+  the pipeline until the last node stops running, draining any open messages in the process.
```
maha-hajja, line 165: draining here means writing them to all destinations?

docs/pipeline_semantics.md
```diff
@@ -0,0 +1,170 @@
...
+- A forceful stop is initiated when a node stops running because it experienced an unrecoverable error. In that case,
+  the context that is shared by all nodes will get canceled, signaling to all nodes simultaneously that they should stop
+  running as soon as possible. Messages that are in the pipeline won't be drained, instead, they are dropped and will be
+  requested from the source again once the pipeline is restarted. The error returned from the first node that failed
+  will be stored in the pipeline and can be retrieved through the API.
```
maha-hajja, line 170: should we specify which API?

maha-hajja (commented):
Awesome work! this is a really nicely written and informative doc

docs/pipeline_semantics.md
lovromazgon: In _Message states_ we explain the "Acked" status:

> Acked - a message was successfully processed and acknowledged. This can be done either by a processor (e.g. **it filtered the message out**) or by a destination.
> /.../
> Pipeline nodes will either leave the message open and send it to the next node for processing or **ack/nack it and not send it further down the pipeline**.

To sum up, "filtering a message" means acking it and not sending it to the next node.

Imagine a source connector that first produced message A and then message B. Message A would flow all the way to the destination, while message B would be filtered out by a processor. Since nodes are processing messages in parallel, message B could get filtered out (i.e. acked) faster than message A would reach the destination. Conduit still makes sure that acks will reach the source in the correct order - it waits for message A to be acked before sending the ack for message B back to the source, even though internally the acks might have happened in the opposite order.

The other scenario is that of a badly written destination connector that writes records asynchronously and messes up the order of acks sent back to Conduit - in that case Conduit would again make sure to correctly order the acks sent back to the source connector.

---

I'll see if I can reword it to make it clearer.

docs/pipeline_semantics.md
lovromazgon: Essentially yes. I'll reword it to make this clear.

docs/pipeline_semantics.md
lovromazgon: I didn't go into detail here on purpose, it would be out of scope for this document. To be honest, I was reluctant even mentioning the API here, but didn't find a better way of saying that the error is stored and retrievable.

docs/pipeline_semantics.md
```diff
@@ -0,0 +1,170 @@
...
+  destination. If a pipeline contains multiple destinations the message needs to be negatively acknowledged by at least
+  one destination before it is marked as nacked. When a message is nacked the message is passed to
+  the [DLQ](#Dead-letter-queue).
+
+**Important**: if a message gets nacked and the DLQ handler successfully processes the nack (e.g. stores the message in
+the dead letter queue), the source connector will receive an ack as if the message was successfully processed, even
+though Conduit marks it internally as nacked. In other words, the source connector will receive an ack every time
+Conduit handled a message end-to-end and it can be safely discarded from the source.
+
+Pipeline nodes will either leave the message open and send it to the next node for processing or ack/nack it and not
```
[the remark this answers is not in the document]
lovromazgon: Nothing, that message is "done", the processor will just drop it and fetch the next message. The destination won't see a message that was acked by a processor since the processor node won't send it to the next node.

docs/pipeline_semantics.md
lovromazgon: I went with "...same order that was produced by the source."

docs/pipeline_semantics.md
lovromazgon: That comma would change the meaning of the sentence, I reworded it to make it clearer. I changed "then" into a comma instead.

docs/pipeline_semantics.md
lovromazgon: I went with "pipeline will keep running"

docs/pipeline_semantics.md
lovromazgon: @maha-hajja let me know if the rewording makes it clearer.

docs/pipeline_semantics.md
lovromazgon: @maha-hajja let me know if the rewording makes it clearer.

docs/pipeline_semantics.md
maha-hajja: the change made it much clearer

docs/pipeline_semantics.md
maha-hajja: agree

docs/pipeline_semantics.md
maha-hajja: Ah I had the whole meaning of "filtered out" misunderstood.. I thought filtered meant filtering some fields from the records and was wondering why would we ack it after that.
But now I get it! thank you

maha-hajja: @lovromazgon yep rewording is perfect

docs/pipeline_semantics.md
```diff
@@ -0,0 +1,172 @@
...
+## Message
+
+A message is a wrapper around a record that manages the record's lifecycle as it flows through the pipeline. Messages
+are created in source nodes when they receive records from the source connector, and they are passed down the pipeline
+between nodes until they are acked or nacked. Nodes are only allowed to hold a reference to a single message at a time,
+meaning that they need to pass the message to the next node before taking another message¹. This also means there is no
+explicit buffer in Conduit, a pipeline can only hold only as many messages as there are nodes in the pipeline (see
+[backpressure](#Backpressure) for more information).
+
+¹This might change in the future if we decide to add support for multi-message transforms.
```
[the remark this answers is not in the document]
lovromazgon: It's all speculation at this point and it depends on if/how we want to implement this. But yes, essentially this would mean a node could hold multiple references to messages and do with them more powerful transformations (e.g. combine into a single record, batch). Calling this an "explicit buffer" is a bit of a stretch IMO, since the buffering would be more of a side-effect than the purpose of such functionality. Technicalities aside, it would definitely create a buffer, that much is certain.

docs/pipeline_semantics.md
```diff
@@ -0,0 +1,172 @@
...
+### Pipeline stop
+
+A pipeline can be stopped in two ways - either it's stopped gracefully or forcefully.
+
+- A graceful stop is initiated either by Conduit shutting down or by the user requesting the pipeline to stop. Only the
+  source connector nodes will receive the signal to stop running. The source nodes will stop running and close their
+  outgoing channels, notifying the downstream nodes that there will be no more messages. This behavior propagates down
+  the pipeline until the last node stops running. Any messages that were being processed while the pipeline received a
+  stop signal will be processed normally and written to all destinations.
+- A forceful stop is initiated when a node stops running because it experienced an unrecoverable error. In that case,
```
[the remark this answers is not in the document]
lovromazgon: I'll add an example. Two scenarios come to mind when I hear unrecoverable error - either there is no DLQ set up and a nacked message causes the node to stop running, or we experience an issue with a connector (e.g. it has a bug causing it to panic or produce an error, or it can't communicate with the 3rd party system for some reason).

docs/pipeline_semantics.md
```diff
@@ -0,0 +1,172 @@
...
+
+### Messages are delivered at least once
+
+Between pipeline restarts, it is guaranteed that any message that is processed successfully by all nodes and not
+filtered out will be delivered to a destination connector at least once. Multiple deliveries can occur in pipelines with
+multiple destinations that stopped because of a negatively acknowledged record, or pipelines where a destination
+negatively acknowledged a record and processed more messages after that. For this reason, we strongly recommend
+implementing the write operation of a destination connector in an idempotent way (if possible).
+
+The delivery guarantee can be changed to "at most once" by adding a [dead letter queue](#Dead-letter-queue) handler that
```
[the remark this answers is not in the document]
lovromazgon: By default, the DLQ handler would just return an error, causing the node that nacked the message to stop running.

On the other hand, you could configure a handler that sends the message to another destination (that's the DLQ), or a handler that just ignores nacked messages and does not cause the node to stop.

For more info about this see the PR description, bullet point "Failed messages stop the pipeline". Note that we still need to implement this behavior, right now we only have the default.

docs/pipeline_semantics.md
```diff
@@ -0,0 +1,172 @@
...
+source connector, otherwise the order delivery guarantee would be violated. The absence of an acknowledgment after the
+source connector teardown is initiated can be interpreted as a negative acknowledgment.
+
+### Backpressure
+
+The usage of unbuffered channels between nodes and the absence of explicit buffers results in backpressure. This means
+that the speed of the destination connector dictates the speed of the whole pipeline.
+
+There is an implicit buffer that needs to be filled up before backpressure takes effect. The buffer is equal to the
+number of nodes in a pipeline - similar to how a longer garden hose holds more water, a longer Conduit pipeline can hold
```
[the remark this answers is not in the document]
lovromazgon: I almost didn't include this, thinking it felt a bit out of place 😅  Good to know it makes sense!

docs/pipeline_semantics.md
```diff
@@ -0,0 +1,172 @@
...
+
+- **Open** - all messages start in the open state. This means the message is currently in processing, either by a node
+  or a destination connector. A pipeline won't stop until all messages transition from the open state into one of the
+  other two states.
+- **Acked** - a message was successfully processed and acknowledged. This can be done either by a processor (e.g. it
+  filtered the message out) or by a destination. If a pipeline contains multiple destinations, the message needs to be
+  acknowledged by all destinations before it is marked as acked. Acks are propagated back to the source connector and
+  can be used to advance the position in the source system if applicable.
+- **Nacked** - a message was negatively acknowledged either by a processor (e.g. a transform failed) or by a
+  destination. If a pipeline contains multiple destinations, the message needs to be negatively acknowledged by at least
```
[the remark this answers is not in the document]
lovromazgon: > NACK (received with errors or was otherwise unreadable, etc)

I assume you're talking about the NACK signal in the context of network protocols? While the term is borrowed from that space, it does not necessarily mean the same thing in Conduit. Here a nack means that the processing of a message failed and resulted in an error.

I'll try to clarify this part!

hariso (approved):
Great stuff!

It might be good clarifying what happens when a message state handler fails and also adding notes about parts which are not implemented yet (e.g. the DLQ).

docs/pipeline_semantics.md
```diff
@@ -0,0 +1,173 @@
...
+A pipeline node can register state change handlers on a message that will be called when the message state changes. This
+is used for example to register handlers that reroute nacked messages to a dead letter queue or to update metrics when a
+message reaches the end of the pipeline.
+
+## Semantics
+
+### Messages are delivered in order
+
+Since messages are passed between nodes in channels and a node only processes one message at a time, it is guaranteed
+that messages will flow through the Conduit pipeline in the same order that was produced by the source.
```
hariso, line 93: >  it is guaranteed that messages will flow 
Just so nobody gets the impression it's messages generally that will flow through a pipeline in the same order, it might be good to say "that messages from a single source connector will flow...". I do see there's a related caveat mention, but I think it's good to be clear here too.

lovromazgon:
@ConduitIO/conduit-core, @uchennakevinm1 Thanks for the reviews, I think I addressed all the feedback so far! I didn't see any objections to the tasks identified in the PR description, so I will create them. I'll keep this PR open until tomorrow and merge it if there are no more comments.

docs/pipeline_semantics.md
```diff
@@ -0,0 +1,182 @@
+# Pipeline semantics
+
+This document describes the inner workings of a Conduit pipeline, its structure, and behavior. It also describes a
+Conduit message and its lifecycle as it flows through the pipeline.
+
+**NOTE**: Some parts of this document describe behavior that is not yet fully implemented (e.g. DLQs, order of acks).
+For more information see [#383](https://github.com/ConduitIO/conduit/pull/383#pullrequestreview-963196784). This note
```
hariso, line 7: Nitpick: I believe that it was intended to include the link to the PR only, not the PR review.

lovromazgon: Oko sokolovo 🦅

hariso: loooool :D

uchennakevinm1:
LGTM! I learned quite a bit from this document.

Issue #33, opened 2021-06-21
Load pipeline config files on startup


lovromazgon:
Conduit should by default load all YAML files in folder `pipelines` (relative to the `conduit` binary) and provision pipelines when it starts. The folder location should be configurable through a CLI flag.

Depends on #493.

Issue #31, opened 2021-06-21
CLI configuration


lovromazgon:
The goal is to make the Conduit CLI configurable. We should support the combination of parsing of flags, environment variables and/or a config file. We need to decide what things need to be configurable (e.g. log level, API ports, path to config file, path to pipeline configs) and the corresponding flag/env var/config field names. We already have some simple CLI flags, the goal of this issue is to implement it in a way that makes it easy to add new options in the future.

lovromazgon:
Relevant for this issue - https://github.com/peterbourgon/ff

PR #384, opened 2022-04-26
docs: Loading Conduit pipelines from config files

dylanlott:
### Description

This PR includes an initial design for loading Conduit pipelines from configuration files. It proposes a solution of an in-memory database populated with a user-provided pipeline configuration file and environment information that allows Conduit immutability at runtime without sacrificing connector durability between restarts or failures.

Related to #33


docs/design-documents/20220408-loading-pipeline-configuration-files.md
```diff
@@ -0,0 +1,241 @@
...
+
+- We don't care about the structure or schema of configuration files.
+- We shouldn't implement key management or encryption for secrets.
+
+## Scope
+
+We are strictly concerned with the loading of pipeline files at start time and 
+how those files are handled through Conduit's full operation cycle.
+
+Configuration file schema and other design design choices are out of scope.
```
lyuboxa, line 31: Is the design stutter intentional?

docs/design-documents/20220408-loading-pipeline-configuration-files.md
```diff
@@ -0,0 +1,241 @@
...
+separately from Conduit's state.
+
+### Immutability
+
+Conduit should be immutable at runtime. We achieve this by stubbing out all
+of the mutable methods as discussed in previous sections.
+
+## Open questions
+
+- Do we ever want to support generating config files from Conduit's other data
```
lyuboxa, line 204: I think this will be cool, taking snapshot of the runtime config that is.

docs/design-documents/20220408-loading-pipeline-configuration-files.md
```diff
@@ -0,0 +1,240 @@
...
+limited version of the GUI instead.
+
+Conduit Issues #32 and #33 describe two different features to address some of 
+these concerns, but there are additional design decisions to make.
+
+## Design 
+
+We propose the creation of a single configuration file that contains all of
+Conduit's configuration values as well as those for `n` number of Pipelines.
+We consider a pipeline as `m` number of Connectors and `o` number of Processors.
```
hariso, line 66: > We propose the creation of a single configuration file that contains all of
Conduit's configuration values

Just confirming this is intended, as the PR title and the original issue appear to be about pipeline configs only.

docs/design-documents/20220408-loading-pipeline-configuration-files.md
```diff
@@ -0,0 +1,240 @@
...
+serve a read-only optimized view of Conduit.
+
+Configuration must securely accept secret information. We don't want to solve
+the problem of secrets management, but we must be in the business of secret
+concealment. Additionally, we already allow environment variables so we must 
+continue handling those correctly. 
+
+We propose merging environment variables with the user's provided configuration 
+file at runtime and replacing templated variables in the configuration file with 
+their secret values at runtime.
```
hariso, line 83: How about CLI args? For example, what if we have the HTTP port specified in the config file and through the CLI arg too? 

And a nitpick: it might be good to be explicit about the config priorities (config file > environment variable).

docs/design-documents/20220408-loading-pipeline-configuration-files.md
```diff
@@ -0,0 +1,240 @@
...
+file into a full pipeline by hydrating it with other necessary information 
+including environment information, connector configurations, and processors.
+
+This step will be responsible for populating the service's individual DB
+references with the resources necessary before `Init` is called after `Run`.
+
+### Immutability
+
+Conduit should be immutable at runtime. We could achieve this by stubbing 
+out all mutable methods as discussed in previous sections.
```
hariso, line 190: How about pipeline status and the error (if any)? When Conduit starts, a pipeline's status is set to running (if it was stopped).

docs/design-documents/20220408-loading-pipeline-configuration-files.md
dylanlott: Good point! I updated the document to reflect this. I think we should merge environment variables and CLI flags down into a cohesive `Environment` object that can contain sensitive information.

docs/design-documents/20220408-loading-pipeline-configuration-files.md
dylanlott: Initially, I was approaching it as a configuration file for the entire Conduit server, but I have since tightened the scope of these files to be about pipeline configuration only and updated the doc to reflect that.

docs/design-documents/20220408-loading-pipeline-configuration-files.md
```diff
@@ -0,0 +1,270 @@
...
+
+## Key Terms 
+
+**pipeline** - a Conduit pipeline with `n` number of Source connectors, 
+`m` number of Processors, and `o` number of Destination connectors.
+
+**production mode** - a mode of operation where Conduit pipelines must be
+statically defined and cannot change after Conduit starts. No UI is served in
+production mode, and only pipeline lifecycle methods can be called over the
+API.
```
lovromazgon, line 50: I feel like such a mode would impose a tough limitation. I would rather we limit the immutability to the level of a pipeline, essentially saying that if a certain pipeline is provisioned using a config file it can't be mutated through the API.

docs/design-documents/20220408-loading-pipeline-configuration-files.md
```diff
@@ -0,0 +1,270 @@
...
+Conduit that either doesn't serve the application GUI or serves a functionally 
+limited version of the GUI instead.
+
+Conduit Issues #32 and #33 describe two different features to address some of 
+these concerns, but there are additional design decisions to make.
+
+## Design 
+
+We propose the creation of a pipeline configuration file that contains all of
+the configuration values needed for `n` number of full pipelines.
```
lovromazgon, line 70: What if we instead have a folder that is scanned on startup and all `.yml`/`.yaml` files in that folder are loaded? Users can still choose to create a single file with [multiple documents](https://yaml.org/spec/1.2.2/#22-structures) by separating them with `---` or split them into separate files.

docs/design-documents/20220408-loading-pipeline-configuration-files.md
```diff
@@ -0,0 +1,270 @@
...
+config file into a full pipeline by hydrating it with other necessary 
+information including environment information, connector configurations, 
+and processors.
+
+This step will be responsible for populating the service's individual `Stores` 
+with the necessary resources before `Init` is called during start.
+
+This process can fail, in which a pipeline reports `StatusDegraded`.
+Because we aren't dictating restart behavior in configuration files, this
+should cause Conduit to fail and report the pipeline errors it received.
```
lovromazgon, line 195: Are we saying that Conduit should stop running if a pipeline defined in a config file encounters an error? If so, that would mean any other pipeline running on that instance would be stopped as well 😕  If a user wanted resilience, they would have to run one Conduit instance per pipeline, I'm not convinced it's the right thing to do.

docs/design-documents/20220408-loading-pipeline-configuration-files.md
lovromazgon: IMO the priority should be CLI flags (highest priority) > environment variables (lower priority) > configuration file (lowest priority).

That said, I feel like we are mixing apples and oranges. CLI flags would only apply to the main Conduit configuration file (which we don't yet support, relevant issue: https://github.com/ConduitIO/conduit/issues/31), while this document is describing the pipeline configuration file. I don't think CLI flags would apply to pipeline configuration files, they should be able to reference environment variables though.

docs/design-documents/20220408-loading-pipeline-configuration-files.md
dylanlott: I should clear this up, but no, that is not what was intended. Conduit's current behavior of failing the pipeline is what I was meaning to describe here, no changes to that are necessary in my opinion.

docs/design-documents/20220408-loading-pipeline-configuration-files.md
dylanlott: I considered this but after initial feedback on the document I thought that doing less and making it simpler seemed appropriate so I went with just the single file design. I would be open to making this as you described, though.

docs/design-documents/20220408-loading-pipeline-configuration-files.md
dylanlott: I built a good chunk of this design with conversations in mind where we had discussed a stricter production-mode standard for Conduit, so this limitation is admittedly part of my design. I would be open to considering this change, though. I think it achieves nearly similar results, with perhaps broader use cases. Let me think more about it! 🤔

docs/design-documents/20220408-loading-pipeline-configuration-files.md
dylanlott: Status would be considered part of the connector lifecycle and therefore mutable by necessity.

docs/design-documents/20220408-loading-pipeline-configuration-files.md
dylanlott: I agree about both priority and that the CLI flags is an overlapping but unrelated item; I only want this to concern itself with loading environment variables necessary for pipeline files, not how they're managed or set. I can update this to be more specific about scope around environment variable handling and set a non-goal for CLI flag settings or anything like that.

docs/design-documents/20220408-loading-pipeline-configuration-files.md
dylanlott: I've updated the design to include language for supporting multiple pipeline configuration files. 👍

dylanlott: I've updated the design to include language for supporting multiple pipeline configuration files. 👍

dylanlott:
Updated design document to reflect multiple configuration file support.

docs/design-documents/20220408-loading-pipeline-configuration-files.md
```diff
@@ -0,0 +1,282 @@
...
+
+This document describes the design and implementation for loading pipeline 
+configuration files. It details how Conduit loads configuration files at start.
+
+In order to provide a detailed history of configuration changes, better secret
+handling, and increased confidence in a given configuration, we propose
+adding support for loading and subsequently running pipelines entirely from 
+configuration files loaded into Conduit from a configured directory.
+
+Our solution proposes an immutable Conduit runtime backed by an in-memory DB
```
lovromazgon, line 13: I'm not sure I follow the reasoning behind the in-memory DB. IIUC we would like to build the pipeline from the config file and keep it only in memory without persisting it anywhere, right? At the same time the document mentions that the `ConnectorPersister` would still be able to persist the connector state (position) between restarts. If that's what we are trying to do, it's not going to work that easily, because the connector persister is persisting the whole connector and not only its state. We would need to make sure that we can (a) persist the connector state separately and (b) that we can associate the state with a specific connector.

To solve problem (a) we could persist the pipeline, connectors and processors same as if they were created through the API and update them when the pipeline config is loaded. Then the config file only plays a role in how the pipeline is created, everything else functions the same.

To solve problem (b) we need a way to associate the persisted pipeline with a certain pipeline config file. If we persist pipelines in the store (as proposed in the previous paragraph) then the simple solution is to add a field to the pipeline config file that uniquely identifies a single pipeline (i.e. user explicitly sets the pipeline id). Otherwise we would need to allow the user to uniquely identify every single connector so that we can associate it to the stored state.

docs/design-documents/20220408-loading-pipeline-configuration-files.md
lovromazgon: The document still largely talks about a single pipeline configuration file (chapter Design, Implementation), if we decided to support multiple configuration files loaded from a single folder then I think we should change the wording.

docs/design-documents/20220408-loading-pipeline-configuration-files.md
```diff
@@ -0,0 +1,282 @@
...
+information between restarts.
+
+Since we must persist connector information, but we want Conduit services 
+to be immutable, we should pass the `ConnectorPersister` its own database
+reference at start that it can mutate without changing Conduit's state. 
+
+## Lifecycle methods 
+
+Configuration files should specify whether or not a service should run at
+startup, but should not dictate restart behavior.
```
lovromazgon, line 218: Did you mean "pipeline" instead of "service"?

docs/design-documents/20220408-loading-pipeline-configuration-files.md
```diff
@@ -0,0 +1,282 @@
...
+
+We propose merging CLI flags, environment variables, and the user's provided 
+configuration file, in that order, at runtime into template variables in the 
+configuration file with their secret values at runtime.
+
+We propose merging environment variables and CLI flags into a set of values 
+that is then injected into the provided set of pipeline files.
+
+We propose supporting Go templating in all pipeline files to allow for dynamic
+injection of those environment variables at runtime.
```
lovromazgon, line 94: Can you explain what you had in mind with Go templating?

docs/design-documents/20220408-loading-pipeline-configuration-files.md
lovromazgon: I feel like this was not addressed.

docs/design-documents/20220408-loading-pipeline-configuration-files.md
```diff
@@ -0,0 +1,282 @@
...
+	// store. After Close is called other methods should return an error.
+	Close() error
+
+	Set(ctx context.Context, key string, value []byte) error
+	Get(ctx context.Context, key string) ([]byte, error)
+	GetKeys(ctx context.Context, prefix string) ([]string, error)
+}
+```
+
+### Configuration driver
```
lovromazgon, line 156: The configuration driver would implement `database.DB` and only allow "read-only" operations, right? I feel like this was written with "production mode" in mind, so this would need to be rethought if we decide we only want to make single pipelines immutable instead of the whole Conduit instance. Even in case of "production mode" the problem from my first comment still persists.

docs/design-documents/20220408-loading-pipeline-configuration-files.md
dylanlott: I imagine something like [benjohnson/tmpl](https://github.com/benbjohnson/tmpl) but custom. Go style template variables would be used to declare where a value should be used, and then at start-time the corresponding variable would be injected.

Issue #244, opened 2022-02-18
Make it possible to create a Kafka connector plugin from the UI


hariso:
### Feature description

It should be possible to create a Kafka connector plugin from the UI. The path to the plugin will be `pkg/plugins/kafka-connector-plugin`.

lovromazgon:
@hariso is this feature talking about adding https://github.com/conduitio/conduit-kafka-connect-wrapper? Because that's going to be a problem - we don't ship Conduit with the plugin so we can't hardcode it in the UI, and the plugin API won't return the specs for the plugin because it only returns them for builtin plugins right now.

hariso:
@lovromazgon Yes, this is about the wrapper. Looks like the best option we have now is to simply add the wrapper connector via the HTTP API. In that case we can probably add something to the docs, explaining how to add a standalone plugin (for now).

PR #399, opened 2022-05-10
Describe Plugin API UI support

jmar910:
### Description
+ Removes front-end hardcoded connector plugin config "blueprints" in favor of blueprints returned from the Conduit API
+ Update the front-end logic/spec for "blueprints" to conform to what the API returns (Fields as keys, type constants)
+ Sets up a proper ember-data relationship between connectors & plugins
+ Updates our hardcoded transform config "blueprints" to use the same spec as the API. This future proofs us for when we decide to return transform blueprints from the API.

Fixes https://github.com/ConduitIO/conduit/issues/244

### Quick checks:


ui/app/components/pipeline-editor/config-field.js
```diff
@@ -30,23 +30,20 @@ export default class PipelineEditorConfigFieldComponent extends Component {
...
-    const isNumber = field.type === 'int' || field.type === 'long';
-
-    const isPassword = field.type === 'password';
+    const isNumber = field.type === 'TYPE_NUMBER';
 
     // TODO clean this up, and also check for special case of number, but using inclusion validation
     // for numbers in a list?
     const isTextInput =
-      isString && !field.rawValidations.findBy('type', 'inclusion');
+      isString && !field.rawValidations.findBy('type', 'TYPE_INCLUSION');
```
jayjayjpg, line 40: This isn't a blocker and just a thought: would it make sense to create a `constants.js` file or something similar from which these values can be imported? This way, if the value should change in the future, the update would only require a one line change

What are thoughts on this?

ui/app/models/connector.js
```diff
@@ -10,25 +10,14 @@ export default class ConnectorModel extends Model {
...
-    if (this.plugin && this.type) {
-      return this.store
-        .peekAll('connector-plugin')
-        .filterBy('pluginPath', this.plugin)
-        .findBy('connectorType', this.type);
-    } else {
-      return null;
-    }
-  }
+  @belongsTo('plugin')
```
jayjayjpg, line 19: Neat!

ui/app/serializers/connector.js
```diff
@@ -20,11 +20,22 @@ export default class ConnectorSerializer extends ApplicationSerializer {
...
 
+  extractRelationship(modelName, value) {
+    if (modelName === 'plugin' && value) {
+      return {
+        id: value,
+        type: modelName,
+      };
+    }
+
+    super.extractRelationship(modelName, value);
```
jayjayjpg, line 36: Would this also require a `return` statement or is this irrelevant in this case?

```js
return super.extractRelationship(modelName, value);
```

ui/tests/acceptance/pipeline/index/connector-modal-test.js
```diff
@@ -62,29 +77,39 @@ module('Acceptance | pipeline/index/connectors/kafka', function (hooks) {
...
               type: 'TYPE_SOURCE',
             },
             'it calls the API with the correct protocol'
           );
-          return connectors.create(attrs);
+
+          const plugin = plugins.find(attrs.plugin);
+          const pipeline = pipelines.find(attrs.pipeline_id);
+
+          return connectors.create({
```
jayjayjpg, line 104: This is neat!

jayjayjpg (approved):
This reads great! I left a few more suggestions and questions to learn more, but this is already good to go! :sparkles:

PR #418, opened 2022-05-25
HTTP Request processor

lovromazgon:
### Description

This adds a processor with the name `httprequest` that sends an HTTP request to the specified URL with the specified HTTP method (default is POST). The record payload is used as the request body and the raw response body is put into the record payload.

### Quick checks:


pkg/processor/transform/txfbuiltin/httprequest.go
```diff
@@ -0,0 +1,106 @@
...
+)
+
+func init() {
+	processor.GlobalBuilderRegistry.MustRegister(httpRequestName, transform.NewBuilder(HTTPRequest))
+}
+
+// HTTPRequest builds a transform that sends an HTTP request to the specified
+// URL with the specified HTTP method (default is POST). The record payload is
+// used as the request body and the raw response body is put into the record
+// payload.
```
hariso, line 43: Since headers are left out (maybe out of scope for now?), we will likely not be able to authenticate. Also, some endpoints will require certain headers to be present (e.g. Content-Type), or the request cannot be processed.

maha-hajja (approved):
LGTM

pkg/processor/transform/txfbuiltin/httprequest.go
lyuboxa: Minimal is ok, though, it will be nice to have some rudimentary retry policy that could handle blips. Timeouts as well.

lovromazgon: I implemented a configurable backoff retry that is disabled by default, but can be enabled by setting `backoffRetry.count` to a number greater than 0. You can also fine-tune the backoff retry with `backoffRetry.min`, `backoffRetry.max` and `backoffRetry.factor` (the delay between retries starts with min and is multiplied by factor until it reaches max). Any request that returns a status code greater than 299 will be retried.

@lyuboxa let me know if that's what you had in mind.

pkg/processor/transform/txfbuiltin/httprequest.go
```diff
@@ -0,0 +1,181 @@
...
+			return record.Record{}, cerrors.Errorf("%s: error trying to execute HTTP request: %w", transformName, err)
+		}
+		defer resp.Body.Close()
+
+		body, err := ioutil.ReadAll(resp.Body)
+		if err != nil {
+			return record.Record{}, cerrors.Errorf("%s: error trying to read response body: %w", transformName, err)
+		}
+
+		if resp.StatusCode > 299 {
```
hariso, line 109: We can maybe reduce this to just 429 + some 5xx codes. 3xx will all be redirects, i.e. not errors, so there's probably no need to retry those. 4xx would be client errors, and they will mostly fail again (except for maybe 408 and 429). Similar for  the 5xx ones.

hariso (approved):
LGTM! I had a suggestion around retries for different HTTP codes, but not a blocker.

pkg/processor/transform/txfbuiltin/httprequest.go
lovromazgon: I agree with you, those codes you mention probably don't make sense to be retried, although they also don't make sense to succeed IMO, at least it would be unexpected to me if they would mean the transform was successful. Given the current approach it wouldn't be so easy to skip those retries (the retry logic has no idea about the status code). Then again, even if we add this logic, there will inevitably be users that need more control, ideally I think we could have a list of successful and failed status codes that the user can configure (and sane defaults).

That said, I hope this solution is good enough for now, given that this processor was meant to address an immediate problem. So unless @lyuboxa says this doesn't cover their use case, I'd say we can expand this functionality in the future.

pkg/processor/transform/txfbuiltin/httprequest.go
lyuboxa: looks good

pkg/processor/transform/txfbuiltin/httprequest.go
lyuboxa: I think we are good for now, we will iterate as the need arises.

lyuboxa (approved):
This is great thanks!

pkg/processor/transform/txfbuiltin/httprequest.go
lyuboxa: Yes, we can do for now ;-)

Issue #393, opened 2022-05-05
Feature: Ack/nack handler node


lovromazgon:
Introduce a separate node to handle acks/nacks.
- Each source node should get a node right after it that registers an ack/nack handler on the message that sends acks back to the source connector, also it should reroute nacked messages to the DLQ handler and, if they are processed correctly, send acks back to the source connector.
- Acks/nacks should be processed in the order they were registered (guarantee order of acks sent to source connector and order of messages in DLQ).

This shouldn't change the current behavior for nacked messages (i.e. if a message gets nacked it should still stop the pipeline), although it can change where the decision happens (e.g. not in the message handler but rather in a DLQ handler).

PR #451, opened 2022-06-13
Semaphore

lovromazgon:
### Description

This adds an implementation of a semaphore that works based on tickets. It's based on [this semaphore](https://pkg.go.dev/golang.org/x/sync/semaphore), but is highly customized for our needs. Essentially the `Acquire` function is split into `Enqueue` and `Acquire`, where `Enqueue` returns a `Ticket` that needs to be supplied to `Acquire`. This way the caller can enqueue first, do some work, and then come back to acquire the lock, while still ensuring that it didn't lose its place in the queue.

This will be needed in https://github.com/ConduitIO/conduit/issues/393 to control the order of acks going back to the source. The message will take a ticket when it enters the pipeline and will be able to "redeem" it in the ack/nack handler node when it will try to be acked/nacked. The ticket queue will ensure tickets are redeemed in the correct order, ensuring that acks will be processed in the same order as the order of messages entering the pipeline.

### Quick checks:


maha-hajja (approved):
lgtm

hariso (commented):
You said it's highly customized for our needs. Do you think other Go projects might have similar needs, so we can maybe have this published as a library?

lovromazgon:
We could publish it as a library, it's self-contained so it wouldn't be too hard. In fact, the same could be done for essentially any package inside of `foundation`. I don't think it's a priority right now though, I don't think this use case is very common, otherwise I'd expect to see a library already for such a data structure.

pkg/foundation/semaphore/semaphore.go
```diff
@@ -0,0 +1,139 @@
...
+	acquired bool
+}
+
+// Ticket reserves a place in the queue and can be used to acquire access to a
+// resource.
+type Ticket struct {
+	index int
+	batch int64
+}
+
```
hariso, line 48: A thought around the waiters and tickets: how about we remove one of them, and only have a queue (slice) of waiters (or tickets)? Whenever we `Enqueue()` a caller, we put a ticket at the end of the underlying slice. Each ticket would have this `ready` channel. `Acquire()` would wait for the `ready` channel to be closed. `Release` would remove a ticket from the slice, and close the `ready` channel of the next ticket.

hariso (commented):
Great one! :muscle: I had a thought about maybe simplifying the semaphore. Let me know what you think about it.: )

pkg/foundation/semaphore/semaphore.go
lovromazgon: I didn't do it that way because I assumed it would make it slower since we would be allocating and passing a channel around in the `Ticket`. I didn't want to rely on my assumption though, so I tried it out (see [what I did here](https://github.com/ConduitIO/conduit/blob/lovro/ticketqueue-chan-in-ticket/pkg/foundation/semaphore/semaphore.go)). The code is much simpler indeed, the problem is that the benchmarks show the code is noticeably slower. These are benchmarks for the simpler implementation (only `Ticket`):
```
goos: darwin
goarch: amd64
pkg: github.com/conduitio/conduit/pkg/foundation/semaphore
cpu: Intel(R) Core(TM) i9-9880H CPU @ 2.30GHz
BenchmarkNewSem-16             	1000000000	         0.2400 ns/op	       0 B/op	       0 allocs/op
BenchmarkEnqueueOneByOne/acquire-1-16         	11396336	       102.2 ns/op	      96 B/op	       1 allocs/op
BenchmarkEnqueueOneByOne/acquire-2-16         	 5948336	       202.3 ns/op	     192 B/op	       2 allocs/op
BenchmarkEnqueueOneByOne/acquire-8-16         	 1484432	       809.7 ns/op	     768 B/op	       8 allocs/op
BenchmarkEnqueueOneByOne/acquire-64-16        	  177836	      6420 ns/op	    6144 B/op	      64 allocs/op
BenchmarkEnqueueOneByOne/acquire-128-16       	   89942	     12790 ns/op	   12288 B/op	     128 allocs/op
BenchmarkEnqueueAll/enqueue/release-1-16      	11270307	       102.9 ns/op	      96 B/op	       1 allocs/op
BenchmarkEnqueueAll/enqueue/release-2-16      	 1926535	       622.9 ns/op	     240 B/op	       3 allocs/op
BenchmarkEnqueueAll/enqueue/release-8-16      	  298285	      3637 ns/op	    1104 B/op	      15 allocs/op
BenchmarkEnqueueAll/enqueue/release-64-16     	   36523	     32633 ns/op	   11472 B/op	     131 allocs/op
BenchmarkEnqueueAll/enqueue/release-128-16    	   18286	     65706 ns/op	   23760 B/op	     261 allocs/op
```

And these are the benchmarks for the current implementation (`waiter` + `Ticket`):
```
goos: darwin
goarch: amd64
pkg: github.com/conduitio/conduit/pkg/foundation/semaphore
cpu: Intel(R) Core(TM) i9-9880H CPU @ 2.30GHz
BenchmarkNewSem-16             	1000000000	         0.2308 ns/op	       0 B/op	       0 allocs/op
BenchmarkEnqueueOneByOne/acquire-1-16         	15892650	        71.66 ns/op	      96 B/op	       1 allocs/op
BenchmarkEnqueueOneByOne/acquire-2-16         	 8394110	       143.3 ns/op	     192 B/op	       2 allocs/op
BenchmarkEnqueueOneByOne/acquire-8-16         	 2123018	       570.4 ns/op	     768 B/op	       8 allocs/op
BenchmarkEnqueueOneByOne/acquire-64-16        	  243423	      4578 ns/op	    6144 B/op	      64 allocs/op
BenchmarkEnqueueOneByOne/acquire-128-16       	  129955	      9232 ns/op	   12288 B/op	     128 allocs/op
BenchmarkEnqueueAll/enqueue/release-1-16      	15322562	        73.21 ns/op	      96 B/op	       1 allocs/op
BenchmarkEnqueueAll/enqueue/release-2-16      	 6426733	       181.3 ns/op	     192 B/op	       2 allocs/op
BenchmarkEnqueueAll/enqueue/release-8-16      	 1428888	       821.6 ns/op	     768 B/op	       8 allocs/op
BenchmarkEnqueueAll/enqueue/release-64-16     	  172747	      6829 ns/op	    6144 B/op	      64 allocs/op
BenchmarkEnqueueAll/enqueue/release-128-16    	   82825	     13587 ns/op	   12288 B/op	     128 allocs/op
```

lovromazgon: I'd be happy to find a faster solution though, because this semaphore sits in the hot path, every single message will use it, so we can potentially make Conduit faster by making the semaphore faster.

lovromazgon: Ok I gave it another go and came up with an even simpler implementation, that runs as fast as the current one:
```
goos: darwin
goarch: amd64
pkg: github.com/conduitio/conduit/pkg/foundation/semaphore
cpu: Intel(R) Core(TM) i9-9880H CPU @ 2.30GHz
BenchmarkNewSem-16             	1000000000	         0.2458 ns/op	       0 B/op	       0 allocs/op
BenchmarkEnqueueOneByOne/ticket-count-1-16         	12780835	        88.15 ns/op	      96 B/op	       1 allocs/op
BenchmarkEnqueueOneByOne/ticket-count-2-16         	 6918836	       172.4 ns/op	     192 B/op	       2 allocs/op
BenchmarkEnqueueOneByOne/ticket-count-8-16         	 1724630	       712.4 ns/op	     768 B/op	       8 allocs/op
BenchmarkEnqueueOneByOne/ticket-count-64-16        	  202321	      5621 ns/op	    6144 B/op	      64 allocs/op
BenchmarkEnqueueOneByOne/ticket-count-128-16       	  104689	     11228 ns/op	   12288 B/op	     128 allocs/op
BenchmarkEnqueueOneByOne/ticket-count-1024-16      	   13489	     87872 ns/op	   98304 B/op	    1024 allocs/op
BenchmarkEnqueueAll/ticket-count-1-16              	13173651	        86.85 ns/op	      96 B/op	       1 allocs/op
BenchmarkEnqueueAll/ticket-count-2-16              	 6731160	       184.2 ns/op	     192 B/op	       2 allocs/op
BenchmarkEnqueueAll/ticket-count-8-16              	 1635159	       716.0 ns/op	     768 B/op	       8 allocs/op
BenchmarkEnqueueAll/ticket-count-64-16             	  193597	      5875 ns/op	    6144 B/op	      64 allocs/op
BenchmarkEnqueueAll/ticket-count-128-16            	  102830	     11902 ns/op	   12288 B/op	     128 allocs/op
BenchmarkEnqueueAll/ticket-count-1024-16           	   13107	     94714 ns/op	   98304 B/op	    1024 allocs/op
```

hariso:
> @hariso the current implementation is a lot simpler and has a semantic difference - `Acquire` is an idempotent function that waits until the ticket is acquired and after it returns it will simply never block again. Since the ticket needs to be discarded after `Release` is called it's not a big issue anyway, the big benefit on the other hand is that `Acquire` and `Release` don't need to acquire any lock anymore and the implementation is much simpler and easier to understand.
> 
> Benchmarks for the new implementations are on par with the previous one (see [#451 (comment)](https://github.com/ConduitIO/conduit/pull/451#discussion_r918037194)).
> 
> Please have another look when you can +1

Excellent work, Lovro!

PR #466, opened 2022-06-15
JS transforms: Add docs

### LINKED ISSUE #447: Javascript Transformations: Add docs
### Feature description

Add docs for JS transformations.

hariso:
### Description

* Adds more user docs for JS transforms. 
* Adds a runnable example (the extract-field-transform.sh script)

Fixes #447 

### Quick checks:


README.md
```diff
@@ -98,6 +99,16 @@ Kafka Connect connector as part of a Conduit pipeline.
 
 Conduit is also able to run standalone connectors. If you are interested in writing a connector yourself, have a look at our [Go Connector SDK](https://github.com/ConduitIO/conduit-connector-sdk). Since standalone connectors communicate with Conduit through gRPC they can be written in virtually any programming language, as long as the connector follows the [Conduit Connector Protocol](https://github.com/ConduitIO/conduit-connector-protocol).
 
+## Transforms
+
+Transforms are operations applied on single records. Transforms are optional parts of a pipeline and, when present, they
+come in the form of pipeline components called processors.
+
+Conduit provides a number of built-in transforms, such as filtering fields, replacing them, posting payloads to HTTP endpoints etc.
```
jayjayjpg, line 107: What do you think about the following change?

```suggestion
Conduit provides a number of built-in transforms, which can be used to filter and replace fields, post payloads to HTTP endpoints etc.
```

docs/transforms.md
```diff
@@ -0,0 +1,114 @@
...
+### Built-in transforms
+
+An up-to-date list of all built-in transforms and detailed descriptions can be found [here](https://pkg.go.dev/github.com/conduitio/conduit/pkg/processor/transform/txfbuiltin).
+An example is available in [extract-field-transform.sh](/examples/transforms/extract-field-transform.sh). The script will
+set up a pipeline with the built-in extract-field transform.
+
+### JavaScript transforms
+
+JavaScript transforms make it possible to write custom transforms. The API name for JavaScript transforms (used in the
+request to create a transform) is `js`. There's only one configuration parameters, `script`, which is the script itself.
```
jayjayjpg, line 65: I think this would be:

```suggestion
request to create a transform) is `js`. There's only one configuration parameter, `script`, which is the script itself.
```

jayjayjpg (commented):
This reads great overall and thank you for putting this together! I only made two small wording suggestions and will leave the final review to other folks on the Conduit team!

docs/transforms.md
hariso: Good one!

README.md
hariso: I like it!

examples/transforms/extract-field-transform.sh
```diff
@@ -0,0 +1,95 @@
...
+         "settings":
+         {
+             "field": "name"
+         }
+     }
+ }'
+)
+curl -Ss -X POST 'http://localhost:8080/v1/processors' -d "$TRANSFORM_REQ" > /dev/null
+
+echo > file_destination.txt
```
lovromazgon, line 70: This creates the file on the local disk, but Conduit runs in a container that doesn't have access to this file. When running the example it fails with this error:
```
node a2f165a0-34ff-495a-8d94-b500238320f4 stopped with error: could not open destination connector: open /Users/lovro/dev/conduitio/conduit/examples/transforms/file_destination.txt: no such file or directory
```

docs/transforms.md
```diff
@@ -0,0 +1,114 @@
...
+processor can be found in the [Architecture documentation](/docs/architecture.md)). As processors, transforms can be 
+attached either to a connector or to a pipeline. In other words, transforms can transform:
+1. Records coming out of a source connector (if the transform is attached to a source connector)
+2. Records coming into a destination connector (if the transform is attached to a destination connector)
+3. All records flowing through a pipeline (if the transform is attached to a pipeline)
+
+We'll go a little deeper into the case of a pipeline transform. At a high-level, source connectors, pipeline transforms
+and destination connector are laid out as in this image:
+
+![Pipeline](data/pipeline.svg)
```
lovromazgon, line 15: I see we are simplifying here and only include pipeline transforms in the drawing, I guess it's for the sake of simplicity. I'm not sure this is the right place to simplify though, I think people will come here to dive into transforms and get the full information about them. I also think the drawing can be misleading since we first explain the difference between transforms attached to a source, destination and pipeline, while the picture only shows pipeline transforms; as a user, I'd be left with questions how the other two fit in.

The very least I expect here is that we point the user to the [pipeline structure](https://github.com/ConduitIO/conduit/blob/main/docs/pipeline_semantics.md#pipeline-structure) doc where the structure is described in detail. But _ideally_ I think we should include source/destination transforms to show the full picture.

docs/transforms.md
```diff
@@ -0,0 +1,114 @@
...
+    {
+        "settings":
+        {
+            // configuration map for this processor
+            "field": "name" 
+        }
+    }
+}
+```
+The request to create a processor is described in [api.proto](/proto/api/v1/api.proto).
```
lovromazgon, line 49: I'm thinking if we should rather link to the swagger doc, since we are referencing the HTTP API?

lovromazgon (commented):
Very nice doc overall!

examples/transforms/extract-field-transform.sh
hariso: Last minute refactoring...: ) Fixed.

docs/transforms.md
hariso: Good one. I actually forgot we have them in the repo (I thought they are present only at build time and then included into a build.

docs/transforms.md
```diff
@@ -0,0 +1,116 @@
...
+
+All the helpers are defined in [helpers.go](/pkg/processor/transform/txfjs/helpers.go). Following is an example of a JavaScript
+transform:
+```javascript
+// Parses the record payload as JSON
+function parseAsJSON(record) {
+    return JSON.parse(String.fromCharCode.apply(String, record.Payload.Bytes()))
+}
+
+function transform(record) {
```
neovintage, line 103: @hariso Do I understand correctly that `transform` is the entry point for javascript functions? 

@lovromazgon have we thought about parity between our `turbine` frameworks and the primitives used there vs what we've chosen to do here? I'm thinking about `process` in turbine.

lovromazgon: You are right, `transform` is the entrypoint right now, this was designed before Turbine even existed. That said, I agree with you, we should probably revisit this design. AFAIK @hariso is working on a refactoring that will simplify processors and remove the notion of transforms/filters - maybe as part of that we can rename `tranform` to `process`, that way we move to a more intuitive name given that we will only have "processors" and we achieve parity with Turbine.

docs/transforms.md
lovromazgon: I won't block because of this, but I honestly don't see a reason why we would not show the whole picture in a doc specifically talking about transforms 🤷 For you it might have been obvious where source and destination transforms fit in and had problems understanding pipeline transforms, for someone else it might be the other way around.

docs/transforms.md
hariso: @neovintage That's correct, `transform` is the entry point for JS functions. The name can be changed at zero cost, so if we come up with a different name, it's doable. 

I can chime in a little bit about the primitives. We considered the following options (together with @jmar910 to get some insight into Turbine JS too):
1. Processor with type TRANSFORM, and a `transform` JS function **and** a processor with type FILTER.
The processor type feels a redundant piece of information, so we were looking for ways to drop it.
2. Processor (no type defined) with a single JS function, `transform` (or whatever we call it). The actual action performed by Conduit (transform or filter/drop record) is decided based on the output: if the function returns `null`, the record is dropped.
3.  Processor (no type defined), with two functions: `transform` and `filter`. This would make it analogous with `map` and `filter` used on JS arrays (and Turbine does work on arrays of records). However, James' opinion was that it would actually make it more confusing.

lovromazgon: My vote goes to option 2, sounds simple and straightforward.

hariso: FTR, we discussed the 3rd option too in details in a call, and went for option 2 for the reasons Lovro mentioned. I'll explain the details in docs.

docs/transforms.md
hariso: If you feel very strongly about that, I can definitely change. As for me, I don't think that any option is much ahead of the other one.: )

docs/processors.md
```diff
@@ -0,0 +1,158 @@
...
+processors are created by specifying the corresponding source connector as the parent entity.
+2. **Pipeline processors**: these processors receive all messages that flow through the pipeline, regardless of the
+   source or destination. Pipeline processors are created by specifying the pipeline as the parent entity.
+3. **Destination processors**: these processors receive only messages that are meant to be sent to a specific
+   destination connector. Destination processors are created by specifying the corresponding destination connector as the
+   parent entity.
+
+Given that every processor can have one (and only one) parent, processors cannot be shared. In case the same processing 
+needs to happen for different sources or destinations, you have two options:
+1. If records from all sources (or all destinations) need to be processed in the same way, then you can obviously create
```
lovromazgon, line 22: I would remove "obviously".

docs/processors.md
```diff
@@ -0,0 +1,158 @@
...
+   // we will keep it.
+   // otherwise we return null (i.e. we drop the record from the pipeline)
+    if (r.Metadata["keepme"] != undefined) {
+        return r
+    }
+    return null;
+}
+```
+
+The script, of course, is not constrained to having only this function, i.e. you can have something like this:
```
lovromazgon, line 114: I would remove "of course".

docs/processors.md
```diff
@@ -0,0 +1,158 @@
...
+it in Go code, i.e. you can write this for example: `logger.Info().Msgf("hello, %v!", "world")`
+2. `Record()` - constructs a `record.Record`.
+3. `RawData()` - constructs `record.RawData`.
+
+Following is an example of a JavaScript processor, where we transform a record and utilize a number of tools mentioned 
+above:
+```javascript
+// Parses the record payload as JSON
+function parseAsJSON(record) {
+    // we can, of course, use all of the JavaScript built-in functions
```
lovromazgon, line 138: I would remove "of course".

docs/processors.md
```diff
@@ -0,0 +1,158 @@
...
+
+### Built-in processors
+
+An up-to-date list of all built-in processors and detailed descriptions can be found [here](https://pkg.go.dev/github.com/conduitio/conduit/pkg/processor/procbuiltin).
+An example is available in [extract-field-transform.sh](/examples/processors/extract-field-transform.sh). The script will
+set up a pipeline with the built-in extract-field processors.
+
+### JavaScript processors
+
+JavaScript processors make it possible to write custom processors. The API name for JavaScript processors (used in the
```
lovromazgon, line 69: I think it would be great if we mention that we use https://github.com/dop251/goja and that users can go there to find out more about what is possible and what is not.

lovromazgon (approved):
No blockers from my side, just suggestions how to make it _even better_ 🙃 Nice doc!

docs/processors.md
```diff
@@ -0,0 +1,158 @@
...
+
+## Adding and configuring a processor
+
+Processors are created through the `/processors` endpoint. Here's an example:
+
+```json lines
+POST /v1/processors
+{
+    // name of the processor in Conduit
+    // note that this is NOT a user-defined name for this processor
```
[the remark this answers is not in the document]
hariso: Yup, that was my thought as well!

.gitignore
```diff
@@ -63,3 +63,6 @@ conduit.db
 .DS_Store
 
 dist/
+
+# Examples
+file_destination.txt
```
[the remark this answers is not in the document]
hariso: Good one, done in ea295a5.

examples/processors/extract-field-transform.sh
```diff
@@ -0,0 +1,103 @@
...
+# * a generator source, generating structured records
+# * a file destination (it writes to file_destination.txt)
+# * a processor with the built-in extract-field transform.
+
+SCRIPT_DIR=$( cd -- "$( dirname -- "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )
+DEST_FILE="$SCRIPT_DIR/file_destination.txt"
+> "$DEST_FILE"
+
+echo "Running Conduit..."
+docker run --rm  --name conduit -v "$DEST_FILE":/file_destination.txt -p 8080:8080 -d ghcr.io/conduitio/conduit:latest
```
hariso, line 15: Since this pulls the latest Docker image, this won't work until a nightly build with this change is out. I've verified this with a local image.

docs/processors.md
hariso: I like to add a little bit of less formal vocabulary, because otherwise (at least to me), the docs can become quite boring.: ) But I completely see the arguments for having it more formal too.

PR #471, opened 2022-06-17
JavaScript filters

### LINKED ISSUE #44: Entity locking in orchestration layer
Initially, each service (pipeline, connector, processor service) had its own instance lock. The problem was twofold - the services locked the instances only for the duration of the operation inside the service and they did not lock any related instance. Let's take the creation of a processor as an example: it requires us to get the pipeline, lock it in place so it's not modified in the meantime, create the processor, add it to the pipeline, commit the transaction to the DB and only then release the lock.

Now that we have the orchestration layer it should be the responsibility of this layer to lock any entities that will be modified in an operation before making changes. The goal is to make Conduit safe for concurrent use, i.e. multiple requests changing the same resource at the same time.

hariso:
### Description

This PR is also extracting the JS related code into a single place, which is making the transforms and filters basically converge.

Fixes #44 

### Quick checks:

### DIFF
--- a/pkg/conduit/runtime.go
+++ b/pkg/conduit/runtime.go
@@ -59,6 +59,7 @@ import (
 	"gopkg.in/tomb.v2"
 
 	// NB: anonymous import triggers transform registry creation
+	_ "github.com/conduitio/conduit/pkg/processor/filter/filterjs"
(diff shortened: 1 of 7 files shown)


hariso:
Closing in favor of #486.

PR #477, opened 2022-06-21
Ignore untracked files when adding suffix dirty to version

lovromazgon:
### Description

I noticed the version of Conduit in our released Docker images is wrong, it has the suffix `-dirty`. The latest nightly build prints the version `v0.3.0-nightly.20220621-dirty` when Conduit starts.

The same happens if you build the image yourself locally. This is _possibly_ because Docker creates temporary files while building the image (not entirely sure). What I am sure though is that this PR solves the problem for local builds, the suffix `-dirty` will now only be added if some tracked file was changed, it ignores untracked files. I assume it will solve it for CI builds as well.

### Quick checks:


Makefile
```diff
@@ -1,6 +1,6 @@
 .PHONY: test test-integration build run proto-api proto-lint clean download install-tools generate check-go-version
 
-VERSION=`./scripts/get-tag.sh`
+VERSION=`git describe --tags --dirty`
```
hariso, line 3: The reason why I used the script with `git diff-index` was that not all changes I made locally were being detected by `git describe`. I'll try to find what those changes were.

lovromazgon: IIRC the difference was to detect untracked files, but that seems to be a bit too strict for Docker.

hariso: Right... Personally, I'd like the tag to still be dirty, when files are added (simply to be sure there are no unintentional commits or such). On the other hand, everything which already got into GitHub is tracked, so whatever is untracked can only be generated files, which we don't really care about.

hariso: Btw, if you don't mind leaving a comment explaining why we have the command as it is, that'd be really useful.

hariso (approved):

Makefile
lovromazgon: I added a doc above the command 👍

PR #482, opened 2022-06-22
parse config file from yaml to Go

### LINKED ISSUE #32: Pipeline file

maha-hajja:
### Description

Fixes #32 

### Quick checks:

### DIFF
--- a/pkg/conduit/pipelines.yml
+++ b/pkg/conduit/pipelines.yml
@@ -1,0 +1,44 @@
+---
+- status: STATUS_STOPPED
+  config:
+    name: p1e2ee
+    description: hif
+  connectors: []
+  processors: []
+- status: STATUS_RUNNING
+  config:
+    name: two
+    description: twoo
+  connectors:
+    - type: TYPE_DESTINATION
+      plugin: 'builtin:s3'
+      config:
+        name: s3-dest
+        settings:
+          aws.region: us-east-1
+          aws.bucket: maha-test
+    - type: TYPE_SOURCE
+      plugin: 'builtin:pg'
+      config:
+        name: pg-source
+        settings:
+          pg.region: us-east-1
+          pg.bucket: maha-test
+  processors:
+    - name: d
+      type: TYPE_TRANSFORM
+      parent:
+        type: TYPE_CONNECTOR
+        name: s3-dest # name instead of id, parent not created yet
+      config:
+        settings:
+          additionalProp1: string
+          additionalProp2: string
+          additionalProp3: string
+---
+- status: TEST
+  config:
+    name: p1e2ee
+    description: hif
+  connectors: []
+  processors: []
--- a/pkg/conduit/pipelines_parser.go
+++ b/pkg/conduit/pipelines_parser.go
@@ -1,0 +1,75 @@
+// Copyright © 2022 Meroxa, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// Package conduit wires up everything under the hood of a Conduit instance
+// including metrics, telemetry, logging, and server construction.
+// It should only ever interact with the Orchestrator, never individual
+// services. All of that responsibility should be left to the Orchestrator.
+package conduit
+
+import (
+	"bytes"
+	"fmt"
+	"gopkg.in/yaml.v3"
+)
+
+type Pipelines []struct {
+	Status string `yaml:"status"`
+	Config struct {
+		Name        string `yaml:"name"`
+		Description string `yaml:"description"`
+	} `yaml:"config"`
+	Connectors []struct {
+		Type   string `yaml:"type"`
+		Plugin string `yaml:"plugin"`
+		Config struct {
+			Name     string      `yaml:"name"`
+			Settings interface{} `yaml:"settings"`
+		} `yaml:"config"`
+	} `yaml:"connectors,omitempty"`
+	Processors []struct {
+		Name   string `yaml:"name"`
+		Type   string `yaml:"type"`
+		Parent struct {
+			Type string `yaml:"type"`
+			Name string `yaml:"name"`
+		} `yaml:"parent"`
+		Config struct {
+			Settings interface{} `yaml:"settings"`
+		} `yaml:"config"`
+	}
+}
+
+func Parse(data []byte) ([]Pipelines, error) {
+	//p := Pipelines{}
+	//
+	//err := yaml.Unmarshal(data, &p)
+	//if err != nil {
+	//	return nil, cerrors.Errorf("error: %w", err)
+	//}
+	//return &p, nil
+
+	dec := yaml.NewDecoder(bytes.NewReader(data))
+
+	var docs []Pipelines
+	for {
+		var doc Pipelines
+		if err := dec.Decode(&doc); err != nil {
+			fmt.Println("stopping because of error:", err)
+			break
+		}
+		docs = append(docs, doc)
+	}
+	return docs, nil
+}
--- a/pkg/conduit/pipelines_parser_test.go
+++ b/pkg/conduit/pipelines_parser_test.go
@@ -1,0 +1,40 @@
+// Copyright © 2022 Meroxa, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package conduit
+
+import (
+	"fmt"
+	"io/ioutil"
+	"path/filepath"
+	"testing"
+)
+
+func Test_Parser(t *testing.T) {
+	filename, err := filepath.Abs("./pipelines.yml")
+	if err != nil {
+		t.Error(err)
+	}
+
+	data, err := ioutil.ReadFile(filename)
+	if err != nil {
+		t.Error(err)
+	}
+
+	p, err := Parse(data)
+	if err != nil {
+		t.Error(err)
+	}
+	fmt.Printf("--- t:\n%v\n\n", p)
+}


pkg/conduit/pipelines_parser.go
```diff
@@ -0,0 +1,75 @@
...
+	Processors []struct {
+		Name   string `yaml:"name"`
+		Type   string `yaml:"type"`
+		Parent struct {
+			Type string `yaml:"type"`
+			Name string `yaml:"name"`
+		} `yaml:"parent"`
+		Config struct {
+			Settings interface{} `yaml:"settings"`
+		} `yaml:"config"`
```
maha-hajja, line 50: note: all of these ``yaml:"config"`` can be removed.. the default is the field name lowercased

pkg/conduit/pipelines_parser.go
```diff
@@ -0,0 +1,75 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// Package conduit wires up everything under the hood of a Conduit instance
+// including metrics, telemetry, logging, and server construction.
+// It should only ever interact with the Orchestrator, never individual
+// services. All of that responsibility should be left to the Orchestrator.
```
lovromazgon, line 18: We already have a package doc in `runtime.go` (probably where you copied it from), no need to add it here.

pkg/conduit/pipelines_parser.go
```diff
@@ -0,0 +1,75 @@
...
+package conduit
+
+import (
+	"bytes"
+	"fmt"
+	"gopkg.in/yaml.v3"
+)
+
+type Pipelines []struct {
+	Status string `yaml:"status"`
```
lovromazgon, line 28: I think we will need an identifier field for each entity (pipeline, connector, processor) to be able to distinguish them between runs. For example, if a config file changes and we reload it we need to be able to know which entity changed, because we will have the previous pipeline config already stored somewhere. Or if a config file is removed we need to know which pipeline to remove. All of that can't be deterministically done without an identifier field (happy to hear suggestions if you think it is 🙂).

I see two options:
- either we add an `id` field to each entity
  ```
  connectors:
    - id: my-source
      type: source
      ...
    - id: my-destination
      type: destination
      ...
  ```
- or we change the array into a map, where the key of the map is the ID of the connector/processor
  ```
  connectors:
    my-source:
      type: source
      ...
    my-destination:
      type: destination
      ...
  ```

The second one looks nicer to me. If we decide on the second option, I think it might make sense to add a new root element `pipelines` and allow creating multiple in the same file (I was against that at first, but with this approach it seems it would make for a more complete design). For example:
```
pipelines:
  my-pipeline:
    state: running
    connectors:
      my-source:
        type: source
        ...
      my-destination:
        type: destination
        ...
    processors:
      my-processor:
        ...
  my-other-pipeline:
    ...
```

WDYT @ConduitIO/conduit-core?

lovromazgon (commented):
Going in the right direction 👍

pkg/conduit/pipelines_parser.go
maha-hajja: thought that I only copied the license :\

hariso: @maha-hajja I can suggest using GoLand's live templates. I've saved the Meroxa copyright notice as `lcns`, and whenever I need it, I type `lcns` and GoLand inserts it.

pkg/conduit/pipelines_parser.go
hariso: IMHO, that depends on the use case. Conduit always being in sync with the files appears to be the problem which we want to solve with the above. For v1 I think it's useful to just have the ability to load pipeline(s) and either fail or overwrite if a pipeline with the same config exists.

hariso: Also, re: status: the only two status which we can have here are actually "stopped" and "running". Having a degraded pipeline in a config is a bit strange.

pkg/conduit/pipelines_parser.go
```diff
@@ -0,0 +1,75 @@
...
+	Config struct {
+		Name        string `yaml:"name"`
+		Description string `yaml:"description"`
+	} `yaml:"config"`
+	Connectors []struct {
+		Type   string `yaml:"type"`
+		Plugin string `yaml:"plugin"`
+		Config struct {
+			Name     string      `yaml:"name"`
+			Settings interface{} `yaml:"settings"`
```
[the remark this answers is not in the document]
maha-hajja: works fine

pkg/conduit/pipelines_parser.go
lovromazgon: > fail or overwrite if a pipeline with the same config exists

The "overwrite" part is what I'm talking about. If we just overwrite (i.e. delete the previous pipeline and create a new one) we lose the state of the pipeline. I'm not talking about periodically hot reloading files (although that would also be nice), but restarting Conduit and reloading pipeline config files - we still need to be sure that we don't lose the position the pipeline processed last.

> Having a degraded pipeline in a config is a bit strange.

I was thinking about this as well, almost wrote a comment about it but didn't after all. First of all, "stopped" isn't the same as "degraded", because there would be no error stored in the pipeline. I think it might be useful for cases when you want to temporarily pause a pipeline but keep it created in Conduit for some later time. Or maybe it's a pipeline that you want to have prepared and start/stop manually on demand. So I see some use cases for this field.

That said, I think the field should be optional and should default to "running" if not specified otherwise (that shouldn't be part of the parser).

maha-hajja: I think having an id for each entity is good in general, especially because we need the id for the parent in the processors and we didn't have that, also for the reason lovro mentioned (even if we dismiss it for the first version).
 About the preferred format for the yaml file, I don't have a strong preference honestly, first one looks easier for the user to fill {it's like a form that you have all the questions (the keys) and you fill the value}, the second approach is more of you decide what the name of the key is, and fill the value, which would only need from us to have good documentation showing that the id has to be unique.

pkg/conduit/pipelines_parser.go
```diff
@@ -0,0 +1,75 @@
...
+		} `yaml:"config"`
+	} `yaml:"connectors,omitempty"`
+	Processors []struct {
```
lovromazgon, line 41: Proposal - what if we have the same exact field nested inside of the `connectors` field? So if you want to add processors to a connector you specify it under that connector, if you want to add the processor to the pipeline you specify it here. WDYT?

maha-hajja:
while designing this I was thinking of another approach that I thought wouldn't be parsable, but now I think it is.
it's basically having three root entities (pipelines, connectors, processors) and each one has an id. And the pipelines would look sth like this: 
```
pipelines:
  - pipeline1:
     state: running
     connectorIds:
          - con1
          - con2
```
so it would have a list of ids, instead of nesting everything in one object. what do you think? @lovromazgon @hariso

pkg/conduit/pipelines_parser.go
maha-hajja: @hariso @lovromazgon thank you for the ideas 👀

pkg/conduit/pipelines_parser.go
```diff
@@ -0,0 +1,87 @@
...
+}
+
+// mergeMaps takes an array of PipelinesConfig and merges them into one map
+func mergeMaps(arr []PipelinesConfig) PipelinesConfig {
+	var mp PipelinesConfig
+	pipelines := make(map[string]PipelineConfig, 0)
+
+	for _, config := range arr {
+		for k, v := range config.Pipelines {
+			pipelines[k] = v
```
lovromazgon, line 82: We should first check if `pipielines[k]` already exists and return an error if it does, otherwise we might ignore one pipeline config. I'd rather be loud about it than silently lose one config file.

pkg/conduit/pipelines.yml
```diff
@@ -0,0 +1,71 @@
+---
+pipelines:
+  pipeline1:
+    status: running
+    config:
+      name: pipeline1
+      description: nothing
+    connectors:
+      con1:
+        type: TYPE_DESTINATION
```
lovromazgon, line 10: Didn't we say we'd use more "human-friendly" values for types?

pkg/conduit/pipelines_parser_test.go
```diff
@@ -0,0 +1,41 @@
...
+	data, err := ioutil.ReadFile(filename)
+	if err != nil {
+		t.Error(err)
+	}
+
+	p, err := Parse(bytes.NewReader(data))
+	if err != nil {
+		t.Error(err)
+	}
+	fmt.Printf("--- t:\n%v\n\n", p)
```
lovromazgon, line 40: Let's assert that what we parsed is what we expected.

pkg/conduit/pipelines_parser.go
```diff
@@ -0,0 +1,87 @@
...
+			Name     string            `yaml:"name"`
+			Settings map[string]string `yaml:"settings"`
+		} `yaml:"config"`
+		Processors map[string]struct {
+			Name   string `yaml:"name"`
+			Type   string `yaml:"type"`
+			Config struct {
+				Settings map[string]string `yaml:"settings"`
+			} `yaml:"config"`
+		} `yaml:"processors,omitempty"`
```
lovromazgon, line 42: Nitpick: we could reuse this struct in the pipeline. Actually we could create separate structs to make this more readable, like `ConnectorConfig` and `ProcessorConfig` and then just plug them together in a `PipelineConfig`.

pkg/conduit/pipelines.yml
```diff
@@ -0,0 +1,71 @@
+---
```
lovromazgon, line 1: Another thing we might want to add is a field like `version` which would right now just be `version: '1'` to already make it possible to evolve this schema in the future. Essentially the same as [docker-compose](https://docs.docker.com/compose/compose-file/compose-versioning/) does it. WDYT?

pkg/conduit/pipelines.yml
maha-hajja: yeah it's just a string.. forgot to change it

pkg/conduit/pipelines.yml
maha-hajja: @lovromazgon does it have to be the same version for all the yaml docs? cuz if not, then we can't merge all the docs in one map, since they could have different versions. WDYT?

maha-hajja: we also should decide on what the default value for the version should be, when it's not provided

lovromazgon: Right now we will only have a single parser, which only knows how to parse the config file of version 1. If the version string contains something else than `1` or `1.0` (I think it would make sense to have minor and major versions) the parser should probably fail because it can't be sure it can parse that document.

In the future when we create a new schema version we would change the parser and allow additional values in the version field (e.g. `1.1`) or create a new parser alongside the existing one (e.g. if we bump the major version to `2`), and some logic around that. But let's not worry about that, we can do all of this when/if that ever happens. For now we can assume we have this one parser and we can merge all parsed documents.

> what the default value for the version should be

Why not make the field mandatory? I mean the parser shouldn't care, but in the second step we'll need to validate whatever the parser read, and at that point we would raise an error if the `version` field is not specified.

pkg/conduit/pipelines_parser.go
```diff
@@ -0,0 +1,111 @@
...
+}
+
+func Parse(data io.Reader) (PipelinesInfo, error) {
+	dec := yaml.NewDecoder(data)
+
+	var docs []PipelinesInfo
+	for {
+		var doc PipelinesInfo
+		err := dec.Decode(&doc)
+		if err != nil && cerrors.Is(err, io.EOF) {
```
hariso, line 72: Is `err != nil` needed here?

pkg/conduit/pipelines_parser.go
```diff
@@ -0,0 +1,111 @@
...
+
+// mergeMaps takes an array of PipelinesInfo and merges them into one map
+func mergeMaps(arr []PipelinesInfo) (PipelinesInfo, error) {
+	var mp PipelinesInfo
+	pipelines := make(map[string]PipelineInfo, 0)
+
+	for _, config := range arr {
+		for k, v := range config.Pipelines {
+			if _, ok := pipelines[k]; ok {
+				return PipelinesInfo{}, cerrors.Errorf("found a duplicated pipeline id: %s", k)
```
hariso, line 103: IIUC, each yaml file can contain multiple pipelines. If that's true, it would be helpful to know which pipeline is duplicated.

pkg/conduit/pipelines_parser_test.go
```diff
@@ -0,0 +1,151 @@
...
+	"bytes"
+	"io/ioutil"
+	"path/filepath"
+	"testing"
+
+	"github.com/conduitio/conduit/pkg/foundation/assert"
+)
+
+func Test_Parser(t *testing.T) {
+	filename, err := filepath.Abs("./test/pipelines1.yml")
```
hariso, line 27: Nitpick: it might be good to have the test yml files named after the use case (e.g. test/pipelines_duplicated.yml)

pkg/conduit/pipelines_parser_test.go
```diff
@@ -0,0 +1,151 @@
...
+		t.Error(err)
+	}
+
+	data, err := ioutil.ReadFile(filename)
+	if err != nil {
+		t.Error(err)
+	}
+
+	p, err := Parse(bytes.NewReader(data))
+	assert.Error(t, err)
```
hariso, line 133: Ideally, whenever there's an error, we also make sure that the correct error is returned too. I was just looking at pipeline2.yml, and the version of the first doc is 1.1, so I have a feeling the error we have in the test here is not about a duplicated pipeline, but it's about an invalid version.

pkg/conduit/pipelines_parser_test.go
```diff
@@ -0,0 +1,151 @@
...
+	assert.Equal(t, p, PipelinesInfo{})
+}
+
+func Test_Parser_Invalid_Version(t *testing.T) {
```
hariso, line 137: Maybe also add a test for missing version?

pkg/conduit/pipelines_parser_test.go
```diff
@@ -0,0 +1,151 @@
...
+
+import (
+	"bytes"
+	"io/ioutil"
+	"path/filepath"
+	"testing"
+
+	"github.com/conduitio/conduit/pkg/foundation/assert"
+)
+
```
hariso, line 25: I'd like to suggest also a test for an empty file and also an invalid yaml.

pkg/conduit/pipelines_parser.go
```diff
@@ -0,0 +1,111 @@
...
+	Pipelines map[string]PipelineInfo `yaml:"pipelines"`
+}
+
+func Parse(data io.Reader) (PipelinesInfo, error) {
+	dec := yaml.NewDecoder(data)
+
+	var docs []PipelinesInfo
+	for {
+		var doc PipelinesInfo
+		err := dec.Decode(&doc)
```
hariso, line 71: How does the decoder handle unknown fields? Does it ignore them or does it fail?

hariso:
@maha-hajja Great PR!

pkg/conduit/pipelines_parser_test.go
```diff
@@ -0,0 +1,151 @@
...
+	assert.Equal(t, want, got)
+}
+
+func Test_Parser_Duplicate_Pipeline_Id(t *testing.T) {
+	filename, err := filepath.Abs("./test/pipelines2.yml")
+	if err != nil {
+		t.Error(err)
+	}
+
+	data, err := ioutil.ReadFile(filename)
```
lovromazgon, line 127: Nitpick: this comment is not so important for a test, I'm still going to point it out because the upcoming PRs will need to do the same in production code and we should do it differently there. I'm talking about `ioutil.ReadFile` - this function will read the whole file into memory, then `Parse` will decode the file. It would be more efficient to just decode it while it's being read, without the need to put the whole file into memory. This can be done by passing the file directly to `Parse`, because it expects an `io.Reader` and `os.File` implements that interface:
```
file, err := os.Open("path/to/my/pipeline_file.yml")
if err != nil { ... } // handle error
defer file.Close()    // close file once we are done with it
parsedFile, err := Parse(file)
```

pkg/conduit/pipelines_parser.go
```diff
@@ -0,0 +1,111 @@
...
+		} else if err != nil {
+			return PipelinesInfo{}, cerrors.Errorf("parsing error: %w", err)
+		}
+		// check if version is empty
+		if doc.Version == "" {
+			return PipelinesInfo{}, cerrors.New("version field is not specified")
+		}
+		// check if version is invalid
+		if doc.Version != "1" && doc.Version != ParserVersion {
+			return PipelinesInfo{}, cerrors.Errorf("invalid version : %s , try version %s instead", doc.Version, ParserVersion)
```
lovromazgon, line 83: Nitpick: the version is _unsupported_, not necessarily _invalid_.

pkg/conduit/pipelines_parser.go
lovromazgon: The error contains the pipeline ID, what else did you have in mind?

pkg/conduit/pipelines_parser.go
maha-hajja: it ignores them

pkg/conduit/pipelines_parser_test.go
maha-hajja: @lovromazgon  this works yes, but I'm still gonna need to read the file because of the env vars, check https://github.com/ConduitIO/conduit/pull/497 
I changed the method back to accept `[]byte `because that's what's needed for the `ExpandEnv` method.
I could still pass the `io.Reader `but I'm gonna have to read it and convert it to `[]byte` just so I could get the env vars.
WDYT?

lovromazgon: Alright let's go with this now, even though it's not ideal, it's good enough!

lovromazgon (commented):
We're almost there, good work with the tests! 👍 We can also move this code to the package `pkg/provisioning`, or do it in a later PR, whatever is easier.

lovromazgon (approved):
🙌 great work @maha-hajja!

PR #483, opened 2022-06-22
Source Acker Node

### LINKED ISSUE #393: Feature: Ack/nack handler node

### LINKED ISSUE #394: Feature: Remove message status "dropped"
After introducing the ack/nack handler node we centralized the decision if a message should be dropped or not, making the dropped status obsolete. Nodes should only be able to signal if the message was processed successfully/unsuccessfully, deciding what happens next should not be in their hands.

Depends on #393

lovromazgon:
### Description

This PR adds a `SourceAckerNode` that is responsible for forwarding acks to the source connector. It will also be responsible for handling nacks in the future and forwarding them to the DLQ.

The PR also simplifies the message status change handler by removing middleware and using simple handlers instead. This ensures that all handlers are called and errors are collected (previously a middleware could stop the downstream middleware from being called).

Depends on https://github.com/ConduitIO/conduit/pull/451.

Fixes #393

### Quick checks:


pkg/pipeline/stream/message.go
```diff
@@ -150,9 +127,9 @@ func (m *Message) ID() string {
 
 // RegisterStatusHandler is used to register a function that will be called on
 // any status change of the message. This function can only be called if the
-// message status is open, otherwise it panics. Middlewares are called in the
+// message status is open, otherwise it panics. Handlers are called in the
 // reverse order of how they were registered.
-func (m *Message) RegisterStatusHandler(mw StatusChangeMiddleware) {
+func (m *Message) RegisterStatusHandler(mw StatusChangeHandler) {
```
hariso, line 132: Nitpick: we can rename mw (middleware) to whatever sounds better for a handler.

hariso (approved):
Great work, Lovro! :rocket:

PR #486, opened 2022-06-23
JS filters

### LINKED ISSUE #448: JavaScript Processors: Support for filtering records
### Feature description

Currently, we are not able to write a JavaScript filter for records (i.e. drop records based on criteria written in JS).

hariso:
### Description

This PR implements JS filters in the following way: a function with a predefined name (transform, to be changed to process) needs to provided. If the function returns a record, it's considered a transformation. If the function returns `null`, the record is dropped. Returning something else results in a processing error.

Additionally:
* The package structure is a bit flattened, packages renamed
* processor type is removed

ToDos:
- [x] rename transform into process
- [x] rename txfjs into jsprocessor
- [ ] rename processor.name to processor.type -- will be done separately: https://github.com/ConduitIO/conduit/issues/498
- [ ] update docs -- will be done separately
- [x] test metrics

Fixes #448 

### Quick checks:

- [ ] I have made sure that the PR is of reasonable size and can be easily reviewed. -- it's big, but it's also possible to split it up. Review has already started, so we're leaving it as it is.

### DIFF
--- a/ui/package.json
+++ b/ui/package.json
@@ -38,7 +38,7 @@
     "d3-selection": "^3.0.0",
     "d3-zoom": "^3.0.0",
     "ember-auto-import": "^2.4.2",
-    "ember-basic-dropdown": "^4.0.3",
+    "ember-basic-dropdown": "^6.0.1",
     "ember-changeset": "^3.15.0",
     "ember-changeset-validations": "^3.16.0",
     "ember-cli": "~4.1.1",
--- a/ui/yarn.lock
+++ b/ui/yarn.lock
@@ -1293,7 +1293,7 @@
     ember-cli-babel "^7.10.0"
     ember-modifier-manager-polyfill "^1.1.0"
 
-"@ember/render-modifiers@^1.0.2 || ^2.0.0", "@ember/render-modifiers@^2.0.3", "@ember/render-modifiers@^2.0.4":
+"@ember/render-modifiers@^1.0.2 || ^2.0.0", "@ember/render-modifiers@^2.0.4":
   version "2.0.4"
   resolved "https://registry.yarnpkg.com/@ember/render-modifiers/-/render-modifiers-2.0.4.tgz#0ac7af647cb736076dbfcd54ca71e090cd329d71"
   integrity sha512-Zh/fo5VUmVzYHkHVvzWVjJ1RjFUxA2jH0zCp2+DQa80Bf3DUXauiEByxU22UkN4LFT55DBFttC0xCQSJG3WTsg==
@@ -1352,19 +1352,7 @@
     "@embroider/shared-internals" "^1.8.0"
     semver "^7.3.5"
 
-"@embroider/macros@0.41.0", "@embroider/macros@^0.41.0":
-  version "0.41.0"
-  resolved "https://registry.yarnpkg.com/@embroider/macros/-/macros-0.41.0.tgz#3e78b6f388d7229906abf4c75edfff8bb0208aca"
-  integrity sha512-QISzwEEfLsskZeL0jyZDs1RoQSotwBWj+4upTogNHuxQP5j/9H3IMG/3QB1gh8GEpbudATb/cS4NDYK3UBxufw==
-  dependencies:
-    "@embroider/shared-internals" "0.41.0"
-    assert-never "^1.1.0"
-    ember-cli-babel "^7.23.0"
-    lodash "^4.17.10"
-    resolve "^1.8.1"
-    semver "^7.3.2"
-
-"@embroider/macros@1.8.0", "@embroider/macros@^1.0.0", "@embroider/macros@^1.6.0", "@embroider/macros@^1.8.0":
+"@embroider/macros@1.8.0", "@embroider/macros@^0.50.0 || ^1.0.0", "@embroider/macros@^1.0.0", "@embroider/macros@^1.2.0", "@embroider/macros@^1.6.0", "@embroider/macros@^1.8.0":
   version "1.8.0"
   resolved "https://registry.yarnpkg.com/@embroider/macros/-/macros-1.8.0.tgz#9e947e6292f0d7768a0f743d883c3e794457eaec"
   integrity sha512-jeRoUMPbZ/ADBU0FsyBGS1UgV5vQmwIPnpbVUsJ0XcZFZq/8ZHV6Ts4yeubvH/lePnUpHaKogs4dDvGk5tpQKw==
@@ -1376,6 +1364,18 @@
     find-up "^5.0.0"
     lodash "^4.17.21"
     resolve "^1.20.0"
+    semver "^7.3.2"
+
+"@embroider/macros@^0.41.0":
+  version "0.41.0"
+  resolved "https://registry.yarnpkg.com/@embroider/macros/-/macros-0.41.0.tgz#3e78b6f388d7229906abf4c75edfff8bb0208aca"
+  integrity sha512-QISzwEEfLsskZeL0jyZDs1RoQSotwBWj+4upTogNHuxQP5j/9H3IMG/3QB1gh8GEpbudATb/cS4NDYK3UBxufw==
+  dependencies:
+    "@embroider/shared-internals" "0.41.0"
+    assert-never "^1.1.0"
+    ember-cli-babel "^7.23.0"
+    lodash "^4.17.10"
+    resolve "^1.8.1"
     semver "^7.3.2"
 
 "@embroider/macros@^0.42.3":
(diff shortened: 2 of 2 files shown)


proto/api/v1/api.proto
```diff
@@ -120,11 +112,10 @@ message Processor {
   // -- immutable fields -------------------------------------------------------
 
   string name = 4 [(google.api.field_behavior) = IMMUTABLE];
-  Type type = 5 [(google.api.field_behavior) = IMMUTABLE];
-  Parent parent = 6 [(google.api.field_behavior) = IMMUTABLE];
+  Parent parent = 5 [(google.api.field_behavior) = IMMUTABLE];
```
lovromazgon, line 115: IMO we shouldn't change existing field numbers as that would be a breaking change. I would also prefer to have nicely numbered fields of course, but it's not worth it if it introduces a breaking change. I'd say we can go through the proto files before releasing version 1 of Conduit and clean them up since we would be bumping a major version at that point.

pkg/processor/jsprocessor/jsprocessor.go
```diff
@@ -0,0 +1,184 @@
...
+	entrypoint = "process"
+)
+
+// jsProcessor is able to run transformations defined in JavaScript.
+type jsProcessor struct {
+	runtime  *goja.Runtime
+	function goja.Callable
+}
+
+func NewJSProcessor(src string, logger zerolog.Logger) (processor.Processor, error) {
```
lovromazgon, line 37: A couple of nitpicks here:
- Keep in mind that every call to this package will contain the package name, so `jsProcessor.NewJSProcessor` sounds like stuttering (obligatory [reference link](https://github.com/golang/go/wiki/CodeReviewComments#package-names)).
- The general rule of thumb is to accept interfaces and return concrete types, meaning that here it would be better to not hide `jsProcessor` behind `processor.Processor`, but rather to export `jsProcessor` and return the concrete type, letting the caller decide if they want to use it through the `processor.Processor` interface.

pkg/processor/builtin/util.go
```diff
@@ -12,28 +12,40 @@
...
-	"github.com/conduitio/conduit/pkg/processor/transform"
 	"github.com/conduitio/conduit/pkg/record"
 )
 
+// funcProcessor is a stateless processor.Processor implementation
+// which is using a Go function to process records.
+// todo move out of this file.
+type funcProcessor struct {
+	fn func(context.Context, record.Record) (record.Record, error)
+}
```
lovromazgon, line 32: A nicer way to do this would be to use the same approach as [`http.HandlerFunc`](https://cs.opensource.google/go/go/+/refs/tags/go1.18.3:src/net/http/server.go;l=2080) ([godoc](https://pkg.go.dev/net/http#HandlerFunc)), and because `HandlerFunc` is so widely used I'd also rename it to `ProcessorFunc` to match the same naming.
```suggestion
type ProcessorFunc func(context.Context, record.Record) (record.Record, error)
```

pkg/processor/service.go
```diff
@@ -120,7 +115,7 @@ func (s *Service) Create(
 	}
 
 	s.instances[instance.ID] = instance
-	measure.ProcessorsGauge.WithValues(name, t.String()).Inc()
+	measure.ProcessorsGauge.WithValues(name).Inc()
```
lovromazgon, line 118: Don't forget to change the `ProcessorsGauge` and remove the `type` label.

proto/api/v1/api.proto
hariso: Sure, I can revert. Question before that: do we care about breaking changes at this point?

pkg/processor/jsprocessor/builder.go
```diff
@@ -12,12 +12,11 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-package txfjs
+package jsProcessor
```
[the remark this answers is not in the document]
hariso: I was thinking about `proc` too, as `processor` and `jsprocessor` are a bit long and "heavy". The reason why I left `processor` is because it describes a thing in the pipeline (similar to: source, destination, connector), and `proc` is used as an acronym for process. But that's not to say I don't like `procX`.: )

proto/api/v1/api.proto
lovromazgon: Not that much really, but it's an easy one to avoid.

pkg/processor/builtin/util.go
hariso: Hah, I was actually looking for a way to "wrap" all the built-in functions, and actually already had this. What I was missing is the `Execute` method, but for whatever reason I was convinced that I can't add any methods to `type Foo func()` :S

hariso: How about we move this type to instance.go, where we have the Processor interface?

lovromazgon: Makes sense, do it!

pkg/processor/service.go
hariso: Done.

pkg/processor/jsprocessor/jsprocessor.go
```diff
@@ -0,0 +1,184 @@
...
+
+	"github.com/conduitio/conduit/pkg/foundation/cerrors"
+	"github.com/conduitio/conduit/pkg/processor"
+	"github.com/conduitio/conduit/pkg/record"
+	"github.com/dop251/goja"
+	"github.com/rs/zerolog"
+)
+
+const (
+	entrypoint = "process"
```
lovromazgon: ❤️

pkg/processor/instance.go
```diff
@@ -25,29 +24,27 @@ import (
...
 // can be executed on one record and manipulate it.
-type Processor interface {
-	// Type returns the processor type.
-	Type() Type
-	// Execute runs the processor function on a record.
-	Execute(ctx context.Context, record record.Record) (record.Record, error)
+type Interface interface {
+	// Process runs the processor function on a record.
+	Process(ctx context.Context, record record.Record) (record.Record, error)
+}
```
lovromazgon, line 40: ❤️

pkg/processor/instance.go
```diff
@@ -25,29 +24,27 @@ import (
...
+
+// InterfaceFunc is a stateless processor.Interface implementation
```
lovromazgon, line 42: Nitpick of the highest order: it isn't _necessarily_ stateless 🙈
```
func NewStatefulProcessorFunc() processor.InterfaceFunc {
	myState := 0
	return func(_ context.Context, r record.Record) (record.Record, error) {
		myState++
		return r, nil
	}
}
```

pkg/foundation/metrics/measure/measure.go
```diff
@@ -32,8 +32,8 @@ var (
 		"Number of connectors by type.",
 		[]string{"type"})
 	ProcessorsGauge = metrics.NewLabeledGauge("conduit_processors",
-		"Number of processors by type.",
```
lovromazgon, line 35: Well, it is by processor name, and since we will rename field `name` to `type` it will be by type at that point 😄 We have to remember to update this gauge when renaming `name` to `type`.

pkg/processor/procjs/jsprocessor.go
```diff
@@ -0,0 +1,184 @@
...
+	"github.com/dop251/goja"
+	"github.com/rs/zerolog"
+)
+
+const (
+	entrypoint = "process"
+)
+
+// JSProcessor is able to run processors defined in JavaScript.
+type JSProcessor struct {
```
lovromazgon, line 32: Nitpick: given that it's the only processor in this package I'd just name it `Processor` and the file `processor.go`.

pkg/processor/procjs/jsprocessor.go
```diff
@@ -0,0 +1,184 @@
...
+
+func (p *JSProcessor) jsContentRaw(goja.ConstructorCall) *goja.Object {
+	// TODO accept arguments
+	// We return a record.RawData struct, however because we are
+	// not changing call.This instanceof will not work as expected.
+
+	r := record.RawData{}
+	// We need to return a pointer to make the returned object mutable.
+	return p.runtime.ToValue(&r).ToObject(p.runtime)
+}
```
lovromazgon, line 111: TBH I'd prefer if these functions were separate from the processor struct. Previously they were static, but maybe it would be best to gather them on a struct like `Helper` or something, to make it clear they are here to be called from the JS script.

pkg/processor/instance.go
```diff
@@ -13,7 +13,6 @@
 // limitations under the License.
 
 //go:generate mockgen -destination=mock/processor.go -package=mock -mock_names=Processor=Processor . Processor
```
lovromazgon, line 15:
```suggestion
//go:generate mockgen -destination=mock/processor.go -package=mock -mock_names=Interface=Interface . Interface
```

lovromazgon (commented):
Very nice, can't wait to get this in!

pkg/processor/procjs/jsprocessor.go
```diff
@@ -0,0 +1,184 @@
...
+		r.Key = &v
+	case record.StructuredData:
+		r.Key = &v
+	}
+
+	// we need to send in a pointer to let the user change the value and return it, if they choose to do so
+	return p.runtime.ToValue(&r)
+}
+
+func (p *JSProcessor) toInternal(v goja.Value) (*record.Record, error) {
```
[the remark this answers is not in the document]
hariso: It was a bit long for me, but I can revert it.: )

pkg/processor/procjs/jsprocessor.go
hariso: I actually had all of the JS stuff in a separate package, a "JS engine" which had nothing to do with processing itself. However, I realized that in that case the JS processor isn't doing much, so it felt like a premature refactoring.

pkg/processor/instance.go
hariso: Yeah, it's an outdated comment from the version where I had a function field, so the old `funcProcessor` never stored any state itself, but was always just forwarding the calls to the fn function field. I'll update the comment. Thanks for pointing it out!

PR #487, opened 2022-06-24
Remove message status dropped

### LINKED ISSUE #394: Feature: Remove message status "dropped"

lovromazgon:
### Description

This PR simplifies message statuses by removing the status "dropped", messages can now get either acked or nacked. If a message can't be processed by a node it must get nacked, it can't just be dropped anymore. This ensures that `SourceAckerNode` can decide what to do if a message can't get processed and not the node itself (e.g. it can decide to store it in a DLQ). The states and expected behavior are fully explained [here](https://github.com/ConduitIO/conduit/blob/main/docs/pipeline_semantics.md#message-states).

Depends on #483.

Fixes #394

### Quick checks:

- [X] I have written unit tests (changed existing ones).


hariso (approved):
Let's drop the status "dropped".: )

PR #497, opened 2022-06-27
inject env vars in yaml

### LINKED ISSUE #490: Bind environment variables into pipeline config
Make sure that environment variables can be injected into the pipeline config with the syntax `${MY_ENV_VAR}`.

Depends on #489.

maha-hajja:
### Description

Fixes #490 

### Quick checks:


pkg/conduit/pipelines_parser_test.go
```diff
@@ -145,7 +145,62 @@ func Test_Parser_Invalid_Version(t *testing.T) {
...
+		t.Fatalf("Failed to write env var: $AWS_SECRET")
+	}
+	err = os.Setenv("AWS_KEY", "my-aws-key")
+	if err != nil {
+		t.Fatalf("Failed to write env var: $AWS_KEY")
+	}
+	err = os.Setenv("AWS_URL", "aws-url")
+	if err != nil {
+		t.Fatalf("Failed to write env var: $AWS_URL")
+	}
```
hariso, line 171: It might be good to clean up these. If they are used when normally developing something (e.g. when working on the S3 connector), they may interfere. Another option is to use some bogus or random variable names.

hariso (approved):

pkg/conduit/pipelines_parser_test.go
lovromazgon: These env vars will only be set for the lifetime of the test run (i.e. one invocation of `go test`). Granted, they will be visible in any test that is run after this test, but they shouldn't be visible after `go test` stops running, so not a lot of harm is done, I don't think other tests use these variables. That said, I agree that we could maybe use var names that we know won't interfere with anything like `CONDUIT_TEST_PARSER_X` or something.

PR #502, opened 2022-07-01
validate pipeline config

### LINKED ISSUE #491: Pipeline config validator
Create a function that takes a pipeline config and validates it for its correctness (e.g. mandatory fields, enums). Think about if we should validate for pipeline correctness (e.g. the plugin exists, processor exists, pipeline has at least a source and destination connector), I'm leaning towards _not_ checking these things as they will be reported by the service when we try to provision the pipeline anyway and we would be duplicating logic here.

Depends on #489.

maha-hajja:
### Description

Fixes #491 

### Quick checks:

### DIFF
--- a/pkg/provisioning/pipeline_config_validator.go
+++ b/pkg/provisioning/pipeline_config_validator.go
@@ -1,0 +1,101 @@
+// Copyright © 2022 Meroxa, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package provisioning
+
+import "github.com/conduitio/conduit/pkg/foundation/cerrors"
+
+var (
+	ErrMandatoryField = cerrors.New("mandatory field not specified")
+	ErrInvalidField   = cerrors.New("invalid field value")
+)
+
+// ValidatePipelinesConfig sets default values for empty fields and validates config field values for pipelines
+func ValidatePipelinesConfig(mp map[string]PipelineConfig) (map[string]PipelineConfig, error) {
+	var err error
+	mp = enrichPipelinesConfig(mp)
+	for k, cfg := range mp {
+		cfg.Connectors, err = validateConnectorsConfig(cfg.Connectors)
+		if err != nil {
+			return nil, cerrors.Errorf("pipeline `%s`: %w", k, err)
+		}
+		cfg.Processors, err = validateProcessorsConfig(cfg.Processors)
+		if err != nil {
+			return nil, cerrors.Errorf("pipeline `%s`: %w", k, err)
+		}
+		mp[k] = cfg
+	}
+	return mp, nil
+}
+
+// validateConnectorsConfig validates config field values for connectors
+func validateConnectorsConfig(mp map[string]ConnectorConfig) (map[string]ConnectorConfig, error) {
+	var err error
+	for k, cfg := range mp {
+		if cfg.Plugin == "" {
+			return nil, cerrors.Errorf("connector `%s`: `plugin` is mandatory: %w", k, ErrMandatoryField)
+		}
+		if cfg.Type == "" {
+			return nil, cerrors.Errorf("connector `%s`: `type` is mandatory: %w", k, ErrMandatoryField)
+		}
+		if cfg.Type != "source" && cfg.Type != "destination" {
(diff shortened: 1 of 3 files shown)


lovromazgon (changes requested):
Don't let the number of comments fool you, I like the PR! 😄 👍

pkg/provisioning/pipeline_config_validator.go
```diff
@@ -0,0 +1,101 @@
...
+
+import "github.com/conduitio/conduit/pkg/foundation/cerrors"
+
+var (
+	ErrMandatoryField = cerrors.New("mandatory field not specified")
+	ErrInvalidField   = cerrors.New("invalid field value")
+)
+
+// ValidatePipelinesConfig sets default values for empty fields and validates config field values for pipelines
+func ValidatePipelinesConfig(mp map[string]PipelineConfig) (map[string]PipelineConfig, error) {
```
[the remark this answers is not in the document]
maha-hajja: it is a separate function, I only called it here for now.. I wanted to discuss it further to see if we should expect the caller to always call the `enrich` before the `validate` function or not.
when u say splitting them.. do you mean having another file for the `enrich` and its own tests? or just not to call the `enrich` function in the `validate`?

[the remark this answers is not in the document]
lovromazgon: I'd do both - put enrich into a separate file and expect the caller to call it before validating the config.

pkg/provisioning/validate.go
```diff
@@ -0,0 +1,83 @@
...
+		err = multierror.Append(err, tmpErr)
+	}
+	return err
+}
+
+// validateConnectorsConfig validates config field values for connectors
+func validateConnectorsConfig(mp map[string]ConnectorConfig) error {
+	var err, pErr error
+	for k, cfg := range mp {
+		if cfg.Plugin == "" {
```
hariso, line 55: Do we want to check for blank strings too (i.e. only whitespace)?

pkg/provisioning/validate_test.go
```diff
@@ -0,0 +1,180 @@
...
+
+import (
+	"testing"
+
+	"github.com/conduitio/conduit/pkg/foundation/cerrors"
+	"github.com/conduitio/conduit/pkg/foundation/multierror"
+	"github.com/matryer/is"
+)
+
+func TestValidator_ConnectorMandatoryField1(t *testing.T) {
```
hariso, line 25: Nitpick and maybe even a preference thing: it would be good that whatever is printed in the logs, when a test fails, is more or less enough to know what test case failed. We can achieve that by using an appropriate test method name or IIUC by setting a test name. Now, `TestValidator_ConnectorMandatoryField1` is already pretty clear, and maybe we can leave it like that. What I'm suggest is having something like `TestValidator_ConnectorMandatoryField_TypeName`.

pkg/provisioning/validate.go
maha-hajja: I tried that before by hand.. an empty field in the yaml file would be parsed as "", this is also the case when the field value is a couple of spaces like "    ", they both equal "" after parsing.

hariso: Awesome, it's great that you checked that already!

pkg/provisioning/validate_test.go
lovromazgon: This would be a great opportunity to create [table driven test](https://github.com/golang/go/wiki/TableDrivenTests) 🙂 Something like `TestValidator_MandatoryFields` and a test case for each mandatory field.

pkg/provisioning/validate_test.go
```diff
@@ -0,0 +1,180 @@
...
+	for _, gotErr := range multierr.Errors() {
+		if cerrors.Is(gotErr, ErrInvalidField) {
+			invalid++
+		}
+		if cerrors.Is(gotErr, ErrMandatoryField) {
+			mandatory++
+		}
+	}
+	is.Equal(invalid, 1)
+	is.Equal(mandatory, 3)
```
lovromazgon, line 179: Would be great if we also asserted that `multierr.Errors()` contains 4 errors in total.

pkg/provisioning/validate.go
```diff
@@ -0,0 +1,83 @@
...
+		mp[k] = cfg
+	}
+	return err
+}
+
+// validateProcessorsConfig validates config field values for processors
+func validateProcessorsConfig(mp map[string]ProcessorConfig) error {
+	for k, cfg := range mp {
+		if cfg.Type == "" {
+			return cerrors.Errorf("processors %q: \"type\" is mandatory: %w", k, ErrMandatoryField)
```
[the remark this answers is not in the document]
maha-hajja: it's only one error in the method.. that will be wrapped in other methods, so didn't think it was needed.
It would be necessary though if we add other validations to the processor, so I'll change it to use multierror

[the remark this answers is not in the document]
lovromazgon: We are looping through a map so there might be multiple processors that lack a `Type`, that's what I was thinking of.

lovromazgon (approved):

PR #515, opened 2022-07-07
Distinguish entities provisioned through config file

### LINKED ISSUE #492: Distinguish entities provisioned through config file
Add a field to entities (pipeline, processor, connector) that can be used to distinguish if a pipeline was provisioned through a config file. I propose adding a field like `provisionedBy` (name is debatable) that would contain `config` or `API` depending on what triggered the creation.

In scope of this issue we should add a check in the orchestrators that prevents the mutation of entities provisioned by a `config` from API calls.

maha-hajja:
### Description

added ProvisionedBy enum to all entites

Fixes #492 

### Quick checks:


pkg/connector/mock/builder.go
```diff
@@ -44,6 +44,8 @@ func (b Builder) Init(c connector.Connector, id string, cfg connector.Config) er
 		m.EXPECT().Type().Return(connector.TypeSource).AnyTimes()
 		m.EXPECT().ID().Return(id).AnyTimes()
 		m.EXPECT().Config().Return(cfg).AnyTimes()
+		m.EXPECT().SetProvisionedBy(connector.ProvisionTypeAPI).AnyTimes()
```
lovromazgon, line 47: We shouldn't expect this method to be called, this can and should be set up in the tests that need it (same for destination). This comment may be redundant if we remove the method entirely.

pkg/connector/connector.go
```diff
@@ -43,6 +50,9 @@ type Connector interface {
 	Config() Config
 	SetConfig(Config)
 
+	ProvisionedBy() ProvisionType
+	SetProvisionedBy(ProvisionType)
```
lovromazgon, line 54: I wonder if we should remove this method, it implies that `ProvisionType` is a mutable field. I feel like this field shouldn't be able to change.

pkg/connector/service.go
```diff
@@ -102,6 +102,7 @@ func (s *Service) Create(ctx context.Context, id string, t Type, cfg Config) (Co
 		return nil, cerrors.Errorf("could not init connector: %w", err)
 	}
 
+	conn.SetProvisionedBy(p)
```
lovromazgon, line 105: If we remove the method we would need to pass `ProvisionType` to the builder when the connector is created in `Builder.Build`.

pkg/connector/connector.go
```diff
@@ -29,11 +29,18 @@ const (
...
+const (
+	ProvisionTypeAPI ProvisionType = iota
+	ProvisionTypeConfig
+)
+
 type (
 	// Type defines the connector type.
 	Type int
 	// Status defines the running status of a connector.
 	Status int
```
lovromazgon, line 41: I realized this type is redundant, can you please remove it?

pkg/orchestrator/connectors_test.go
```diff
@@ -106,6 +107,28 @@ func TestConnectorOrchestrator_Create_PipelineRunning(t *testing.T) {
...
+
+	plsMock.EXPECT().
+		Get(gomock.AssignableToTypeOf(ctxType), pl.ID).
+		Return(pl, nil)
+
+	orc := NewOrchestrator(db, log.Nop(), plsMock, consMock, procsMock, pluginMock)
+	got, err := orc.Connectors.Create(ctx, connector.TypeSource, connector.Config{PipelineID: pl.ID})
+	assert.Error(t, err)
+	assert.True(t, cerrors.Is(err, ErrImmutableProvisionedByConfig), "expected ErrImmutableProvisionedByConfig")
+	assert.Nil(t, got)
```
lovromazgon, line 129: Please use `github.com/matryer/is` in new tests 🙏

pkg/orchestrator/errors.go
```diff
@@ -21,4 +21,5 @@ var (
 	ErrPipelineHasProcessorsAttached  = cerrors.New("pipeline has processors attached")
 	ErrPipelineHasConnectorsAttached  = cerrors.New("pipeline has connectors attached")
 	ErrConnectorHasProcessorsAttached = cerrors.New("connector has processors attached")
+	ErrImmutableProvisionedByConfig   = cerrors.New("entity was provisioned by a config file and cannot be mutated")
```
lovromazgon, line 24:
```suggestion
	ErrImmutableProvisionedByConfig   = cerrors.New("entity was provisioned by a config file and cannot be mutated through the API, please change the corresponding config file instead")
```

pkg/orchestrator/connectors.go
```diff
@@ -46,11 +46,15 @@ func (c *ConnectorOrchestrator) Create(
 		return nil, cerrors.Errorf("couldn't get pipeline: %w", err)
 	}
 
+	if pl.ProvisionedBy == pipeline.ProvisionTypeConfig {
```
lovromazgon, line 49: Nitpick: if we assume that the orchestrator exists just for API requests (which I'm fine with at this point), I think it would be better to turn the condition around and check if the entity was provisioned by the API. In other words, I think it would be better if the orchestrator wouldn't care that there is such a thing as "config provisioned entities", it should just care that it can only mutate "pipeline provisioned entities".
```suggestion
	if pl.ProvisionedBy != pipeline.ProvisionTypeAPI {
```

pkg/connector/connector.go
```diff
@@ -29,11 +29,18 @@ const (
...
+	// ProvisionType defines provisioning type
+	ProvisionType int
```
lovromazgon, line 43: Thinking out loud here - how would it look like if this type was a string? Then we wouldn't need to have constants here (and in every other service package), the service itself wouldn't care what different provision types there are. Only the calling code would need to know what its provision type is (e.g. orchestrators would have the provision type `"api"`) and that it can only mutate entities with the same provision type.

This might remove some boilerplate code, it would also make the provision type readable in logs since it would be a string.

pkg/orchestrator/errors.go
lovromazgon, line 24: We should map the new error to a gRPC status code in the handlers.

lovromazgon (commented):
Love the tests, this does what we want, I wonder if making `ProvisionType` a `string` would make it nicer though.

Also, I feel like the naming of `ProvisionType` could be improved, I thought about `Provisioner` or `Origin` but I'm not too happy with my suggestions either 😅 I'll think about it some more.

pkg/connector/mock/builder.go
```diff
@@ -44,8 +44,7 @@ func (b Builder) Init(c connector.Connector, id string, cfg connector.Config) er
 		m.EXPECT().Type().Return(connector.TypeSource).AnyTimes()
 		m.EXPECT().ID().Return(id).AnyTimes()
 		m.EXPECT().Config().Return(cfg).AnyTimes()
-		m.EXPECT().SetProvisionedBy(connector.ProvisionTypeAPI).AnyTimes()
-		m.EXPECT().ProvisionedBy().Return(connector.ProvisionTypeAPI).AnyTimes()
+		m.EXPECT().ProvisionedBy().AnyTimes()
```
lovromazgon, line 47: I'm wondering why this was changed? I assume calling `ProvisionedBy` would error out right now because the mock won't know what to return.
```suggestion
		m.EXPECT().ProvisionedBy().Return(connector.ProvisionTypeAPI).AnyTimes()
```

lovromazgon (approved):
LGTM 🚀

PR #518, opened 2022-07-08
Update dependabot to check npm weekly

jmar910:
### Description
Updates dependabot's interval from `daily` to `weekly` to cut down on some of the noise. Since we are in a good spot currently with a lot of the latest dep versions (except ember 4 and related deps), this is a good time to update the interval. 

### Quick checks:

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -22,6 +22,6 @@ updates:
   - package-ecosystem: "npm"
     directory: "/ui"
     schedule:
-      interval: "daily"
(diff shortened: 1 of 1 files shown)


lovromazgon (commented):
I feel like we could lower the frequency to bi-weekly or even monthly, but it's your call 👍 LGTM

jmar910:
I teetered between monthly and weekly (no interval in between unfortunately) but your comment convinced me 😄

PR #523, opened 2022-07-12
[UI] Fix supporting dots in key names

jmar910:
Fixes our normalization/serialization to handle plugin key names with periods. Fixes s3 connector config from UI. 

### Description
Most of our built-in plugins use camel casing for the key names and they work just fine. But the s3 built in connector uses periods for the key names, which confuses changesets under the hood when a users goes to create an s3 connector, resulting in an exception. 

In the interest of supporting different variations of key names with casing separators, we normalize/serialize key names at the boundary using `@@` and consistently use this in other places where this might occur (plugins api, processors, etc)

### Quick checks:

### DIFF
--- a/ui/app/serializers/connector.js
+++ b/ui/app/serializers/connector.js
@@ -7,9 +7,9 @@ const CONNECTOR_TYPE_MAP = {
 
 export default class ConnectorSerializer extends ApplicationSerializer {
   serialize(snapshot) {
-    const configSettings = this._replaceKeys(
+    const configSettings = super._replaceKeys(
       snapshot.record.config.settings,
-      '_',
+      '@@',
       '.'
     );
 
@@ -38,7 +38,11 @@ export default class ConnectorSerializer extends ApplicationSerializer {
 
   normalize(typeClass, hash) {
     if (hash.config?.settings) {
-      const configSettings = this._replaceKeys(hash.config?.settings, '.', '_');
+      const configSettings = super._replaceKeys(
+        hash.config?.settings,
+        '.',
+        '@@'
+      );
       hash.config.settings = configSettings;
     }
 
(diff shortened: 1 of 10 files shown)


ui/app/serializers/connector.js
```diff
@@ -7,9 +7,9 @@ const CONNECTOR_TYPE_MAP = {
 
 export default class ConnectorSerializer extends ApplicationSerializer {
   serialize(snapshot) {
-    const configSettings = this._replaceKeys(
+    const configSettings = super._replaceKeys(
       snapshot.record.config.settings,
```
jayjayjpg, line 11: I noticed that `config.settings` is guard checked against in the `normalize` method via `hash.config?.settings` -- would it make sense to do the same here as well?

jayjayjpg (approved):
This reads great and thank you for the swift iteration on this!

Not a blocker, and feel free to take a look in a follow-up PR if you find it valuable: Would it makes sense to add a unit or acceptance test for the normalization and serialization update?

This way, we can not only guard against regressions on this behaviour in the future, but we would also have more internal documentation on why we need to replace the config key separators with `@@` on the client-side

Other than that, this also already good to go and feel free to already merge to get the ball rolling on the s3 -> snowflake demo! :sparkles: :rocket:

ui/app/serializers/connector.js
jmar910: On serialization of a connector, the config object will always be present due to the connector model and how we create/update connectors so its not totally necessary, but I don't think it could hurt

jayjayjpg: I see, thank you for the outline and I agree with you that we could then just go either way here -- I think it's best to be your call!

PR #530, opened 2022-07-13
Revert "When only docs change, don't run build and lint workflows"

hariso:
Reverts ConduitIO/conduit#527.

Even though build and lint can be ignored for docs-only PRs, GitHub still expects them to run if they are required checks. I didn't notice this in my fork, because I didn't have the same protection rules after forking.

Also related: https://github.com/actions/virtual-environments/issues/1281

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -3,11 +3,7 @@ name: build
 on:
   push:
     branches: [ main ]
-    paths-ignore:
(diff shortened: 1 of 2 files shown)


hariso:
> The devil is once again in the details smile

Definitely.:D It's also funny that, I actually had a fork with the protection rules, but I deleted it one or two days ago.: )

PR #533, opened 2022-07-19
add example systemd unit

neovintage:
This PR adds an example script that a developer can use to add Conduit to systemd.


scripts/conduit.service
```diff
@@ -0,0 +1,9 @@
+[Unit]
+Description=Conduit daemon
+
+[Service]
+Type=forking
```
lyuboxa, line 5: might be good to give it a workdir and run it under a specific user?

neovintage: @lyuboxa why the working dir? If conduit is set up in `/usr/bin`, does that matter?

lyuboxa: it needs to create the database file somewhere, it better be somewhere known, like /var/lib/conduit

jkayh (approved):

PR #558, opened 2022-08-04
OpenCDC Integration

### LINKED ISSUE #468: JavaScript transforms: Provide easy way to work with structured data
### Feature description

Currently, it's not possible for JS transforms to return records with changed structured data. This issue has two parts:
1. There's no way in a JS script to create a new structured payload. The helper method Conduit currently provides make it possible to create raw payloads.
2. There's no easy way to modify a structured payload (e.g. replacing a key, adding a new one etc.). The struct used to represent structured data, `record.StructuredData` has the underlying type `map[string]interface{}`, however, due to the explanation given[ here](https://github.com/dop251/goja/issues/399), it's not possible to use it as a map.

### LINKED ISSUE #511: Integrate OpenCDC into Conduit
Change Conduit to use the OpenCDC record instead of its own record.

Depends on #510 and in part on #512 (builtin connectors need to be updated first).

lovromazgon:
### Description

This PR updates Conduit to use a new Record structure that matches OpenCDC records. Apart from that it:
- Updates versions of all builtin plugins and uses `sdk.Connector` to register them in the registry.
- Changes processors `insertfieldkey` and `insertfieldpayload` by removing the config option `timestamp.field` (see [comment](#discussion_r938750791)).
- Moves field `Record.SourceID` to `Message.SourceID`.
- Creates an intermediary record that is passed to the JS processor instead of our internal record. This allows the JS processor to manipulate all fields including structured data.

Fixes #511, fixes #468.

### Quick checks:


pkg/processor/procbuiltin/insertfield.go
```diff
@@ -26,10 +26,9 @@ const (
 	insertFieldKeyName     = "insertfieldkey"
 	insertFieldPayloadName = "insertfieldpayload"
 
-	insertFieldConfigStaticField    = "static.field"
-	insertFieldConfigStaticValue    = "static.value"
-	insertFieldConfigTimestampField = "timestamp.field"
```
lovromazgon, line 31: This setting was removed completely because there is no `Record.CreatedAt` field anymore. It might make more sense to add a setting like `metadata.field` to fetch a field from metadata.

pkg/processor/procjs/processor.go
```diff
@@ -28,6 +28,20 @@ const (
 	entrypoint = "process"
 )
 
+// jsRecord is an intermediary representation of record.Record that is passed to
+// the JavaScript transform. We use this because using record.Record would not
+// allow us to modify or access certain data (e.g. metadata or structured data).
+type jsRecord struct {
```
lovromazgon, line 34: This part needs special attention. An intermediary record struct was introduced to fix handling of `Record.Metadata` in the JS processor, that field had essentially the same problem as structured data (see https://github.com/ConduitIO/conduit/issues/468). That issue should also be fixed now, since structured data is transformed into a simple `map[string]interface{}` when it's passed to the JS function.

pkg/record/record.go
```diff
@@ -12,34 +12,87 @@
...
+	OperationDelete
+	OperationSnapshot
+)
+
+// Operation defines what triggered the creation of a record.
+type Operation int
+
+// Record represents a single data record produced by a source and/or consumed
+// by a destination connector.
 type Record struct {
```
lovromazgon, line 39: We now have essentially the same code for the record in the SDK and in Conduit (see [this](https://github.com/ConduitIO/conduit-connector-sdk/blob/main/record.go#L52)). According to @ahmeroxa they want to add a similar / the same struct to Turbine as well. If we all agree to use the OpenCDC record it would make sense to extract this into a repository that contains the OpenCDC protobuf definition, the Go struct and utilities to transform it into the protobuf definition, then import it as a library in the SDK and Conduit.

pkg/pipeline/stream/message.go
```diff
@@ -54,6 +54,9 @@ type Message struct {
 	// Record represents a single record attached to the message.
 	Record record.Record
 
+	// SourceID contains the source connector ID.
+	SourceID string
```
lovromazgon, line 58: Source ID was moved from the record into the message, since it's an internal Conduit field, not something the record should be concerned with. This field is set in `SourceNode` when it creates the message.

pkg/orchestrator/orchestrator_test.go
```diff
@@ -125,7 +148,7 @@ func TestPipelineSimple(t *testing.T) {
 	t.Log("successfully stopped pipeline")
 
 	// make sure destination file matches source file
-	want, err := os.ReadFile(sourcePath)
+	want, err := os.ReadFile(wantPath)
```
lovromazgon, line 151: The file destination now produces the whole record instead of only the payload so the test expectation had to be adjusted.

lovromazgon (commented):
It's a huge PR, here's a couple of comments describing parts that change more than just the OpenCDC record. Everything else should be just simple changes to follow the new record structure (e.g. instead of using `Record.Payload` the code was changed to use `Record.Payload.After`).

pkg/pipeline/stream/stream_test.go
```diff
@@ -323,16 +324,12 @@ func generatorSource(ctrl *gomock.Controller, logger log.CtxLogger, nodeID strin
 
 		position++
 		return record.Record{
-			// SourceID would normally be the source node ID, but since we need
-			// to add the node ID to the position to create unique positions we
```
lovromazgon, line 327: We don't have to create unique positions anymore, the restriction doesn't exist anymore, so this part was simplified.

pkg/processor/procbuiltin/extractfield_test.go
```diff
@@ -206,13 +206,19 @@ func TestExtractFieldPayload_Process(t *testing.T) {
 			Settings: map[string]string{extractFieldConfigField: "foo"},
 		},
 		args: args{r: record.Record{
-			Payload: record.StructuredData{
-				"foo": 123,
+			Payload: record.Change{
+				Before: nil,
+				After: record.StructuredData{
+					"foo": 123,
+				},
```
hariso, line 213: IIRC, the decision for this and similar built-in processors was to operate on the `after` field. If so, it would be good to mention that in the docs.

Also, let's say we use extractfieldpayload. Should the output be a Change, where `before` remains the same, and `after` is transformed, or should be just some raw data with the extracted field?

pkg/pipeline/stream/processor_test.go
```diff
@@ -35,9 +34,8 @@ func TestProcessorNode_Success(t *testing.T) {
 	ctrl := gomock.NewController(t)
 
 	wantRec := record.Record{
-		Position:  []byte(uuid.NewString()),
-		Metadata:  map[string]string{"foo": "bar"},
-		CreatedAt: time.Now().UTC(),
```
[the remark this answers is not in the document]
lovromazgon: Do you mean the field `CreatedAt`? It's been moved to metadata and is an optional field, if the connector sets the field it's there, otherwise it's not.

pkg/processor/procbuiltin/extractfield_test.go
lovromazgon: Exactly, all built-in transforms were changed to use `Payload.After` instead of just `Payload`. I'll add it to the documentation, I just have to find the right place for this info 🙂

I'm not sure I understand the question about `extractfieldpayload` - the output is the same as before, only that instead of manipulating `Payload` it's manipulating `Payload.After`.

pkg/connector/source.go
```diff
@@ -227,14 +227,16 @@ func (s *source) Read(ctx context.Context) (record.Record, error) {
 		return r, err
 	}
 
+	// TODO rethink if there's an actual benefit in setting these fields
```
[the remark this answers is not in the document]
lovromazgon: It would be cleaner and "more correct" to leave the field `nil`, on the other hand it would mean we have to check for `nil` every time we access a key or payload field, even just calling `Bytes()` on it would panic.

Another option would be to introduce a type `NilData` that just does nothing, then it wouldn't look like it's empty raw data.

pkg/processor/procbuiltin/extractfield_test.go
lovromazgon: I updated the processor comments and the processor doc.

hariso: > I'm not sure I understand the question about `extractfieldpayload` - the output is the same as before, only that instead of manipulating `Payload` it's manipulating `Payload.After`.

Nevermind, my question is actually not valid.:) Prior to this change, record.Payload was being replaced with the field.

So I was wondering if we are supposed to do that here too (and not replace record.Payload.After with the extracted field). But that actually makes no sense, since record.Payload is a `Change`, not `Data`.

pkg/pipeline/stream/processor_test.go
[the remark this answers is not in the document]
hariso: Yup, CreatedAt was my question.

pkg/processor/procjs/processor.go
```diff
@@ -110,75 +125,98 @@ func (p *Processor) jsContentRaw(goja.ConstructorCall) *goja.Object {
...
+	// We return a map[string]interface{} struct, however because we are
+	// not changing call.This instanceof will not work as expected.
+
+	r := make(map[string]interface{})
+	return p.runtime.ToValue(r).ToObject(p.runtime)
+}
+
 func (p *Processor) Process(_ context.Context, in record.Record) (record.Record, error) {
-	jsRecord := p.toJSRecord(in)
+	jsr := p.toJSRecord(in)
```
hariso, line 138: Java Specification Request... Sorry, flashbacks.:D

pkg/record/record.go
hariso: I like the idea. :+1:

pkg/processor/procjs/processor.go
```diff
@@ -110,75 +125,98 @@ func (p *Processor) jsContentRaw(goja.ConstructorCall) *goja.Object {
...
 }
 
 func (p *Processor) toJSRecord(r record.Record) goja.Value {
-	// convert content to pointers to make it mutable
-	switch v := r.Payload.(type) {
-	case record.RawData:
-		r.Payload = &v
-	case record.StructuredData:
-		r.Payload = &v
+	extractData := func(d record.Data) interface{} {
```
hariso, line 157: Nitpick: Maybe we can name it `convertData`, since it looks like the data itself remains unchanged, we're "casting" it merely.

pkg/processor/procjs/processor.go
```diff
@@ -110,75 +125,98 @@ func (p *Processor) jsContentRaw(goja.ConstructorCall) *goja.Object {
...
-		r.Key = &v
+	jsr := jsRecord{
+		Position:  r.Position,
+		Operation: r.Operation.String(),
+		Metadata:  r.Metadata,
+		Key:       extractData(r.Key),
+		Payload: struct {
+			Before interface{}
+			After  interface{}
+		}{
```
hariso, line 175: Do anonymous structs in Go create any overhead?

pkg/processor/procjs/processor.go
```diff
@@ -110,75 +125,98 @@ func (p *Processor) jsContentRaw(goja.ConstructorCall) *goja.Object {
...
 
-	switch v := r.Key.(type) {
-	case *record.RawData:
-		r.Key = *v
-	case *record.StructuredData:
-		r.Key = *v
+	var op record.Operation
+	err := op.UnmarshalText([]byte(jsr.Operation))
+	if err != nil {
+		return record.Record{}, cerrors.Errorf("could not unmarshal operation: %q", err)
```
hariso, line 199: It looks like we're missing a `%w` in the format string.

pkg/record/record.go
```diff
@@ -12,34 +12,121 @@
...
+	Key Data `json:"key"`
+	// Payload holds the payload change (data before and after the operation
+	// occurred).
+	Payload Change `json:"payload"`
+}
+
+// Bytes returns the JSON encoding of the Record.
+func (r Record) Bytes() []byte {
+	if r.Metadata == nil {
+		// since we are dealing with a Record value this will not be seen
```
hariso, line 97: If `Bytes()` is called often, that means, we might be increasing the memory usage?

hariso (approved):
Great work, Lovro!

pkg/record/record.go
lovromazgon: You mean because we are instantiating a map? Well sure, but only temporarily. The first question is if the map even escapes to the heap - if it stays on the stack then the memory will be reclaimed when the function exits. If on the other hand it lands on the heap it will be reclaimed next time the GC runs, because no object will reference the map anymore (the function receiver is a value so the `Metadata` field change is not visible outside of this function).

pkg/processor/procjs/processor.go
lovromazgon: Not that I know of, we'd have to measure it. Although I honestly doubt that it would make a difference in the JS processor which is already relatively slow itself.

pkg/record/record.go
hariso: I was asking because of the record as a whole, and mostly because of the payload. I commented on your comment above, since it mentions dealing with Record values.

lovromazgon: Maybe I should clarify - when I wrote "value" I meant a "Go value", as in "not a pointer". I'm saying that the receiver is a `Record` and not `*Record`.

I'm not sure I understand your concerns 🤔 I _assume_ you mean that because we use a value receiver the record will be copied when `Bytes` is called, which will increase the memory usage? I need to point out the copy is shallow, so if the record is carrying a huge payload it won't copy the payload - that's also the reason why we have to [delete the key](https://github.com/ConduitIO/conduit/blob/f4ce2e1007d55a4a8e99798ffcb3ec1a6979661a/pkg/record/record.go#L107) from metadata before returning from this function, because the metadata map is not copied.

Btw, maybe it's counterintuitive, but using values can be faster than using pointers, even though they are "copied" because Go does some optimizations in the background. I think this [blog post](https://www.ardanlabs.com/blog/2017/05/language-mechanics-on-stacks-and-pointers.html) explains it in depth.

PR #567, opened 2022-08-10
provision config file pipelines

### LINKED ISSUE #493: Provision pipeline from pipeline config
We need to implement logic that takes a pipeline config file and provisions the pipeline defined in it. If a pipeline with that name already exists it needs to be updated to match whatever the config defines, otherwise it needs to be created from scratch.

Depends on #489.

maha-hajja:
### Description

*creates a service that provisions a config file with all its pipelines
 *the service would check if a pipeline already exists, then copy the old pipeline connectors' states to the new pipeline.

todos:
* more tests

Fixes #493 

### Quick checks:


pkg/provisioning/provision.go
```diff
@@ -0,0 +1,240 @@
...
+	return &Service{
+		ctx:              ctx,
+		logger:           logger.WithComponent("provision"),
+		pipelineService:  plService,
+		connectorService: connService,
+		processorService: procService,
+	}
+}
+
+func (s *Service) ProvisionConfigFile(path string) error {
```
lovromazgon, line 53: So there will be a config file directory that we will scan for yaml files and provision all pipelines in that directory. That directory will be determined when Conduit starts - default is `./pipelines` but can be overridden using a CLI flag.

I'm thinking that it would make sense to pass this directory path to `Service` when creating it in `NewService` and store it in the service. This will enable us to write a hot reload function in the future if we need it, since we will already know the directory path.

I would then expect to see an `Init` function here that scans the folder and provisions the pipelines, this way we would follow the pattern we have in other services (see [this](https://github.com/ConduitIO/conduit/blob/30a574c86c10d9280517a0ea086339dd8bbd5662/pkg/conduit/runtime.go#L207-L218)).

pkg/provisioning/provision.go
```diff
@@ -0,0 +1,240 @@
...
+	"context"
+	"io/ioutil"
+
+	"github.com/conduitio/conduit/pkg/connector"
+	"github.com/conduitio/conduit/pkg/foundation/log"
+	"github.com/conduitio/conduit/pkg/pipeline"
+	"github.com/conduitio/conduit/pkg/processor"
+)
+
+var update = false
```
lovromazgon, line 27: Oh no, this can surely be changed to a local variable instead of a global one.

pkg/provisioning/provision.go
```diff
@@ -0,0 +1,240 @@
...
+	// set pipeline status
+	status := pipeline.StatusUserStopped
+	if config.Status == StatusRunning {
+		err := s.pipelineService.Start(*s.ctx, s.connectorService, s.processorService, newPl)
+		if err != nil {
+			return err
+		}
+		status = pipeline.StatusRunning
+	}
+	newPl.Status = status
```
lovromazgon, line 140: We shouldn't modify this field, it is managed by the pipeline itself. It will be changed to running if the pipeline start is successful.

pkg/provisioning/provision.go
```diff
@@ -0,0 +1,240 @@
...
+	}
+
+	if update {
+		switch v := conn.(type) {
+		case connector.Destination:
+			v.SetState(destMap[id])
+		case connector.Source:
+			v.SetState(srcMap[id])
+		}
+	}
```
lovromazgon, line 204: We need to check first if we can even copy over the state, if the plugin changed we shouldn't copy the state and display a warning instead.

Think about extracting the update logic into a separate function.

pkg/provisioning/provision.go
```diff
@@ -0,0 +1,240 @@
...
+
+	return nil
+}
+
+func (s *Service) provisionPipeline(id string, config PipelineConfig) error {
+	var destMap map[string]connector.DestinationState
+	var srcMap map[string]connector.SourceState
+	var err error
+	oldPl, err := s.pipelineService.Get(*s.ctx, id)
```
lovromazgon, line 85: If it returns no error there is a pipeline, if it returns `pipeline.ErrInstanceNotFound` there is no pipeline, if there is some other error we have a bigger problem and should probably stop and exit.

pkg/provisioning/provision.go
```diff
@@ -0,0 +1,240 @@
...
+		err = s.provisionPipeline(k, v)
+		if err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
+
+func (s *Service) provisionPipeline(id string, config PipelineConfig) error {
```
lovromazgon, line 81: It would be great if this operation would happen inside a transaction and rollback any changes if it didn't succeed, otherwise we risk ending up with a partially provisioned pipeline. For the transaction `Service` will need access to the db, for rollbacks there is the utility `rollback.R`. See the [orchestrator](https://github.com/ConduitIO/conduit/blob/30a574c86c10d9280517a0ea086339dd8bbd5662/pkg/orchestrator/processors.go#L35-L42) how we use it there.

pkg/provisioning/provision.go
```diff
@@ -0,0 +1,240 @@
...
+	for k := range config.Connectors {
+		destState, srcState, err := s.getConnectorState(k)
+		if err != nil {
+			return nil, nil, err
+		}
+		destMap[k] = destState
+		srcMap[k] = srcState
+	}
+
+	err := s.pipelineService.Delete(*s.ctx, pl)
```
lovromazgon, line 158: This won't be so easy, we need to first remove all processors (on the pipeline and connectors) and all connectors.

pkg/provisioning/provision.go
lovromazgon: Oh ok, if you wanted to split it, that's fine!

pkg/provisioning/provision.go
```diff
@@ -0,0 +1,240 @@
...
+
+	before, err := Parse(data)
+	if err != nil {
+		return err
+	}
+
+	got := EnrichPipelinesConfig(before)
+
+	for k, v := range got {
+		// skip non valid? multierror?
```
hariso, line 67: I wouldn't skip, because in that case, a user would _have_ to have a way to monitor the pipelines. Otherwise, a failing pipeline will never be detected. Also, an invalid config is quite a thing, and I think we should be loud about it.

pkg/provisioning/provision.go
```diff
@@ -0,0 +1,240 @@
...
+func (s *Service) provisionPipeline(id string, config PipelineConfig) error {
+	var destMap map[string]connector.DestinationState
+	var srcMap map[string]connector.SourceState
+	var err error
+	oldPl, err := s.pipelineService.Get(*s.ctx, id)
+	if err == nil {
+		update = true
+		destMap, srcMap, err = s.copyStatesAndDeletePipeline(oldPl, config)
+		if err != nil {
+			return err
```
hariso, line 90: It would be good to add more info about this error (and all the other returned errors), e.g. the pipeline/connector/processor name.

pkg/provisioning/provision.go
```diff
@@ -0,0 +1,240 @@
...
+	if config.Type == TypeDestination {
+		connType = connector.TypeDestination
+	}
+
+	conn, err := s.connectorService.Create(*s.ctx, id, connType, cfg, connector.ProvisionTypeConfig)
+	if err != nil {
+		return err
+	}
+
+	if update {
```
hariso, line 197: Maybe you don't need the update flag at all. If there's something in the map, we copy it, otherwise, we leave it as it is (empty state).

pkg/provisioning/provision.go
maha-hajja: yep! already on my todo list

pkg/provisioning/provision.go
maha-hajja: yeah I was thinking about provisioning all the valid pipelines... and have a multierror that has all the failing ones, which will be returned at the end and logged

pkg/provisioning/provision.go
hariso: @maha-hajja If you plan to do as part of another PR, I can resolve this conversation.

pkg/provisioning/interfaces.go
```diff
@@ -0,0 +1,51 @@
...
+	Delete(ctx context.Context, id string) error
+
+	AddProcessor(ctx context.Context, connectorID string, processorID string) (connector.Connector, error)
+	RemoveProcessor(ctx context.Context, connectorID string, processorID string) (connector.Connector, error)
+}
+
+type ProcessorService interface {
+	Get(ctx context.Context, id string) (*processor.Instance, error)
+	Create(ctx context.Context, id string, name string, parent processor.Parent, cfg processor.Config, p processor.ProvisionType) (*processor.Instance, error)
+}
```
hariso, line 51: I'm joining the review late, and you may have already discussed this, but why do we need these interfaces, if we have the orchestrator interfaces?

pkg/provisioning/provision.go
```diff
@@ -0,0 +1,345 @@
...
+		if err != nil {
+			// should we return, or add one to the multierror
+			return err
+		}
+	}
+
+	return multi
+}
+
+func (s *Service) provisionPipeline(ctx context.Context, id string, config PipelineConfig) error {
```
hariso, line 85: It would be nice to see this function shorter/refactored, so we can quickly see what's happening at a high level.

pkg/provisioning/provision.go
maha-hajja: it's done

pkg/provisioning/provision.go
```diff
@@ -0,0 +1,353 @@
...
+
+	return nil
+}
+
+func (s *Service) getConnectorState(ctx context.Context, id string, plugin string) (connector.DestinationState, connector.SourceState, error) {
+	var destState connector.DestinationState
+	var srcState connector.SourceState
+	conn, err := s.connectorService.Get(ctx, id)
+	if err != nil {
+		return destState, srcState, cerrors.Errorf("could not get connector %q: connector id cannot be changed", id)
```
lovromazgon, line 290: This doesn't seem right - if the connector doesn't exist the ID was probably changed, but that's a valid thing to do. In that case the user essentially added a new connector to the config and deleted the old one.

pkg/provisioning/provision_test.go
```diff
@@ -0,0 +1,511 @@
...
+			},
+		},
+	},
+}
+
+func TestProvision_PipelineWithConnectorsAndProcessors(t *testing.T) {
+	is := is.New(t)
+	logger := log.Nop()
+	db := &inmemory.DB{}
+	_, ctx, _ := db.NewTransaction(context.Background(), true)
```
lovromazgon, line 125: It took me some time to figure out why we have this line - I assume we do this to get a context that will fit the expectation in the mock. It's a bit hacky and will break if we wrap the context and add something else to it other than the transaction, I'd rather just check that the context is not nil (e.g. `myMock.EXPECT().Call( gomock.Not(gomock.Nil()), ... )`). The best solution would be to have a matcher that makes sure the context contains a transaction, but that's a bit harder.

pkg/provisioning/provision_test.go
```diff
@@ -0,0 +1,511 @@
...
+	// pipeline already exists
+	pipelineService.EXPECT().Get(ctx, "pipeline2").Return(pl2, nil)
+
+	// the old source connector has a different Plugin
+	connService.EXPECT().Get(ctx, "con1").Return(modifiedSource, nil)
+	connService.EXPECT().Get(ctx, "con2").AnyTimes()
+
+	service := NewService(db, logger, pipelineService, connService, procService)
+	err := service.ProvisionConfigFile(context.Background(), "./test/provision2-simple-pipeline.yml")
+	is.True(err != nil)
```
lovromazgon, line 394: As mentioned in the code - this should return no error, the pipeline needs to be provisioned but a warning should be logged.

pkg/provisioning/provision.go
```diff
@@ -0,0 +1,353 @@
...
+	"github.com/conduitio/conduit/pkg/foundation/cerrors"
+	"github.com/conduitio/conduit/pkg/foundation/database"
+	"github.com/conduitio/conduit/pkg/foundation/log"
+	"github.com/conduitio/conduit/pkg/foundation/multierror"
+	"github.com/conduitio/conduit/pkg/foundation/rollback"
+	"github.com/conduitio/conduit/pkg/pipeline"
+	"github.com/conduitio/conduit/pkg/processor"
+)
+
+type Service struct {
```
lovromazgon, line 31: Other packages call the file where the service is implemented `service.go`, we should probably do the same here. While we're renaming things - `pipelines_parser.go` should IMO be renamed to `parse.go`, that's the main function in that file.

pkg/provisioning/provision.go
```diff
@@ -0,0 +1,353 @@
...
+	// check if pipeline already exists
+	oldPl, err := s.pipelineService.Get(ctx, id)
+	if err != nil && !cerrors.Is(err, pipeline.ErrInstanceNotFound) {
+		return cerrors.Errorf("could not get the pipeline %q: %w", id, err)
+	} else if err == nil {
+		destMap, srcMap, err = s.copyStatesAndDeletePipeline(ctx, oldPl, config)
+		if err != nil {
+			return cerrors.Errorf("could not copy states from the pipeline %q: %w", id, err)
+		}
+		// todo: create a pipeline copy in case of error, add to rollback
```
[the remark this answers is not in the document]
maha-hajja: yes, this was my plan... we also need to rollback the connectors with their old states.. so we would need to copy the states too.

pkg/provisioning/service.go
```diff
@@ -193,7 +171,17 @@ func (s *Service) collectConnectorStates(ctx context.Context, pl *pipeline.Insta
...
 			connStates[connID] = func(ctx context.Context) error {
-				_, err := s.connectorService.SetDestinationState(ctx, connID, oldState)
+				newConn, err := s.connectorService.Get(ctx, connID)
+				if err != nil {
+					return fmt.Errorf("could not get connector %q: %w", connID, err)
+				}
+				if newConn.Config().Plugin != oldConn.Config().Plugin {
+					s.logger.Warn(ctx).
+						Str(log.PluginNameField, newConn.Config().Plugin).
+						Msg("plugin name was changes, could not apply old destination state so connector will start from the beginning")
```
lovromazgon, line 181:
```suggestion
						Msg("plugin name changed, could not apply old destination state, the connector will start from the beginning")
```

pkg/provisioning/service.go
```diff
@@ -205,8 +193,28 @@ func (s *Service) collectConnectorStates(ctx context.Context, pl *pipeline.Insta
...
+			connStates[connID] = func(ctx context.Context) error {
+				newConn, err := s.connectorService.Get(ctx, connID)
+				if err != nil {
+					return fmt.Errorf("could not get connector %q: %w", connID, err)
+				}
+				if newConn.Config().Plugin != oldConn.Config().Plugin {
+					s.logger.Warn(ctx).
+						Str(log.PluginNameField, newConn.Config().Plugin).
+						Str(log.ConnectorIDField, connID).
+						Msg("plugin name was changes, could not apply old source state so connector will start from the beginning")
```
lovromazgon, line 205:
```suggestion
						Msg("plugin name changed, could not apply old source state, the connector will start from the beginning")
```

pkg/provisioning/service.go
```diff
@@ -234,51 +242,82 @@ func (s *Service) applyConnectorStates(ctx context.Context, pl *pipeline.Instanc
...
 		}
+		s.rollbackDeleteProcessor(ctx, r, pl.ID, processor.ParentTypePipeline, k, v)
 	}
+
+	r.Append(func() error {
+		err := s.applyConnectorStates(ctx, pl, connStates)
+		if err != nil {
+			return err
+		}
+		if pl.Status == pipeline.StatusRunning {
```
lovromazgon, line 267: Given that we first stop the pipeline this will _probably_ not be `StatusRunning`. We might want to move the stopping of the pipeline into this function and if we need to stop it also add the rollback to start the pipeline again.

pkg/provisioning/service.go
```diff
@@ -234,51 +242,82 @@ func (s *Service) applyConnectorStates(ctx context.Context, pl *pipeline.Instanc
...
 		if err != nil {
 			return err
 		}
+		s.rollbackDeleteConnector(ctx, r, pl.ID, k, v)
 	}
 	err = s.pipelineService.Delete(ctx, pl)
 	if err != nil {
 		return err
 	}
+	s.rollbackDeletePipeline(ctx, r, pl.ID, config)
```
lovromazgon, line 308: I'm relatively sure that once we reach this line the rest of the rollbacks won't work anymore (or they will produce an unexpected result). See, the way the pipeline service is written is that it expects the caller to supply a `pipeline.Instance` in all calls. That's different from how the connector and processor service work, they expect the caller to supply an ID. So once we delete the pipeline, the rollbacks will work on an old `pipeline.Instance` that doesn't exist anymore.

I don't see an easy fix, we will need to change the interface of the pipeline service to also expect pipeline IDs 😕 the problem is that this PR is already becoming quite huge, so I'm fine with pushing this down the road and adding a comment here describing the problem. We shouldn't forget about this though.

lovromazgon (commented):
One last thing before I approve this - I would love to see an integration test where we use the actual pipeline, connector and processor service to provision a pipeline 🙏 We should assert that after the pipeline is provisioned all these services contain the expected entities.

Hint: see the first 25 lines of [`TestPipelineSimple`](https://github.com/ConduitIO/conduit/blob/28e765570610ab5d149a919689f4486d07fb3eec/pkg/orchestrator/orchestrator_test.go#L54) how we set up real services in a test.

pkg/provisioning/service.go
maha-hajja: was there a reason why the pipeline service was built differently than the others?

lovromazgon: Honestly, I don't remember. I don't think it was done intentionally.

pkg/provisioning/test/source-file.txt
```diff
@@ -0,0 +1 @@
+just some testing data :D
```
lovromazgon, line 1: That's some happy testing data 😄

PR #588, opened 2022-08-23
Fix a deadlock in source connectors

### LINKED ISSUE #570: Performance: When running at 20k msg/s, progress is being made only briefly
### Bug description

In the workload where we generate 20k mgs/s, progress is being made only briefly, at the beginning of the test. After that, the test report shows no change in record count.

### Steps to reproduce

1. Run the workload here: https://github.com/ConduitIO/streaming-benchmarks/blob/haris/ec2-helpers/workloads/small-messages-20k-msg-per-sec.sh

### Version

{   "version": "v0.3.0-nightly.20220811",   "os": "linux",   "arch": "amd64" }

PS: The test report cannot be attached for some reason. I'll try to add it later.

hariso:
### Description

Fixes a deadlock in the source connector. Also upgrade golangci and fixes two new reported issues.

**Cause of the deadlock**:
Here's a simplified representation of how the source connector ack's records:
```
source.Ack()
    lock connector
    defer unlock connector
  
    update position in connector
    connectorPersister.Persist()
```
Here's how Persist() works:
```
Persist()
    if batch is full
        triggerFlush()

triggerFlush()
    wait for any running flusher to finish
    // trigger a flush in a separate goroutine
    go flushNow()    

flushNow()
    start new DB transaction
    serialize connector -> requires locking the connector
    persist connector state
```
If the source is acking really fast, then it's possible that, while `flushNow()` is still running (starting a new DB txn, but didn't lock the connector yet), another flush needs to be triggered. 

`Ack()` locks the connector, calls `triggerFlush()` (indirectly), which is waiting for the `flushNow()` from previous batch. `flushNow()` tries to acquire a lock, but cannot since `Ack()` is actually waiting on it to be done with the flush, to trigger a new one. 

Fixes #570.

### Quick checks:

### DIFF
--- a/pkg/connector/source_test.go
+++ b/pkg/connector/source_test.go
@@ -16,6 +16,10 @@ package connector
 
 import (
 	"context"
+	"sync"
+	"testing"
+	"time"
+
 	"github.com/conduitio/conduit/pkg/foundation/database/inmemory"
 	"github.com/conduitio/conduit/pkg/foundation/log"
 	"github.com/conduitio/conduit/pkg/plugin"
@@ -24,9 +28,6 @@ import (
 	"github.com/conduitio/conduit/pkg/record"
 	"github.com/google/uuid"
 	"github.com/matryer/is"
-	"sync"
-	"testing"
-	"time"
 )
 
 func TestSource_Ack_Deadlock(t *testing.T) {
--- a/pkg/connector/source_test.go
+++ b/pkg/connector/source_test.go
@@ -1,0 +1,104 @@
+// Copyright © 2022 Meroxa, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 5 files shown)


pkg/connector/source_test.go
```diff
@@ -0,0 +1,105 @@
...
+		}()
+	}
+
+	if waitTimeout(&wg, 100*time.Millisecond) {
+		is.Fail() // timeout reached
+	}
+}
+
+// waitTimeout was copied from pkg/pipeline/stream/stream_test.go
+func waitTimeout(wg *sync.WaitGroup, timeout time.Duration) bool {
```
hariso, line 93: I'm not too happy with this copy-paste, I'll try to come up with something better.

lovromazgon: Don't bother, I'd rather we focus on finishing more important stuff before the release 😉 This can be easily done in a follow-up, besides, I'm sure we need to add these timeouts in multiple places. I'd rather do it separately by adding a utility like we [wanted to do in the SDK](https://github.com/ConduitIO/conduit-connector-sdk/commit/a9290b473be448836a4a0225afffa1215eda8307) and make sure we have timeouts wherever we need them, not only in these two tests.

lovromazgon (approved):
Great work 🚀

pkg/connector/source_test.go
hariso: Agreed. :+1:

Issue #599, opened 2022-08-25
[The Pipeline detail screen] Displays wrong type of connector  in the "Edit" pop-up when edit the connector


uchennakevinm1:
![Image](https://user-images.githubusercontent.com/102546122/186547060-bdf532d5-b578-459f-b542-ffb8329111b5.png)

PR #618, opened 2022-08-31
Fix connector edit type mismatch && Display UI version

jmar910:
### Description

Please include a summary of the change and what type of change it is (new feature, bug fix, refactoring, documentation).
Please also include relevant motivation and context.
List any dependencies that are required for this change.

Fixes https://github.com/ConduitIO/conduit/issues/599

### Quick checks:

### DIFF
--- a/ui/app/components/pipeline-editor.hbs
+++ b/ui/app/components/pipeline-editor.hbs
@@ -176,6 +176,7 @@
       </div>
     </div>
   </div>
+  <div class="absolute bottom-0 right-2 text-sm text-black">{{app-version shaOnly=true}}</div>
 </div>
 {{#if this.isShowingNewConnectorModal}}
   <PipelineEditor::ConnectorModal
--- a/ui/app/components/pipeline-editor.hbs
+++ b/ui/app/components/pipeline-editor.hbs
@@ -189,7 +189,7 @@
 {{/if}}
 {{#if this.isShowingEditConnectorModal}}
   <PipelineEditor::ConnectorModal
-    @newConnectorType={{this.newConnectorType}}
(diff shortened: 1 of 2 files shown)


ui/app/templates/application.hbs
```diff
@@ -19,6 +19,10 @@
         </LinkTo>
       </div>
 
+      <div class="absolute bottom-0 font-semibold text-sm text-white">{{app-version
```
jayjayjpg, line 22: That's a neat helper!

ui/app/components/pipeline-editor.hbs
```diff
@@ -189,7 +189,7 @@
 {{/if}}
 {{#if this.isShowingEditConnectorModal}}
   <PipelineEditor::ConnectorModal
-    @newConnectorType={{this.newConnectorType}}
+    @newConnectorType={{this.selectedNode.type}}
```
jayjayjpg, line 192: Would it be fairly straightforward to create a regression test for this scenario? I'm not sure how much mocking we already have in place for this to be emulated in an e.g. application test

What do you think?

jayjayjpg (approved):
This reads great! I left one more suggestion on regression testing, but feel free to drop it or pick it up in a follow-up PR -- thank you for the swift fix! :sparkles:

ui/app/components/pipeline-editor.hbs
jmar910: Yep we definitely could. Its a weird bug that is due to leftover state from opening the connector modal, so we just have to write a specific acceptance test for it. I can do it in a follow up 👍

PR #628, opened 2022-09-02
Update Buf

lovromazgon:
### Description

This PR updates the way we use protobufs in this repository to match the approach in https://github.com/ConduitIO/conduit-connector-protocol. We don't need to generate protobuf code locally anymore, it is pushed to the buf schema registry every time we merge a PR with a change in the proto folder. If a user wants to generate the gRPC code (or if we want to provide generated clients ourselves) it's now as easy as creating 1 file and running buf.

The swagger doc is also generated remotely now, I added a `go:generate` instruction to copy it from the dependency into our project, because we need to embed it in the swagger UI.

**NOTE**: buf validation will fail in the PR because the buf file was moved, it compares the contents of the branch to the contents of main, no way to fix it in this PR. Once this PR is merged it will work correctly (it works in the connector protocol repo).

### Quick checks:

- [X] I have written unit tests. (no code actually changed)


hariso (approved):
:zap:

PR #631, opened 2022-09-06
Load pipeline config files on startup

### LINKED ISSUE #33: Load pipeline config files on startup

maha-hajja:
* add cli flag for pipeline directory 
* create provisioning.Init() that loops through the pipelines Dir
* delete old pipelines that were provisioned by config

Fixes #33 

### Quick checks:


pkg/provisioning/enrich.go
```diff
@@ -23,19 +23,35 @@ func EnrichPipelinesConfig(mp map[string]PipelineConfig) map[string]PipelineConf
...
 
 // enrichConnectorsConfig sets default values for connectors config fields
-func enrichConnectorsConfig(mp map[string]ConnectorConfig) map[string]ConnectorConfig {
+func enrichConnectorsConfig(mp map[string]ConnectorConfig, pipelineID string) map[string]ConnectorConfig {
 	for k, cfg := range mp {
 		if cfg.Name == "" {
 			cfg.Name = k
 		}
-		mp[k] = cfg
+		// attache the pipelineID to the connectorID
```
hariso, line 39: Nitpick: attache -> attach.

pkg/provisioning/service.go
```diff
@@ -43,40 +48,79 @@ func NewService(
...
 		pipelineService:  plService,
 		connectorService: connService,
 		processorService: procService,
+		pipelinesPath:    pipelinesDir,
 	}
 }
 
-func (s *Service) ProvisionConfigFile(ctx context.Context, path string) ([]string, error) {
+func (s *Service) Init(ctx context.Context) error {
+	s.logger.Debug(ctx).Msg("initializing the provisioning service")
```
hariso, line 64: Unless we're already doing it somewhere, we should log the pipelines directory.

pkg/provisioning/service.go
```diff
@@ -43,40 +48,79 @@ func NewService(
...
+		}
+		return nil
+	})
+	if err != nil {
+		multierr = multierror.Append(multierr, cerrors.Errorf("could not iterate through the pipelines folder %q: %w", s.pipelinesPath, err))
+		return multierr
+	}
+
+	if len(files) == 0 {
+		s.logger.Warn(ctx).Str("pipelinesDir", s.pipelinesPath).Msg("configuration folder has no YAML files, no pipelines created")
```
hariso, line 80: Maybe we can `return` at this point?

pkg/provisioning/service.go
```diff
@@ -43,40 +48,79 @@ func NewService(
...
 	before, err := Parse(data)
 	if err != nil {
-		return nil, err
+		return nil, cerrors.Errorf("could not parse the file %q: %w", path, err)
 	}
 
 	got := EnrichPipelinesConfig(before)
 
 	var pls []string
-	var multierr error
```
hariso, line 70: Out of curiosity, why did you move the declaration a few lines above, when `multierr` is not used?

maha-hajja: Oh I was gonna add the error from `Parse(data)` to it.. so moved it at the top.. but then changed it back

pkg/provisioning/service.go
maha-hajja: if we return here, then we won't delete old pipelines

pkg/provisioning/service.go
```diff
@@ -43,25 +48,63 @@ func NewService(
...
+	for _, file := range files {
+		provPipelines, err := s.provisionConfigFile(ctx, file)
+		if err != nil {
+			multierr = multierror.Append(multierr, err)
+		}
+		if len(provPipelines) != 0 {
+			pipelines = append(pipelines, provPipelines...)
+		}
+	}
+	s.logger.Info(ctx).Int("count", len(pipelines)).Str("pipelinesDir", s.pipelinesPath).Str("pipelines", strings.Join(pipelines, ", ")).Msg("pipelines successfully provisioned")
```
hariso, line 92: Nitpick: You can use Strs() instead of Str(), so you don't join the strings yourself. Also, it might be nice seeing all the individual calls on separate lines, i.e.
```
log.Info()
    .Str()
    .Msg()
```

hariso (approved):

pkg/provisioning/service.go
```diff
@@ -418,6 +461,76 @@ func (s *Service) createProcessors(ctx context.Context, r *rollback.R, parentID
 	return nil
 }
 
+func (s *Service) deleteOldPipelines(ctx context.Context, pipelines []string) {
+	var err error
```
lovromazgon, line 465: Nitpick: this line can be removed if we use `:=` when assigning to the variable.

pkg/provisioning/service.go
```diff
@@ -418,6 +461,76 @@ func (s *Service) createProcessors(ctx context.Context, r *rollback.R, parentID
...
+	plMp := s.pipelineService.List(ctx)
+	for id, pl := range plMp {
+		if !slices.Contains(pipelines, id) && pl.ProvisionedBy == pipeline.ProvisionTypeConfig {
+			err = s.deletePipelineWithoutRollback(ctx, pl)
+			if err != nil {
+				s.logger.Warn(ctx).Str("pipelineID", id).Msg("could not delete pipeline")
```
lovromazgon, line 471: We have constants for common log fields. I would also add more information to this error to let the user know what's going on. Along the way of "failed to delete pipeline provisioned by a config file, the pipeline is probably in a broken state, Conduit will try to remove it again next time it runs"
```suggestion
				s.logger.Warn(ctx).Str(log.PipelineIDField, id).Msg("could not delete pipeline")
```

pkg/provisioning/service.go
```diff
@@ -43,25 +48,63 @@ func NewService(
...
 	}
 }
 
-func (s *Service) ProvisionConfigFile(ctx context.Context, path string) ([]string, error) {
+func (s *Service) Init(ctx context.Context) error {
+	s.logger.Debug(ctx).Str("pipelinesDir", s.pipelinesPath).Msg("initializing the provisioning service")
+
+	var files []string
+	err := filepath.WalkDir(s.pipelinesPath, func(path string, fileInfo fs.DirEntry, err error) error {
+		if strings.HasSuffix(path, ".yml") {
```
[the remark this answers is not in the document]
lyuboxa: I wonder why not use one yaml file rather than multiple? Won't that make processing simpler?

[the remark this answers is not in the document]
lovromazgon: We support both approaches, it's your choice which one you take. Once you reach a certain number of pipelines it will be nicer to manage them in separate files IMO.

pkg/provisioning/service.go
```diff
@@ -43,25 +48,63 @@ func NewService(
...
+		if err != nil {
+			multierr = multierror.Append(multierr, err)
+		}
+		if len(provPipelines) != 0 {
+			pipelines = append(pipelines, provPipelines...)
+		}
+	}
+	s.logger.Info(ctx).Int("count", len(pipelines)).Str("pipelinesDir", s.pipelinesPath).Str("pipelines", strings.Join(pipelines, ", ")).Msg("pipelines successfully provisioned")
+
+	s.deleteOldPipelines(ctx, pipelines)
```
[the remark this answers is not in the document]
maha-hajja: I think it deserves a log msg on its own.. something like:

> 	s.logger.Info(ctx).
		Int("count", len(deletedPls)).
		Strs("pipelines", deletedPls).
		Msg("old pipelines that were deleted")
WDYT? and do you mean returning this info from the `init()` function?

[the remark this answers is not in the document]
lovromazgon: I mean that `deleteOldPipelines` would have to return the number of deleted pipelines so we can even log it.

It's a matter of taste, I personally would prefer to have 1 info line that contains all the info about what was the result of pipeline provisioning, for example:
```
pipeline configs provisioned component=provisioning.Service created=3 deleted=2 pipelines_path=/Users/lovro/dev/conduitio/conduit/pipelines
```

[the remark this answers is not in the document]
lyuboxa: > 

I like that single line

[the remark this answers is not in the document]
maha-hajja: should I also add the created and deleted pipeline names? or just the number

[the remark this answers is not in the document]
lovromazgon: That's a good idea, we could list the pipeline names (or maybe IDs) instead of providing the count.

```
pipeline configs provisioned component=provisioning.Service created=[pipeline1 pipeline2] deleted=[old-pipeline] pipelines_path=/Users/lovro/dev/conduitio/conduit/pipelines
```

pkg/provisioning/service.go
```diff
@@ -89,9 +87,15 @@ func (s *Service) Init(ctx context.Context) error {
...
 
-	s.deleteOldPipelines(ctx, pipelines)
+	deletedPls := s.deleteOldPipelines(ctx, pipelines)
+	s.logger.Info(ctx).
+		Int("created", len(pipelines)).
+		Int("deleted", len(deletedPls)).
+		Str("pipelines_path", s.pipelinesPath).
+		Strs("created_pipelines", pipelines).
+		Strs("deleted_pipelines", deletedPls).
+		Msg("pipeline configs provisioned")
```
lovromazgon, line 98: Nitpick / Suggestion: I'd keep it simple, people can count if they want to 😄
```suggestion
	s.logger.Info(ctx).
		Strs("created", pipelines).
		Strs("deleted", deletedPls).
		Str("pipelines_path", s.pipelinesPath).
		Msg("pipeline configs provisioned")
```

lovromazgon (approved):
🚀 good work 🙌

PR #635, opened 2022-09-14
docs for configuration files

maha-hajja:
### Description
some documentation for configurations files, how to use them, and yaml file format

### Quick checks:


neovintage:
@maha-hajja I don't understand the `status` field in the schema. Why would I ever do anything besides "running"?

maha-hajja:
@neovintage you're mostly going to have your pipeline as `running`, but in some cases, it could be helpful to have it as `stopped` too: 
* maybe the pipeline is only one part of a bigger project, so you could prepare the pipeline and have it ready in conduit, then set the status to `running` once everything else is in place
* maybe the pipeline needs to be stopped, but should stay in conduit because you want to resume it at a later point in time
* maybe you have a pipeline that you want to only run on demand (e.g. a pipeline that processes DLQ messages)

credit to @lovromazgon for giving these examples.

docs/pipeline_configuration.md
```diff
@@ -0,0 +1,75 @@
+## Overview
+Creating pipelines in Conduit is a simple process with many steps to define the resources and their configurations, this
+could be done through API or Conduit's UI. But to make this process even simpler and to make it possible to load 
+pre-defined pipelines, we introduce **_Pipeline Configurations_**.
```
lovromazgon, line 4: IMO we should add the word "file" somewhere in the mix (e.g. pipeline configuration files). I would also add it in the name of the file itself (e.g. `pipeline_configuration_files.md`).

docs/pipeline_configuration.md
```diff
@@ -0,0 +1,75 @@
...
+
+Using **_Pipeline Configurations_** you will define all the pipelines you want to be provisioned into Conduit at startup,
+only by creating `.yml` files that define all the pipelines, with their connectors, processors, and configurations.
+
+## Getting started
+create a folder called _"pipelines"_ at the same level as your Conduit binary file, add all your `.yml` files there,
```
lovromazgon, line 10:
```suggestion
Create a folder called _"pipelines"_ at the same level as your Conduit binary file, add all your `.yml` files there, 
```

docs/pipeline_configuration.md
```diff
@@ -0,0 +1,75 @@
+## Overview
+Creating pipelines in Conduit is a simple process with many steps to define the resources and their configurations, this
+could be done through API or Conduit's UI. But to make this process even simpler and to make it possible to load 
+pre-defined pipelines, we introduce **_Pipeline Configurations_**.
+
+Using **_Pipeline Configurations_** you will define all the pipelines you want to be provisioned into Conduit at startup,
+only by creating `.yml` files that define all the pipelines, with their connectors, processors, and configurations.
```
lovromazgon, line 7:
```suggestion
Using pipeline configuration files you can define pipelines that are provisioned by Conduit at startup.
It's as simple as creating a `.yml` file that defines pipelines, connectors, processors, and their corresponding configurations.
```

docs/pipeline_configuration.md
```diff
@@ -0,0 +1,75 @@
...
+
+## Getting started
+create a folder called _"pipelines"_ at the same level as your Conduit binary file, add all your `.yml` files there, 
+then run Conduit using the cmd command:
```
lovromazgon, line 11: Seems like `cmd` is redundant.

docs/pipeline_configuration.md
```diff
@@ -0,0 +1,75 @@
...
+only by creating `.yml` files that define all the pipelines, with their connectors, processors, and configurations.
+
+## Getting started
+create a folder called _"pipelines"_ at the same level as your Conduit binary file, add all your `.yml` files there, 
+then run Conduit using the cmd command: 
+```
+./conduit
+```
+If you have your YAML files at a different directory, simply run conduit with the cli flag `pipelines.path` and point to 
+your directory, ex:
```
lovromazgon, line 16:
```suggestion
If you have your YAML files in a different directory, simply run Conduit with the CLI flag `pipelines.path` and point to 
your directory:
```

docs/pipeline_configuration.md
```diff
@@ -0,0 +1,75 @@
...
+```
+./conduit
+```
+If you have your YAML files at a different directory, simply run conduit with the cli flag `pipelines.path` and point to 
+your directory, ex: 
+```
+./conduit -pipeline.path ../my-directory
+```
+
+_**Note**_: provisioned pipelines are **immutable**, any updates needed on a provisioned pipeline have to be done through
```
lovromazgon, line 21: I think immutability is a very important aspect that should have its own chapter rather than only a note. I would move it after the file format though, I think that's what most people will look for at first.

docs/pipeline_configuration.md
```diff
@@ -0,0 +1,75 @@
...
+the configuration file it was provisioned from. you can only control stopping and starting a pipeline through the UI and API.
```
lovromazgon, line 22:
```suggestion
the configuration file it was provisioned from. You can only control stopping and starting a pipeline through the UI and API.
```

docs/pipeline_configuration.md
```diff
@@ -0,0 +1,75 @@
...
+
+### YAML file format
```
lovromazgon, line 24: Maybe "YAML Schema" would be better?

docs/pipeline_configuration.md
```diff
@@ -0,0 +1,75 @@
...
+The file in general has two root keys, the "version", and the "pipelines" map. The map consists of other elements like
```
lovromazgon, line 25: I suggest using a backtick when quoting fields (e.g. `version` instead of "version").

docs/pipeline_configuration.md
```diff
@@ -0,0 +1,75 @@
...
+The file in general has two root keys, the "version", and the "pipelines" map. The map consists of other elements like 
+"status" and "name", which are configurations for the pipeline itself.
+
+To create connectors in that pipeline, simply add another map under the pipeline map, and call it "connectors".
+
+To create processors, either add a "processors" map under a pipeline ID, or under a connector ID, depending on its parent.
+check this YAML file example with explanation for each field:
+
+``` yaml
+version: 1.0                    # parser version, the only supported version for now is 1.0 [mandatory]
```
lovromazgon, line 34: I'd add a blank line after the version.

docs/pipeline_configuration.md
```diff
@@ -0,0 +1,75 @@
...
+pipelines:                      # a map of pipelines IDs and their configurations.
+  pipeline1:                    # pipeline ID, has to be unique.
+    status: running             # pipelines status at startup, either running or stopped. [mandatory]
+    name: pipeline1             # pipeline name. [optional]
+    description: desc           # pipeline describtion. [optional]
```
lovromazgon, line 39: Typo "describtion".

docs/pipeline_configuration.md
```diff
@@ -0,0 +1,75 @@
...
+
+To create processors, either add a "processors" map under a pipeline ID, or under a connector ID, depending on its parent.
+check this YAML file example with explanation for each field:
+
+``` yaml
+version: 1.0                    # parser version, the only supported version for now is 1.0 [mandatory]
+pipelines:                      # a map of pipelines IDs and their configurations.
+  pipeline1:                    # pipeline ID, has to be unique.
+    status: running             # pipelines status at startup, either running or stopped. [mandatory]
+    name: pipeline1             # pipeline name. [optional]
```
lovromazgon, line 38: We can mention that Conduit will use the pipeline ID as the name if no name is specified.

docs/pipeline_configuration.md
```diff
@@ -0,0 +1,75 @@
...
+    description: desc           # pipeline describtion. [optional]
+    connectors:                 # a map of connectors IDs and their configurations.
+      con1:                     # connector ID, has to be unieuq per pipeline.
```
lovromazgon, line 41: Typo "unieuq".

docs/pipeline_configuration.md
```diff
@@ -0,0 +1,75 @@
...
+        type: source            # connector type, either "source" or "destination". [mandatory]
+        plugin: builtin:file    # connector plugin. [mandatory]
+        name: con3              # connector name. [optional]
```
lovromazgon, line 44: Same here, I think by default we use the connector ID as the name, right?

docs/pipeline_configuration.md
```diff
@@ -0,0 +1,75 @@
...
+        settings:               # map of configurations keys and their values.
+          path: ./file1.txt     # for this example, the plugin "bultin:file" has only one configuration, which is path.
+      con2:
+        type: destination
+        plugin: builtin:file
+        name: file-dest
+        settings:
+          path: ./file2.txt
+        processors:             # a map of processor IDs and their configurations, "con2" is the processor parent.
+          proc1:                 # processor ID, has to be unique for each parent
```
lovromazgon, line 54: Misaligned comment (one space too many).

docs/pipeline_configuration.md
```diff
@@ -0,0 +1,75 @@
...
+```
+
+### Updates and Deletes
+Updates and deletes for a provisioned pipeline can only be done through the configuration files.
+
+* To delete a pipeline: simply delete it from the pipelines map from the configuration file, then run conduit again.
+* To update a pipeline: change any field value from the configuration file, and run conduit again to address these updates.
+  
+Note: Updates would keep the status of the pipeline and it would continue working from where it stopped. However, 
+  If the Updated configurations were something like "connector-id", "plugin", "pipeline-id", or "processor-id", then
```
lovromazgon, line 73: We should be very specific about which changes will cause a source connector to lose state. "Something like" sounds very vague and keeps me wondering what else might cause a pipeline to start from the beginning.

Also, I feel like this shouldn't be a "note" but just a normal paragraph.

docs/pipeline_configuration.md
```diff
@@ -0,0 +1,75 @@
+## Overview
```
lovromazgon, line 1: I would expect the main header to be "Pipeline configuration files" if you ask me.

lovromazgon:
Whoops, submitted the review a tad too early - I wanted to point out that you did a nice job, I think the doc gives a nice overview. I personally would prefer if we provided more info (e.g. what happens if a file is invalid, what happens if multiple pipelines have the same ID, what if the pipeline folder is not found, the fact that Conduit searches only for files with extensions `.yml` and `.yaml` recursively in all subfolders, that env variables are supported, that Conduit won't reload changed files unless it's restarted, that connector IDs and processor IDs will get the parent ID prefixed...). But even the current info in the doc is already very valuable.

docs/pipeline_configuration_files.md
```diff
@@ -0,0 +1,96 @@
...
+```
+
+If the file is invalid (missed a mandatory field, or has an invalid configuration value), then the pipeline that has the
+invalid value will be skipped, with an error message logged.
+
+If two pipelines in one file have the same ID, or the `version` field was not specified, then the file would be 
+non-parsable and will be skipped with an error message logged.
+
+If two pipelines from different files have the same ID, the second pipeline will be skipped, with an error message
+specifying which pipeline was not provisioned.
```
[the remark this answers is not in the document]
maha-hajja: because this will be discovered while provisioning, the first case is an invalid yaml file which will be known while parsing, and we parse and provision one file at a time. So we wouldn't know if another file has the same pipelineID, and if it does then we only want to skip the pipeline not skip the whole file like in the first case.

docs/pipeline_configuration_files.md
```diff
@@ -0,0 +1,96 @@
...
+has a connector parent, then the processor ID will be `pipelineID:connectorID:processorID`. 
+
+## Pipelines Immutability
+Pipelines provisioned by configuration files are **immutable**, any updates needed on a provisioned pipeline have to be 
+done through the configuration file it was provisioned from. You can only control stopping and starting a pipeline 
+through the UI or API.
+
+### Updates and Deletes
+Updates and deletes for a provisioned pipeline can only be done through the configuration file that the pipeline was
+originally provisioned from. Updates should be done to the files, then Conduit has to be restarted to reload the changes.
```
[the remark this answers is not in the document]
maha-hajja: yes yes.. will rephrase to make it clearer

docs/pipeline_configuration_files.md
```diff
@@ -0,0 +1,96 @@
...
+
+* To delete a pipeline: simply delete it from the `pipelines` map from the configuration file, then run conduit again.
+* To update a pipeline: change any field value from the configuration file, and run conduit again to address these updates.
+  
+Updates will preserve the status of the pipeline, and will continue working from where it stopped. However,
+the pipeline will not copy the old state and will start from the beginning of the source if one of these values was updated:
```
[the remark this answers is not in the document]
maha-hajja: rephrased.. I don't think users should be concerned about the state at all

docs/pipeline_configuration_files.md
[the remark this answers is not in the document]
hariso: > if it does then we only want to skip the pipeline not skip the whole file like in the first case.

This clears it up for me, thanks!

hariso (approved):

PR #638, opened 2022-09-21
change pipeline interface to accept pipelineID 

### LINKED ISSUE #625: Feature: change the pipeline interface to accept the id (not only the pipeline instance)
### Feature description

for now, the pipeline interface methods expect the pipeline instance as a parameter (connector and processor services expect the entity id instead).
and this would cause a problem with rolling back an error in pipeline provisioning. https://github.com/ConduitIO/conduit/pull/567#discussion_r958749964

we should either add the option to use (pipeline instance or id), or change all the methods to only accept the pipeline id.

maha-hajja:
### Description

change pipeline interface to accept pipelineID instead of pipeline instance

Fixes #625 

### Quick checks:


lovromazgon (approved):
LGTM 🚀

Issue #592, opened 2022-08-23
[Pipelines sreen] The user can't go to the "Pipeline detail" screen


uchennakevinm1:
[Pipelines sreen] The user can't go to the "Pipeline detail" screen if the "failed to delete pipeline: pipeline has connectors attached" or "failed to delete pipeline: pipeline is running" errors have been handled before

uchennakevinm1:
https://user-images.githubusercontent.com/102546122/186279037-47924a63-d50f-4505-b096-5e4c930c3a47.mp4

jmar910:
@uchennakevinm1 fix is coming here https://github.com/ConduitIO/conduit/pull/641

PR #641, opened 2022-09-21
[UI] Rollback pipeline delete if error

jmar910:
### Description
We don't explicitly catch on pipeline destroy when the API errors, so the pipeline itself was going into a deleted state and was causing issues with ember refetching the model. Now we catch on pipeline destroy and do a rollback, which appropriately resets the state. 

Fixes https://github.com/ConduitIO/conduit/issues/592

### Quick checks:

### DIFF
--- a/ui/app/controllers/pipelines.js
+++ b/ui/app/controllers/pipelines.js
@@ -25,7 +25,11 @@ export default class PipelinesController extends Controller {
 
   @action
   async destroyPipeline(pipeline) {
-    await pipeline.destroyRecord();
+    try {
+      await pipeline.destroyRecord();
+    } catch (e) {
+      pipeline.rollbackAttributes();
+    }
     this.setConfirmDeletePipeline(null);
     this.router.transitionTo('pipelines');
   }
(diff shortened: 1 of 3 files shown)


ui/app/controllers/pipelines.js
```diff
@@ -25,7 +25,11 @@ export default class PipelinesController extends Controller {
 
   @action
   async destroyPipeline(pipeline) {
-    await pipeline.destroyRecord();
+    try {
+      await pipeline.destroyRecord();
+    } catch (e) {
```
jayjayjpg, line 30: Nice catch! I vaguely remember us having to guard check against a similar issue on plaform-ui before, so this is great!

ui/tests/acceptance/pipelines/index-test.js
```diff
@@ -0,0 +1,53 @@
...
+  module('deleting a pipeline that is running', function (hooks) {
+    hooks.beforeEach(async function () {
+      this.pipeline = this.server.create(
+        'pipeline',
+        { state: { status: 'STATUS_RUNNING' } },
+        'withFileConnectors'
+      );
+
+      this.server.delete('/pipelines/:id', function () {
+        return new Response(
```
jayjayjpg, line 30: Nice test case! :sparkles:

jayjayjpg (approved):
This reads great and has awesome test coverage -- thank you for the super quick fix! :sparkles:

PR #642, opened 2022-09-22
Redirect handler for /openapi

lovromazgon:
Simple, yet effective. URL `/openapi` is now redirected to `/openapi/` (we need the trailing slash because otherwise relative paths don't work in the Swagger UI).


jmar910 (approved):
🙌

PR #647, opened 2022-09-23
fix pipeline configuration files duplicated pipeline ID bug

### LINKED ISSUE #646: Bug: Duplicated pipeline id in pipeline configuration files fails
### Bug description

a duplicated pipeline id in one pipeline configuration file produces an error (as expected), but having the same pipeline id in two different yaml files crashes Conduit.

### Steps to reproduce

1. create two pipeline configuration files with the same pipeline id
2. run Conduit

### Version

v0.3.0-nightly.20220921-1-ga727888-dirty darwin/amd64

maha-hajja:
### Description

Fixes #646 

### Quick checks:

### DIFF
--- a/pkg/provisioning/service.go
+++ b/pkg/provisioning/service.go
@@ -80,7 +80,7 @@ func (s *Service) Init(ctx context.Context) error {
 	var successPls []string
 	var allPls []string
 	for _, file := range files {
-		provPipelines, all, err := s.provisionConfigFile(ctx, file)
+		provPipelines, all, err := s.provisionConfigFile(ctx, file, successPls)
 		if err != nil {
 			multierr = multierror.Append(multierr, err)
 		}
@@ -101,7 +101,7 @@ func (s *Service) Init(ctx context.Context) error {
 	return multierr
 }
 
-func (s *Service) provisionConfigFile(ctx context.Context, path string) ([]string, []string, error) {
+func (s *Service) provisionConfigFile(ctx context.Context, path string, alreadyProvisioned []string) ([]string, []string, error) {
 	data, err := os.ReadFile(path)
 	if err != nil {
 		return nil, nil, cerrors.Errorf("could not read the file %q: %w", path, err)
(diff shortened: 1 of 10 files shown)


pkg/provisioning/service.go
```diff
@@ -80,7 +80,7 @@ func (s *Service) Init(ctx context.Context) error {
 	var successPls []string
 	var allPls []string
 	for _, file := range files {
-		provPipelines, all, err := s.provisionConfigFile(ctx, file)
+		provPipelines, all, err := s.provisionConfigFile(ctx, file, successPls)
```
hariso, line 83: It would be helpful if there was a comment explaining the difference between `allPls` and `all`.

pkg/provisioning/service.go
```diff
@@ -101,7 +101,7 @@ func (s *Service) Init(ctx context.Context) error {
 	return multierr
 }
 
-func (s *Service) provisionConfigFile(ctx context.Context, path string) ([]string, []string, error) {
+func (s *Service) provisionConfigFile(ctx context.Context, path string, alreadyProvisioned []string) ([]string, []string, error) {
```
hariso, line 104: It would be nice to have a comment explaining what this method returns.

pkg/provisioning/service_test.go
```diff
@@ -577,7 +577,7 @@ func TestProvision_MultiplePipelines(t *testing.T) {
 
 	service := NewService(db, logger, pipelineService, connService, procService, "./test/pipelines3")
 	err := service.Init(context.Background())
-	is.NoErr(err)
+	is.True(err != nil) // duplicated pipeline id
```
hariso, line 580: IIUC, this re-purposes the test for loading multiple pipelines, to actually check for duplicated. IMHO, it would be good to have both tests, i.e. checking if we can load multiple pipelines (happy path) and duplicates in different files.

maha-hajja: it still loads multiple pipelines and they are all created successfully {we made sure of that by the service.Expect() lines}, the only change will be this line, because no other methods will be expected to be called cuz the pipeline will be skipped.
I can easily copy it of course.. but lemme know if you still think it's better

hariso (approved):

Issue #427, opened 2022-05-27
Local directory discovery of connectors


neovintage:
On Conduit server, if I have a directory, conduit will be able to see them and use them for pipelines.

lovromazgon:
After this change Conduit will only run plugins it can find in a certain folder. We need to define how plugins can be referred to. Here's my proposal.

`[PLUGIN-TYPE:]PLUGIN-NAME[@VERSION]`

- `PLUGIN-TYPE` (`builtin` or `standalone`)
  - Defines if the specified plugin should be builtin or standalone.
  - The plugin type is optional. If empty, Conduit will use a builtin plugin if it exists and fall back to a standalone plugin.
- `PLUGIN-NAME`
  - Defines the name of the plugin as specified in the plugin specifications, it has to be an exact match.
- `VERSION`
  - Defines the plugin version as specified in the plugin specifications, it has to be an exact match.
  - The version is optional. If empty, Conduit will default to the latest semantic version.

Examples:
- `postgres`
  - will use the **latest** **builtin** **postgres** plugin
  - will fallback to the **latest** **standalone** **postgres** plugin
- `postgres@v0.2.0`
  - will use the **builtin** **postgres** plugin with version **v0.2.0**
  - will fallback to a **standalone** **postgres** plugin with version **v0.2.0**
- `builtin:postgres`
  - will use the **latest** **builtin** **postgres** plugin (no fallback to standalone)
- `standalone:postgres@v0.3.0`
  - will use the **standalone** **postgres** plugin with version **v0.3.0**

PR #651, opened 2022-09-26
[UI] Plugin lookup logic

### LINKED ISSUE #644: Bug: UI requests plugin on inexistent endpoint
### Bug description

If the pipeline contains a plugin that is unknown to the UI (I assume the plugin fetches a list of plugins and stores it), the UI will try to request it on the URL `v1/plugins/[plugin-name]`. The endpoint doesn't exist.

We can either change how the UI works or provide the endpoint. Creating such an endpoint would make the plugin endpoints consistent with other entities in Conduit.

### Steps to reproduce

1. Run Conduit with `./conduit -pipelines.path ./examples/pipelines`
   * this will run the example file-to-file pipeline which contains two connectors using the plugin `builtin:file`
2. Open the UI and open the pipeline
3. See an error modal, also see an error in the logs:
   ```
   ERR error processing HTTP request error="rpc error: code = NotFound desc = Not Found" http_endpoint="GET /v1/plugins/builtin:file" request_id=1c0e9a73-15f2-4d06-ab79-dafa79e8b813 stack=null
   ```

### Version

v0.3.0-nightly.20220922 darwin/amd64

jmar910:
### Description
A bug came up at the end of last week that I discussed with @lovromazgon due to the changes outlined here https://github.com/ConduitIO/conduit/issues/427#issuecomment-1227396725 (see also https://github.com/ConduitIO/conduit/pull/609)

While this isn't a show stopping bug, it does limit the functionality of the UI when setting relationships between connectors and plugins using the new format.

As a nice-to-have and a temporary fix, I duplicate lookup logic outlined in the above comment/PR to allow for connectors to use the shortened names or special tags (`any`, `latest`) to match the right plugins.

Fixes #644 

BEFORE
Screen Shot 2022-09-26 at 6 36 12 PM

AFTER
Screen Shot 2022-09-26 at 6 37 14 PM

### Quick checks:

### DIFF
--- a/ui/app/serializers/connector.js
+++ b/ui/app/serializers/connector.js
@@ -27,8 +27,86 @@ export default class ConnectorSerializer extends ApplicationSerializer {
 
   extractRelationship(modelName, value) {
     if (modelName === 'plugin' && value) {
+      const pluginNames = this.store.peekAll('plugin').mapBy('name');
+
+      const hasOnlyType = /^(any|builtin|standalone):[\w-]+$/.test(value);
+      const hasOnlyVersion = /^[\w-]+@v\S+$/.test(value);
+      const hasTypeAndVersion = /(any|builtin|standalone):[\w-]+@v\S+$/.test(
+        value
+      );
+
+      let pluginName = value;
+      let pluginType = 'any';
+      let pluginVersion = 'latest';
+
+      if (hasOnlyType) {
+        pluginType = value.split(':')[0];
+        pluginName = value.split(':')[1];
+      }
+
+      if (hasOnlyVersion) {
+        pluginName = value.split('@')[0];
+        pluginVersion = value.split('@')[1];
+      }
+
+      if (hasTypeAndVersion) {
+        const split = value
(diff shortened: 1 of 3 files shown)


ui/app/serializers/connector.js
```diff
@@ -27,8 +27,86 @@ export default class ConnectorSerializer extends ApplicationSerializer {
 
   extractRelationship(modelName, value) {
     if (modelName === 'plugin' && value) {
+      const pluginNames = this.store.peekAll('plugin').mapBy('name');
```
jayjayjpg, line 30: Would it make sense to add a unit test for either the `extractRelationship` method of the serializer or an acceptance test asserting the correct display of connector labels in the UI?

This way, we could not only guard check against regressions, but also make it easier to add new rules for the labelling in the future

jayjayjpg (approved):
This reads great and thank you for the quick fix! :sparkles: I left one more suggestion on unit / acceptance testing, but otherwise this is already good to go!

PR #654, opened 2022-09-28
housekeeping: buf mod update

lovromazgon:
I ran `buf mod update` to update the dependencies, after we merge this the buf action should kick in and upload the proto definitions to the buf schema registry and fix the 404 error on the BSR (https://buf.build/conduitio/conduit).


maha-hajja (approved):
lgtm

PR #663, opened 2022-10-06
rename processor's name to type

### LINKED ISSUE #498: Rename Processor.Name to Processor.Type
Now that we removed the processor type (transform/filter) we can use the field name "type" for defining which processor needs to be executed (e.g. `hoistfield`, `js`, etc.). This would make for a nicer design, right now the field "name" is also present in the connector and pipeline, although there it serves a different purpose (it's a freeform field, it's mutable and is part of the config). I think the field was named "name" originally because "type" was already taken, and since we now removed that field we can rename "name" to "type".

The pipeline config file will already use this field name so this will consolidate the API with the pipeline config file. This will require a change in the UI though.

Sidenote: remember to update the [processor gauge](https://github.com/ConduitIO/conduit/pull/486#discussion_r907511260).

maha-hajja:
Fixes #498 

### Quick checks:

### DIFF
--- a/pkg/web/api/processor_v1.go
+++ b/pkg/web/api/processor_v1.go
@@ -34,7 +34,7 @@ type ProcessorOrchestrator interface {
 	// Get will return a single Interface or an error if it doesn't exist.
 	Get(ctx context.Context, id string) (*processor.Instance, error)
 	// Create will make a new Interface.
-	Create(ctx context.Context, name string, parent processor.Parent, cfg processor.Config) (*processor.Instance, error)
+	Create(ctx context.Context, procType string, parent processor.Parent, cfg processor.Config) (*processor.Instance, error)
 	// Update will update a Interface's config.
 	Update(ctx context.Context, id string, cfg processor.Config) (*processor.Instance, error)
 	// Delete removes a processor
--- a/pkg/web/api/toproto/processor.go
+++ b/pkg/web/api/toproto/processor.go
@@ -30,7 +30,7 @@ func _() {
 func Processor(in *processor.Instance) *apiv1.Processor {
 	return &apiv1.Processor{
 		Id:        in.ID,
-		Name:      in.Name,
+		Name:      in.Type,
 		CreatedAt: timestamppb.New(in.CreatedAt),
 		UpdatedAt: timestamppb.New(in.UpdatedAt),
 		Config:    ProcessorConfig(in.Config),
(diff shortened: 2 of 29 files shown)


pkg/web/api/toproto/processor.go
```diff
@@ -30,7 +30,7 @@ func _() {
 func Processor(in *processor.Instance) *apiv1.Processor {
 	return &apiv1.Processor{
 		Id:        in.ID,
-		Name:      in.Name,
+		Name:      in.Type,
```
lovromazgon, line 33: Note: this field will change once we update to the new proto definition (this PR needs to be merged first so we can update the dependency and then change this field).

proto/api/v1/api.proto
```diff
@@ -111,7 +111,7 @@ message Processor {
 
   // -- immutable fields -------------------------------------------------------
 
-  string name = 4 [(google.api.field_behavior) = IMMUTABLE];
+  string type = 4 [(google.api.field_behavior) = IMMUTABLE];
```
lovromazgon, line 114: For future reference - we are changing the field name which will change the JSON field name and thus it's a breaking change. Old gRPC clients should still work though, because the binary representation of the message should be the same. We are aware of this and are ignoring the validation warning on purpose, the only affected client right now should be the UI and we will change it when we move to the new proto definition (that's not done in this PR).

pkg/web/api/processor_v1.go
```diff
@@ -34,7 +34,7 @@ type ProcessorOrchestrator interface {
 	// Get will return a single Interface or an error if it doesn't exist.
 	Get(ctx context.Context, id string) (*processor.Instance, error)
 	// Create will make a new Interface.
-	Create(ctx context.Context, name string, parent processor.Parent, cfg processor.Config) (*processor.Instance, error)
+	Create(ctx context.Context, procType string, parent processor.Parent, cfg processor.Config) (*processor.Instance, error)
```
lovromazgon, line 37: Note: this file and the accompanying test file will contain errors once we use the new proto definition because the field `Name` will be renamed to `Type` (just a note that this is expected and will be done in a followup PR).

lovromazgon (approved):
Great job 👍 let's merge this so that the new proto schema is uploaded to the BSR, then we can followup by updating the buf dependency and finalize the renaming of the field 🙂

PR #671, opened 2022-10-14
latest buf schema registry

### LINKED ISSUE #498: Rename Processor.Name to Processor.Type

maha-hajja:
### Description

after merging the proto file changes for renaming processors' name to type, a new schema for conduit was pushed to the Buf schema registry, and this PR uses the latest version.

Fixes #498 

### Quick checks:

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -31,7 +31,7 @@ require (
 	github.com/prometheus/client_model v0.2.0
 	github.com/prometheus/common v0.37.0
 	github.com/rs/zerolog v1.28.0
-	go.buf.build/conduitio/conduit/conduitio/conduit v1.1.2
+	go.buf.build/conduitio/conduit/conduitio/conduit v1.1.3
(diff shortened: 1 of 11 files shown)


lovromazgon:
Oh, I almost missed it - @maha-hajja can you please run `make generate` to update the OpenAPI definition? 🙏

lovromazgon (approved):
🚀 🙌  Great team work!

PR #673, opened 2022-10-14
Graceful pipeline stop fixes

### LINKED ISSUE #662: Pipeline stuck in running state with very fast source and a failing destination
### Bug description

This bug was originally reported by @lyuboxa. He reported a problem with a pipeline with:
* a Kafka source with a topic with a very large number of messages, and 
* a Kafka destination with a topic which actually doesn't exist.

The logs show:
```
2022-10-06T17:22:20+00:00 ERR error writing record error="failed to produce messages: [3] Unknown Topic Or Partition: the request is for a topic or partition that does not exist on this broker" connector_id=pipeline1:destination1 plugin_name=builtin:kafka plugin_type=destination record_position=74571ce2-52b4-436e-8a4d-4ed8c92906e7
```
The pipeline, however, is in a running state.

When trying to stop it, we get:
```
{
    "code": 13,
    "details": [],
    "message": "failed to stop pipeline: tried to inject control message but PubNode is not running"
}
```

This issue can be reproduced with a very fast generator source and a Kafka destination which is failing. In the example below, the failure comes from the destination topic not existing, and the Kafka broker not allowing auto-create of topics.

### Steps to reproduce

```yaml
---
version: 1.0
pipelines:
  pipeline1:
    status: running
    name: pipeline1
    description: 'Fast generator source, Kafka destination'
    connectors:
      source1:
        type: source
        plugin: builtin:generator
        name: generator-source
        settings:
          recordCount: "-1"
          readTime: "0ms"
          format.options: "id:int,name:string,company:string,trial:bool"
          format.type: "structured"
      destination1:
        type: destination
        plugin: "builtin:kafka"
        name: kafka-destination
        settings:
          servers: "localhost:9092"
          topic: "non-existing-topic"

```

### Version

v0.4.0-nightly.20221006

lovromazgon:
### Description

This PR fixes 3 separate bugs, all related to a graceful stop of a pipeline.

1. If a message did not have a nack handler, no other handler was called on that message on nack, causing the source node to wait forever. ([fix](https://github.com/ConduitIO/conduit/pull/673/files#diff-2a0555b9f26bdb43d1c85d4c428b347d3fa36ee7e233ab040e32900c7d5cc369R217) - call the handlers regardles)
2. `SourceNode` triggered `Stop` in the connector plugin in the same goroutine as it was fetching new messages. The plugin receives the call, cancels the routine that is producing messages and waits for it to stop running. But since messages can still be produced (e.g. caches can be drained) and `SourceNode` is not fetching the messages it caused a deadlock. ([fix](https://github.com/ConduitIO/conduit/pull/673/files#diff-bbe33d88f163d6ec8f9370f4de6c317d3f06925dd349c1d1f9c76d55536b18e0R206) - trigger `Stop` asynchronously and send a message to `SourceNode` with the last position)
    - While testing this out I found that there's still an issue left where a deadlock can occur and created a test for it ([TestSourceNode_StopWhileNextNodeIsStuck](https://github.com/ConduitIO/conduit/pull/673/files#diff-90ffc02f855999f7e0599bd120e73ccacb252fcc7a27acb6fba425fe08ca55b0R79)). The test is failing so it is skipped for now.
3. The destination acker node can stop because of an error, but that can cause a deadlock because there might still be acks left to process from records already sent to the connector. ([fix](https://github.com/ConduitIO/conduit/pull/673/files#diff-e1b969c4bfcf838dddc7eef325e03c213b1b6a2c0e6e8914121ed7684ce1684fR186-R195) - when tearing down the destination acker node it spins up a goroutine that keeps on fetching acks, essentially draining the acks stream).

Fixes #662

### Quick checks:


pkg/pipeline/stream/source.go
```diff
@@ -175,6 +175,22 @@ func (n *SourceNode) registerMetricStatusHandler(msg *Message) {
...
+	n.stopOnce.Do(func() {
+		// only execute stop once, more calls won't make a difference
+		err = n.stop(ctx, reason)
+		stopExecuted = true
+	})
+	if !stopExecuted {
+		n.logger.Warn(ctx).Msg("source connector stop already triggered, " +
+			"ignoring second stop request (if the pipeline is stuck, please " +
+			"report the issue to the Conduit team)")
+	}
```
hariso, line 189: Maybe we return this as an error, so it's part of the API response?

lovromazgon: I was thinking about it when writing this. The drawback is that only the first request to stop the pipeline would succeed, all calls after that would fail unless the pipeline is started again. The UI would show this error, that didn't seem right 🤔
I agree that returning the error would give the caller the info what's happening, although then it would be nice if the UI caught the error and displayed it more as an info/warning rather than a full blown error.

hariso: > although then it would be nice if the UI caught the error and displayed it more as an info/warning rather than a full blown error.

Agreed, it would be nice, and maybe not even too complex to be done in the front-end?

(But obviously, I don't consider this a blocker for the PR.)

pkg/pipeline/stream/source_test.go
```diff
@@ -0,0 +1,174 @@
...
+
+	nodeDone := make(chan struct{})
+	go func() {
+		defer close(nodeDone)
+		err := node.Run(ctx)
+		is.NoErr(err)
+	}()
+
+	for _, wantRecord := range wantRecords {
+		gotMsg := <-out
```
hariso, line 53: How likely is it that we wait for the message indefinitely (due to a bug in SourceNode)? I'm wondering if it would be an overkill to have a timeout here.

pkg/pipeline/stream/source_test.go
```diff
@@ -0,0 +1,174 @@
...
+	out := node.Pub()
+
+	nodeDone := make(chan struct{})
+	go func() {
+		defer close(nodeDone)
+		err := node.Run(ctx)
+		is.NoErr(err)
+	}()
+
+	// give the record a chance to read the record and start sending it to the
```
hariso, line 104: > give the record a chance to read the record
Probably a slip of the pen and should have been "give the **source** a chance to read the record"?

pkg/pipeline/stream/source_test.go
```diff
@@ -0,0 +1,174 @@
...
+	case <-nodeDone:
+		// all good
+	}
+}
+
+func TestSourceNode_StopWhileNextNodeIsStuck(t *testing.T) {
+	// A pipeline can't be stopped gracefully if the next node after a source
+	// node blocks forever (or for a long time), because we can't inject the
+	// stop message into the stream.
+	t.Skip("bug isn't fixed yet")
```
hariso, line 83: IIUC, this isn't something we want to tackle in this PR. If that is the case, do we have an issue for this?

hariso (approved):

pkg/pipeline/stream/source_test.go
lovromazgon: I took the opportunity to add utilities for receiving from a channel with a timeout and for waiting for a waitgroup with a timeout (remember this? https://github.com/ConduitIO/conduit-connector-sdk/commit/a9290b473be448836a4a0225afffa1215eda8307). Then I added the timeout here and simplified test code that was using the same functionality.

Scope creep? Nooo, just "leaving the campsite nicer than I found it" 😇

pkg/pipeline/stream/source_test.go
lovromazgon: Fixed!

pkg/pipeline/stream/source_test.go
```diff
@@ -0,0 +1,174 @@
...
+		is.Fail() // expected node to stop running
+	case <-nodeDone:
+		// all good
+	}
+}
+
+func TestSourceNode_StopWhileNextNodeIsStuck(t *testing.T) {
+	// A pipeline can't be stopped gracefully if the next node after a source
+	// node blocks forever (or for a long time), because we can't inject the
+	// stop message into the stream.
```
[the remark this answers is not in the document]
lovromazgon: Yeah that might make sense, although it doesn't completely solve the issue of stopping a pipeline with a stuck node. I feel like we need to tackle this separately and think about what behavior we want if this happens.

PR #684, opened 2022-10-19
fix bug: Pipeline is run before being deprovisioned

### LINKED ISSUE #672: Bug: Pipeline is run before being deprovisioned
### Bug description

When a pipeline is provisioned by a configuration file it is stored in the Conduit database. When Conduit restarts it first loads the pipeline from the database and starts it before loading the configuration files. This means that when the pipeline configuration file is removed (i.e. pipeline should be deprovisioned), the pipeline will still start for a brief moment before it's stopped and deprovisioned. Sometimes even an error is logged that the pipeline couldn't be stopped so the deprovisioning fails completely.

We should be able to fix this if we change the order of operations:
1. Load pipelines from the database without starting them
2. Load pipelines from config files and create/update/delete them
3. Start pipelines that need to be started

### Steps to reproduce

1. Run Conduit with a pipeline config file (easiest way is to use the example pipeline `./conduit -pipelines.path ./examples/pipelines`)
2. Stop conduit (ctrl+c)
3. Run Conduit without the config file (just `./conduit`)
4. See in the logs that the pipeline was started and right after it got stopped and deleted

### Version

v0.4.0-nightly.20221013 darwin/amd64

maha-hajja:
### Description

to fix this bug, pipeline service initializes the pipelines from store without running any of them, using the function `Init`. It marks the running pipelines with the status `systemStopped` so they can be identified after provisioning.

after provisioning is done, `InitStatus` function from pipeline service is called, to run the pipelines that had the running status in store.

Fixes #672 

### Quick checks:

### DIFF
--- a/pkg/pipeline/service.go
+++ b/pkg/pipeline/service.go
@@ -72,8 +72,8 @@ func (s *Service) Init(ctx context.Context) error {
 	return err
 }
 
-// InitStatus run pipelines that had the running state in store.
-func (s *Service) InitStatus(
+// Run runs pipelines that had the running state in store.
+func (s *Service) Run(
 	ctx context.Context,
 	connFetcher ConnectorFetcher,
 	procFetcher ProcessorFetcher,
--- a/pkg/pipeline/service.go
+++ b/pkg/pipeline/service.go
@@ -46,14 +46,9 @@ func NewService(logger log.CtxLogger, db database.DB) *Service {
 	}
 }
 
-// Init fetches instances from the store and starts pipelines that are supposed
-// to be running. Connectors and processors should be initialized before calling
-// this function.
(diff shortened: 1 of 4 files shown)


pkg/pipeline/service.go
```diff
@@ -62,15 +57,32 @@ func (s *Service) Init(
...
+
+	return err
+}
+
+// InitStatus run pipelines that had the running state in store.
+func (s *Service) InitStatus(
+	ctx context.Context,
+	connFetcher ConnectorFetcher,
+	procFetcher ProcessorFetcher,
+) error {
```
hariso, line 80: Maybe call it Run?

hariso (approved):

PR #690, opened 2022-10-25
docs: readme for proto files

lovromazgon:
### Description

Adds readme for proto files regarding the Conduit API.

### DIFF
--- a/proto/README.md
+++ b/proto/README.md
@@ -1,0 +1,32 @@
+# Conduit API Proto files
+
+This folder contains protobuf files that define the Conduit gRPC API and
+consequently also the HTTP API via
+the [grpc-gateway](https://github.com/grpc-ecosystem/grpc-gateway). The proto
+schema is uploaded to
+the [Buf schema registry](https://docs.buf.build/bsr/introduction) and can be
+found here: https://buf.build/conduitio/conduit
+
+Buf schema registry provides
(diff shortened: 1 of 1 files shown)


proto/README.md
```diff
@@ -0,0 +1,32 @@
...
+Because we use remote code generation provided by the Buf schema registry there
+is no locally generated code. When developing locally we don't want to push a
+new version of the proto files every time we make a change, that's why in that
+case we can switch to locally generated protobuf code.
+
+To switch to locally generated protobuf code follow the following steps:
+- run `cd proto && buf generate`
+- cd into the newly generated folder `proto/gen`
+- create a `go.mod` file by running `go mod init go.buf.build/conduitio/conduit/conduitio/conduit && go mod tidy`
+- cd into the root of the project and run `go mod edit -replace go.buf.build/conduitio/conduit/conduitio/conduit=./proto/gen && go mod tidy`
```
maha-hajja, line 32: should we add something about how to reverse this command? after the local development is done?

hariso, line 32: Nitpick: would be cool if we had this as a make target.

hariso (approved):

proto/README.md
lovromazgon: Agreed, I hope we can live with this piece of documentation for now though 🙃

hariso: We definitely can.:)

PR #691, opened 2022-10-25
test: make source node test more stable

lovromazgon:
### Description

There ~is~ was a [flaky test](https://github.com/ConduitIO/conduit/actions/runs/3320689900/jobs/5487359936#step:4:1046). The issue was that the source connector mock returned an error in `Read` after `Stop` was called. This does not happen with real connectors, they keep running until `Teardown` is called which actually closes the stream. This is now changed, it also changes the expectation for how many times `Read` is called on the mock, making it even more stable.

I ran the test in a loop 10.000 times, before the change it failed ~10 times, after the change no fails anymore.

### Quick checks:


hariso (approved):
:rocket:

Issue #17, opened 2021-05-03
Stream Inspector


neovintage:
The goal is to give folks the ability to see what's happening inside the stream. This is not like `tail`. This is more like `peek`. Pull out some information from the stream so that a developer can see the data and data types. 

- [x] #692 
- [ ] UI changes
- [x] #700 
- [x] #701 
- [x] #702

lyuboxa:
Inspector will be nice, but I will raise you a REPL for this, it will be magnificent.

neovintage:
@lyuboxa what would a repl look like? Would you have to create a "conduit cli" to make this possible?

lyuboxa:
I think any tooling that can be used to inspect data real time and does not involve explicit parsing will be useful. When @hariso and I were discussing this something like this came to mind:

```
$ conduit-cli
> pipelines
<... print list of pipelines>
> c  = pipeline["abc"].connector["source"]
> c.peek(1) 
-> <message in structured format>
> c.processors["anonymize"].Peek(1)
-> <message in structured format> 
```

I think an interactive REPL will be useful, a console or whatever you want to call it.

lovromazgon:
@lyuboxa A REPL doesn't sound very composable to me 😕 Why not rather provide a simple CLI command to get the records? This would let you compose it with other tools like `grep` & co. Example:

```
$ conduit-cli inspect id:of:my:component --limit 2
<message>
<message>
```

Sidenote: I think it's a bit off-topic to talk about a CLI/REPL in this issue, it sounds like a whole separate feature by itself.

lyuboxa:
> A REPL doesn't sound very composable to me .

Does it have to be? REPL are interactive tools for evaluation, not unix commands. Both options have their merits.
Former is more suitable for interactive work (which CLI is not tbh, but it does need a way to enter an interactive session). 
However, if you only want to dump data and move on, then this is likely not the most optimal solution.

> Sidenote: I think it's a bit off-topic to talk about a CLI/REPL in this issue, it sounds like a whole separate feature by itself.

Yup for sure.

hariso:
> > Sidenote: I think it's a bit off-topic to talk about a CLI/REPL in this issue, it sounds like a whole separate feature by itself.
> 
> Yup for sure.

Here's a separate issue for that: https://github.com/ConduitIO/conduit/issues/697.

hariso:
All the tasks are done, closing the issue.

PR #692, opened 2022-10-25
docs: Stream inspector design

hariso:
### Quick checks:


docs/design-documents/20221024-stream-inspector.md
```diff
@@ -0,0 +1,73 @@
...
+### Chosen implementation
+[grpc-websocket-proxy](https://github.com/tmc/grpc-websocket-proxy/) mention in option 1 is relatively popular and is
+open-source, so using it is no risk. Overall
+
+## Questions
+
+* Should we rename it to pipeline inspector instead? Pipeline is the term we use in the API, streams are used
+  internally.
+    * Cons: We cannot call it CSI (Conduit's Stream Inspector) anymore.
+* Is metadata needed (such as time the data was captured)?
```
lovromazgon, line 83: Hmm, I'd say we don't need it in the first iteration. It would be interesting to know the time (e.g. we could measure how long the record took to reach that point in the pipeline), but I don't think we need to make our lives harder because of this. If we go with the push solution the latency should _ideally_ be very low anyway so the client would see the record roughly at the same time when it passed through the pipeline.

Sidenote: if we still decide to add this info I advise against adding it in the record metadata, I think it's important to be able to inspect the record exactly as it is in the pipeline. So in that case we would probably need to wrap the record in some other structure (inspector record?) that contains its own metadata.

docs/design-documents/20221024-stream-inspector.md
```diff
@@ -0,0 +1,87 @@
...
+
+## Future work
+
+* Inspector REPL
```
lovromazgon, line 87: When I hear REPL I think of it being able to evaluate commands, what exactly do we mean by this? Are we talking about a CLI interface to inspect records?

docs/design-documents/20221024-stream-inspector.md
```diff
@@ -0,0 +1,87 @@
...
+grpc-gateway doesn't support WebSockets (see [this](https://github.com/grpc-ecosystem/grpc-gateway/issues/168)). There's 
+an open-source proxy for it though, available [here](https://github.com/tmc/grpc-websocket-proxy/). 
+
+### Implementation option 2: Pure gRPC
+Conduit provides a streaming endpoint. The stream is consumed as such, i.e. a gRPC endpoint. A JavaScript implementation
+of gRPC for browser clients exists (called [grpc-web](https://github.com/grpc/grpc-web)). While that would mean no 
+changes in Conduit itself, it would require a lot of changes in the UI.
+
+### Implementation option 3: Server-sent events 
+TBD
```
lovromazgon, line 70: Did you forget to delete this option or is there another one coming?

docs/design-documents/20221024-stream-inspector.md
```diff
@@ -0,0 +1,87 @@
...
+
+### Chosen implementation
+[grpc-websocket-proxy](https://github.com/tmc/grpc-websocket-proxy/) mention in option 1 is relatively popular and is
+open-source, so using it is no risk. The other option is much costlier.
+
+
+
+## Questions
```
lovromazgon, line 78: * Should there be a limit on how long a stream inspector can run?
* Should there be a limit on how many records a stream inspector can receive?

docs/design-documents/20221024-stream-inspector.md
```diff
@@ -0,0 +1,87 @@
...
+Three methods are required in total:
+1. One to inspect a connector (there's a single endpoint for source and destination connectors)
+2. One to inspect records coming into a processor
+3. One to inspect records coming out of a processor
+
+### Implementation option 1: WebSockets
+Conduit provides a streaming endpoint. The stream is exposed using the WebSockets API.
+
+grpc-gateway doesn't support WebSockets (see [this](https://github.com/grpc-ecosystem/grpc-gateway/issues/168)). There's 
+an open-source proxy for it though, available [here](https://github.com/tmc/grpc-websocket-proxy/).
```
lovromazgon, line 62: I'm curious, did you try out the websocket proxy (like a proof of concept) to know that this is a viable option?

docs/design-documents/20221024-stream-inspector.md
```diff
@@ -0,0 +1,87 @@
...
+   inspectable.
+    * For v1, it's enough that destinations are inspectable.
+2. A source inspector should show the records coming out of the respective source.
+3. A destination inspector should show the records coming into the respective destination.
+4. A processor has two inspectors: one for the incoming data and one for the resulting data (i.e. the transformed
+   records).
+5. The inspector should provide the data about the component being inspected, from the time the inspection started.
+   Historical data is not required.
+6. The inspector should have a minimal footprint on the resource usage (CPU, memory etc.)
+7. It should be possible to specify which data is shown.
```
lovromazgon, line 19: What options do you have in mind here?

docs/design-documents/20221024-stream-inspector.md
```diff
@@ -0,0 +1,87 @@
...
+
+1. All pipeline components (accessible "from the outside", i.e. sources, processors and destinations) should be
+   inspectable.
+    * For v1, it's enough that destinations are inspectable.
+2. A source inspector should show the records coming out of the respective source.
+3. A destination inspector should show the records coming into the respective destination.
+4. A processor has two inspectors: one for the incoming data and one for the resulting data (i.e. the transformed
+   records).
+5. The inspector should provide the data about the component being inspected, from the time the inspection started.
+   Historical data is not required.
```
lovromazgon, line 17: Is this what you mean? I want to make sure I understand this point.
```suggestion
5. The inspector should provide records flowing into/out of the component being inspected from when the inspection started.
   Historical data is not required.
```

lovromazgon (commented):
I think we should dive just a bit deeper into the implementation options, specifically we should explore how we can extract records while they are flowing through the pipeline. I see at least two options:
- somehow wrap the connector [`Read`](https://github.com/ConduitIO/conduit/blob/d19379efc04d20d12ab9c80df82a29fcef7e8afd/pkg/connector/connector.go#L91) and [`Write`](https://github.com/ConduitIO/conduit/blob/d19379efc04d20d12ab9c80df82a29fcef7e8afd/pkg/connector/connector.go#L116) functions as well as the processor [`Process`](https://github.com/ConduitIO/conduit/blob/d19379efc04d20d12ab9c80df82a29fcef7e8afd/pkg/processor/instance.go#L48) function
- or give the [`PubNode`](https://github.com/ConduitIO/conduit/blob/d19379efc04d20d12ab9c80df82a29fcef7e8afd/pkg/pipeline/stream/node.go#L68) and [`SubNode`](https://github.com/ConduitIO/conduit/blob/d19379efc04d20d12ab9c80df82a29fcef7e8afd/pkg/pipeline/stream/node.go#L80) the ability to somehow attach to it and inspect incoming/outgoing records.

The second option is more generic (we write it once and can attach to any node), but we currently can't address running nodes, they are spawned and run but not stored anywhere.

Maybe there's an even better solution, but in any case, I think it's worth discussing here.

docs/design-documents/20221024-stream-inspector.md
hariso: Lyubo provided an example here: https://github.com/ConduitIO/conduit/issues/17#issuecomment-1292440999

docs/design-documents/20221024-stream-inspector.md
hariso: At the moment, I'm mainly interested in seeing if there's enough interest in this. This way or the other, it should be possible, because you can always (and pretty easily) extract the data you want from the API response.

docs/design-documents/20221024-stream-inspector.md
hariso: Good ones, I've added them.

hariso:
> I think we should dive just a bit deeper into the implementation options, specifically we should explore how we can extract records while they are flowing through the pipeline. I see at least two options:

Good callouts! I didn't mention those for two reasons: I wanted to make sure we're all on the same page w/r to the requirements. And I felt like it's more of a thing for the PRs and so I left the implementation in the design doc high-level. But I can mention it in this PR as well.

> but we currently can't address running nodes, they are spawned and run but not stored anywhere.

How about [this](https://github.com/ConduitIO/conduit/blob/main/pkg/pipeline/instance.go#L60) though?

docs/design-documents/20221024-stream-inspector.md
```diff
@@ -0,0 +1,90 @@
...
+
+## Questions
+
+* Should we rename it to pipeline inspector instead? Pipeline is the term we use in the API, streams are used
+  internally.
+    * Cons: We cannot call it CSI (Conduit's Stream Inspector) anymore.
+* Is metadata needed (such as time the data was captured)?
+* Should there be a limit on how long a stream inspector can run?
+* Should there be a limit on how many records a stream inspector can receive?
+* Are we interested in more than the records? Is there some other data we'd like to see (now or in future)?
```
hariso, line 86: I believe that the records are the main thing we're interested in, so it doesn't look like we should be concerned with other types of data. At the moment, I can't think of something really useful, except the records themselves.

lovromazgon: Yeah I think that's the main point, we want to be able to inspect the records flowing through the pipeline, no need to design for anything more.

docs/design-documents/20221024-stream-inspector.md
```diff
@@ -0,0 +1,90 @@
+# Stream Inspector
```
lovromazgon, line 1: (Adding a comment on this line just so we can have a discussion, otherwise it's hard to follow)

> > but we currently can't address running nodes, they are spawned and run but not stored anywhere.
>
> How about [this](https://github.com/ConduitIO/conduit/blob/main/pkg/pipeline/instance.go#L60) though?

Oh, I forgot we had that 🙂 That pretty much solves the problem of addressing a node, the nodes that represent a certain entity (source, destination, processor) have the same ID as that entity, so it should be easy to fetch the right node. The other side of the problem is how we attach to the stream (and detach afterwards), but that's something we can figure out when we start implementing it 👍

docs/design-documents/20221024-stream-inspector.md
lovromazgon: Any opinions there? I'm thinking if there is a reason we would want to add limits? 🤔 I mean it probably wouldn't be advisable to have an inspector running forever, so we could prevent users from mistakenly having inspectors open. If some inspector would crash and we wouldn't notice it this would also make sure the inspector is closed after some time. On the other hand, having some arbitrary limit might make our user's life miserable if they are waiting for a certain message to arrive and it stops just when it flows by.

docs/design-documents/20221024-stream-inspector.md
lovromazgon: We can take inspiration from [github.com/eapache/channels#Tee](https://pkg.go.dev/github.com/eapache/channels#Tee) or [`io.TeeReader`](https://pkg.go.dev/io#TeeReader).

lyuboxa: Having the ability to tee any output, without interruption, can be pretty nice.

docs/design-documents/20221024-stream-inspector.md
lyuboxa: Maybe? But should that be configurable, I can see users attaching an inspector and running it over a day to determine patterns or something else. I have done this with tcpdump. However this will exact some cost to the service.

docs/design-documents/20221024-stream-inspector.md
lyuboxa: Hm some metadata that maybe, could be useful is to show the next node destination? This will help hint at directionality of the message.

docs/design-documents/20221024-stream-inspector.md
hariso: I believe it depends on the implementation, but generally, every record sent through the inspector can be "cleaned up", so I don't think we'll have some significant cost to the service. And, at least in the WebSocket PoC, as soon as the request is done (e.g. client crashes), the inspection ends. It might make sense limiting the number of inspections for a pipeline or node, but it might be too early for that too.

So, at least for now, I'm inclined towards no limits at all.

docs/design-documents/20221024-stream-inspector.md
hariso: > The other side of the problem is how we attach to the stream (and detach afterwards), but that's something we can figure out when we start implementing it +1

When you say "stream" what exactly do you mean?

docs/design-documents/20221024-stream-inspector.md
hariso: > Sidenote: if we still decide to add this info I advise against adding it in the record metadata, I think it's important to be able to inspect the record exactly as it is in the pipeline. So in that case we would probably need to wrap the record in some other structure (inspector record?) that contains its own metadata.

I'm thinking the same. In my PoC too I wrap the record in a response struct, so it's easier to change it if needed, since I think they'll change for different reasons.

docs/design-documents/20221024-stream-inspector.md
hariso: It looks like this question is a duplicate of the one above "is metadata needed?". 

Regarding directionality of the message: to my knowledge, a message is either dropped (in a processor) or it goes to the next node as specified in the pipeline topology (e.g. another processor, a destination with or without processors etc.). Unless I'm maybe missing something, I'm not sure how much additional value would we actually get from that. Maybe we could do it only to make information more easily accessible (i.e. if it's a complex pipeline, you don't have to go back to the UI to check the topology).

But that reminds of something I think someone already mentioned, and that is adding the information about the nodes through which a record passed, before coming to the inspection point. That might be a useful thing maybe? I believe it would always boil down to the source node.

docs/design-documents/20221024-stream-inspector.md
lovromazgon: I meant how we listen to messages flowing through a channel between two nodes and how this affects the pipeline performance.

That opens another question - what happens if the pipeline inspector is slower than the rate of messages that flow through the pipeline? Does it cause the pipeline to slow down (i.e. it applies backpressure)? Or should messages just be dropped if the inspector can't handle them in time? It sounds to me like the inspector should guarantee that it will return all messages that flow by in the correct order, but that necessarily means the pipeline performance can be impacted if the inspector is too slow.

docs/design-documents/20221024-stream-inspector.md
lovromazgon: I'll argue we should leave this out of the scope of this issue. It sounds like it's a matter of how the data is represented to the user, which should be handled on the client side IMO. The only use case where I see value in this is if the payload is soo huge you wouldn't want to send it all to the client, but that's probably an uncommon use case. I'd keep it simple for now.

docs/design-documents/20221024-stream-inspector.md
lovromazgon: I have nothing to add, no limits sounds good for now 👍

docs/design-documents/20221024-stream-inspector.md
hariso: Agreed :+1:

docs/design-documents/20221024-stream-inspector.md
lovromazgon: A separate issue was created for this https://github.com/ConduitIO/conduit/issues/697, let's continue the discussion there.

docs/design-documents/20221024-stream-inspector.md
hariso: You're right. I'll then resolve this conversation, since we can continue in the comment you mention.:)

docs/design-documents/20221024-stream-inspector.md
lovromazgon: > I believe it would always boil down to the source node.

That's right, if you know the source node you know the path the message has taken to the inspection node. You'd have to know the topology of the pipeline, but that's static and doesn't change while a pipeline is running.

This sounds valuable, I propose we automatically add a metadata field to each record with the source ID or name.

Sidenote: we already have similar fields prepared for metadata about the _connector plugin_ ([here](https://github.com/ConduitIO/conduit-connector-protocol/blob/4b89a6c94401924eb11a2554298be0e5ab3b9b88/proto/connector/v1/connector.proto#L8-L9)) but IIRC they are not used yet.

docs/design-documents/20221024-stream-inspector.md
maha-hajja: I agree with no limits, for me using an inspector means it's going to keep inspecting until I stop it or delete it.

docs/design-documents/20221024-stream-inspector.md
hariso: Interesting question! :+1: 

I think we need to explicitly mention that the inspector is expected to have an insignificant impact on the pipeline, if at all. There's already a requirement that the inspector has insignificant resource usage. In other words, we make our best efforts to make inspector lightweight. If an inspection is slow, it's the client's fault.:) With that, what we could is to keep a copy of the records, only for the inspector. 

> what happens if the pipeline inspector is slower than the rate of messages that flow through the pipeline?

If the pipeline inspector or the client is slow for whatever reason, and we start dropping messages, that might make someone think that messages are being lost for real. So we should either send a warning to the client that messages are going to be dropped, or we close the inspection.

docs/design-documents/20221024-stream-inspector.md
```diff
@@ -0,0 +1,73 @@
...
+changes in Conduit itself, it would require a lot of changes in the UI.
+
+### Chosen implementation
+[grpc-websocket-proxy](https://github.com/tmc/grpc-websocket-proxy/) mention in option 1 is relatively popular and is
+open-source, so using it is no risk. Overall
+
+## Questions
+
+* Should we rename it to pipeline inspector instead? Pipeline is the term we use in the API, streams are used
+  internally.
```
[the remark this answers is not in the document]
maha-hajja: +1

docs/design-documents/20221024-stream-inspector.md
hariso: That metadata field is really convenient. :+1:

docs/design-documents/20221024-stream-inspector.md
lovromazgon: > With that, what we could is to keep a copy of the records, only for the inspector.

Having a buffer might help, although it doesn't prevent the inspector from impacting the performance of a pipeline (the buffer can't be unbounded).

> If the pipeline inspector or the client is slow for whatever reason, and we start dropping messages, that might make someone think that messages are being lost for real.

Good point. The more I think about dropping messages the more it's clear we should not do that. I (as a user) would rather have a slower pipeline and know I see everything that flows by than get an incomplete stream of messages.

> we make our best efforts to make inspector lightweight. If an inspection is slow, it's the client's fault.

That sounds good to me.

> we need to explicitly mention that the inspector is expected to have an insignificant impact on the pipeline

That doesn't sound good to me 🙃 I mean we can say that _generally_ this is the case, but we need to be transparent that it can impact the pipeline performance if the pipeline inspector client doesn't fetch messages fast enough, in that case we will apply backpressure.

hariso: > Having a buffer might help, although it doesn't prevent the inspector from impacting the performance of a pipeline **(the buffer can't be unbounded)**.

Why not though?

> that it can impact the pipeline performance if the pipeline inspector client doesn't fetch messages fast enough, in that case we will apply backpressure.

IMO, troubleshooting a problem (i.e. inspecting a pipeline) shouldn't introduce another problem (pipeline getting slower). The pipeline and the inspector should be as isolated one from another as possible, so that we're as confident as possible that bugs in one won't cause bugs in the other. For me personally, it would give much more freedom to actually inspect and not be afraid I'll make things worse if I need to look around or do some tests.

docs/design-documents/20221024-stream-inspector.md
```diff
@@ -0,0 +1,105 @@
...
+
+From what we know so far, the push based approach has more advantages and is easier to work with, so that's the 
+approach chosen here. Concrete implementation options are discussed below.
+
+For context: gRPC is the main API for Conduit. The HTTP API is generated using [grpc-gateway](https://github.com/grpc-ecosystem/grpc-gateway).
+
+### API
+Inspectable pipeline components themselves should expose an `Inspect` method, for example
+```go
+func(s Source) Inspect() chan Record
```
hariso, line 52: I'll need to think more about this, as it won't work for processors (incoming + outcoming records can be inspected).

docs/design-documents/20221024-stream-inspector.md
```diff
@@ -0,0 +1,110 @@
...
+
+The API layer will use this method to get the records it needs to stream to the client which requested an inspection.
+
+For any given pipeline component, only one gRPC/HTTP API method is required, one which starts an inspection (i.e. 
+sending of data to a client).
+
+Three methods are required in total:
+1. One to inspect a connector (there's a single endpoint for source and destination connectors)
+2. One to inspect records coming into a processor
+3. One to inspect records coming out of a processor
```
lovromazgon, line 68: I'll supply another option - we could have one endpoint (e.g. `/pipeline/{id}/inspect`) that takes arguments describing which entity to inspect (e.g. `{"entity":"processor","id":"my-processor"}`). Didn't think through if it's better or not 🙂

docs/design-documents/20221024-stream-inspector.md
hariso: > Physics? smile We can potentially fill up the memory, possibly causing the system to use swap space. That will cause a performance impact far greater than just the one pipeline. I'd rather try to restrict the possible impact.

I'm looking back at your original message and am surprised to see "unbounded", I was thinking: why prevents us from limiting the buffer?:)

> after the buffer is full it will start to throttle the speed of the pipeline though.

Personally, I wouldn't like to see that as someone inspecting a pipeline. A hanging UI/CLI, an incautious developer, a network latency etc. could cause the inspection to be slow, but I'd never expect something like that to throttle a pipeline. The more connected the pipeline and an inspector are, the higher are the changes for bugs and performance degradations.

lyuboxa: I want to circle back at the inspector being non-blocking only, I think inserting an inspector which allows for:
1. see data
2. data ok/not-ok -> ack/nack from the inspector

Could be useful in some interesting scenarios where but I think it will be less useful if we cannot put the data back.

docs/design-documents/20221024-stream-inspector.md
hariso: I like the idea! It would help with the direction, and IMO is better than having different endpoints for different directions.

When it comes to having a single endpoint for all: I'm viewing the connectors and processors as the actual resources, and "inspect" as the action taken on them. So at this moment, I find it a bit more natural to have `/connectors/{id}/inspect` then `/pipeline/{id}/inspect`.

lovromazgon: You're right, endpoints on the entities probably make more sense. My design likely doesn't even work because the request opens a websocket, so it needs to be a `GET` request.

We need to see if we can send the direction as a URL parameter and if that can be retrieved in the gRPC endpoint. Otherwise we'll have to resort to another option (Header? Separate endpoints? Initial message after the WS is opened that specifies the direction?). No need to figure this out here, it can be left for the implementation if you ask me ;)

docs/design-documents/20221024-stream-inspector.md
lovromazgon: @lyuboxa that's what I was afraid of 😆 We need to define the purpose of the inspector - I see it as a tool to inspect and debug records, not more, not less. Not for filtering or transforming records - that's a processor. It's also not a tool for consuming records and writing them to some resource - that's a destination. People can abuse the tool, but we should not design for that.

What you described would be a pretty interesting processor though, one where a human would be involved to confirm or reject a record. Could be used to do things like fraud detection 🙂

lyuboxa: > What you described would be a pretty interesting processor though, one where a human would be involved to confirm or reject a record. Could be used to do things like fraud detection slightly_smiling_face

Hmm you might be onto something, high latency human processor ;-)

hariso: @lyuboxa What would be those scenarios (manually acking/nacking messages and inserting data into a pipeline)?

docs/design-documents/20221024-stream-inspector.md
hariso: I believe the need for a wrapped struct depends on the need to have any data besides records or record metadata which we can't/shouldn't put into the "regular" record metadata (i.e. the OpenCDC metadata map). 

On the other hand, besides the UI (which we are developing ourselves), there's no other integration which would depend on the inspector, so I think we're pretty free to change the API in future if we need to.

docs/design-documents/20221024-stream-inspector.md
hariso: Here's the conclusion from today's team sync: https://github.com/ConduitIO/conduit/pull/692/files#diff-88c2b66f0b7457fc1c81362cacfa6ca628eb888f2326028c0abd9d5f538a79d4R123-R131

docs/design-documents/20221024-stream-inspector.md
```diff
@@ -0,0 +1,143 @@
+# Stream Inspector
+
```
hariso, line 2: Need to add a note that some inspector data may be dropped (e.g. if inspector is too slow).

lovromazgon (approved):
🚀

Issue #680, opened 2022-10-18
Built-in Config Validation 


uchennakevinm1:
issues needed to implement builtin config validations:

- [x] #720  
- [x] #694 
- [x] #695 
- [x] #746
- [x] #696

lovromazgon:
Related to parsing and validating configs I see these points:
* Validate constraints before the config reaches the connector or supply utilities to validate them in the connector.
  * This is what the issue mentions - validating and parsing.
  * Figure out if we should do this in Conduit or in the connector SDK. Doing it in Conduit means we could add or fix validations without changing the connector. On the other hand, the connector developer can't do anything if a validation is not doing what they want except removing the validation from the specs. Providing a utility function for validating the config in the SDK would give the developer more flexibility (maybe the only want to automatically validate a part of the config and do the other part manually), also the utility could aid the developer in parsing the config into a struct (we can take ideas from [this](https://pkg.go.dev/encore.dev/config#Load)). Having this in the SDK could potentially make it possible for a developer to specify a custom validation.
* Allow connector developers to specify validations/constraints on parameters (Conduit already kinda supports this but it's not exposed in the SDK).
* Supply a utility function that can be used in `Parameters` to build the parameters map out of a config struct (if it can't be done in runtime we might want to supply a generator).

hariso:
Closing this as all tasks are done.

PR #698, opened 2022-10-28
design doc for builtin config validation

maha-hajja:
Related to #680 

### Quick checks:


docs/design-documents/20221027-builtin-config-validation.md
```diff
@@ -0,0 +1,144 @@
...
+          Type: sdk.ValidationGreaterThan,
+          Value: 0,
+        },
+      }
+    }
+  }
+```
+
+2. Provide a function that takes the parameters' validations and validates them in the config function.
+3. Generate connector configurations from a Go struct, which will give the ability to generate the connector's
```
hariso, line 134: Do we plan to work on parsing? The original issue mentions validation only, but I see some comments mentioning parsing too. I think it would be good to have an SDK utility which would take a map, validate it (using the specs above or the tags here), and then validate, and if valid, parse it.

docs/design-documents/20221027-builtin-config-validation.md
```diff
@@ -0,0 +1,144 @@
...
+the SDK wouldn't cover. So we decided on implementing validating from the SDK side.
+
+**Q**: Should generating the Configurations from a Go struct be in the scope of this feature?
+
+**A**: We decided to add the generation as a subtask for now, and see how much time left we have for it. We agreed that this
+would be super helpful for developers and easier to use, and will continue adding to the feature in the future.
+
+## Implementation
+
+Implementing this feature consists of three main steps:
```
lovromazgon, line 110: I think there's another step (well at least I'd mark it as a separate step) - we need to make sure the connector protocol is adjusted and that the validations for parameters are added there. I'd also explicitly state that the change needs to be done in a backwards compatible way.

docs/design-documents/20221027-builtin-config-validation.md
lovromazgon: Currently step 2 provides a function for validation, we might want to have another step that provides a function for parsing into a struct. Arguably we could also provide a utility method that does both at the same time.

One could also argue that providing the validation utility is enough because the user can already use something like [`mitchellh/mapstructure`](https://github.com/mitchellh/mapstructure) for parsing the config into a struct.

docs/design-documents/20221027-builtin-config-validation.md
```diff
@@ -0,0 +1,144 @@
...
+      Type: sdk.ParameterTypeNumber,
+      Validations: []sdk.Validation{
+        sdk.Validation{
+          Type: sdk.ValidationRequired,
+        },
+        sdk.Validation{
+          Type: sdk.ValidationGreaterThan,
+          Value: 0,
+        },
+      }
```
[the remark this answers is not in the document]
maha-hajja: Yes this looks much neater! we're still going to have to convert this to the first format, because that's the one needed to match the proto message design, which will happen on the SDK and the developer won't have to worry about it.

docs/design-documents/20221027-builtin-config-validation.md
maha-hajja: I didn't go into much details to explicitly mention the connector protocol, the only change I see would be adding the `Validation` as a struct, and then adding `validations` slice into the parameter. did I miss something?
and agreed, will add this as a separate step!

docs/design-documents/20221027-builtin-config-validation.md
hariso: Well, that's a nice little lib, I've starred it.:) I'd usually say it would be nice to have it in the SDK too, but since it's that simple with the lib, we're not making it much simpler for the developers.

docs/design-documents/20221027-builtin-config-validation.md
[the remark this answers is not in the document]
hariso: I think I'm missing something... I read in this doc that the validations will be done in the SDK. And the proto files mean we're going to send them to Conduit. Why though?

docs/design-documents/20221027-builtin-config-validation.md
lovromazgon: Well, it will affect the connector protocol, the SDK as well as Conduit. It's a decent amount of work, worth to make it a separate step IMO 🙂 Also the backwards compatibility part will mean we need to parse the old `required` field and convert it into a validation, that can be part of this. It's probably the first thing that will need to happen so we have the validation structs we need in the SDK to populate the parameter specs.

docs/design-documents/20221027-builtin-config-validation.md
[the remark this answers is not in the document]
lovromazgon: We're sending them over so the UI can do the validations and provide a better UX.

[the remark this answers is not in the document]
hariso: So if a configuration has a combination of built-in and custom validations, the UI would be doing the built-in validations, and then the custom validations will be done in the connector itself?

docs/design-documents/20221027-builtin-config-validation.md
```diff
@@ -0,0 +1,144 @@
...
+        // Parameter must be greater than {value}.
+        TYPE_GREATER_THAN = 2;
+        // Parameter must be less than {value}.
+        TYPE_LESS_THAN = 3;
+        // Parameter must be included in the comma separated list {value}.
+        TYPE_INCLUSION = 4;
+        // Parameter must not be included in the comma separated list {value}.
+        TYPE_EXCLUSION = 5;
+        // Parameter must match the regex {value}.
+        TYPE_REGEX = 6;
```
hariso, line 41: Is this the list of validations we plan to support in the first step?

maha-hajja: yes, first step will allow the developer to add these validations to the params, second step will be the one validating

docs/design-documents/20221027-builtin-config-validation.md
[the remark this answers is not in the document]
maha-hajja: yes! If we see that this shouldn't be the case, then I'm thinking of one simple solution for it, the UI could also use the `validate` endpoint for Conduit, since it calls the `Configure` method from the connector.
But in general, I'm on the fence about whether we should validate both the builtin and the custom validations or not. WDYT?

docs/design-documents/20221027-builtin-config-validation.md
hariso: I didn't use the above lib myself, but it looks like parsing is as simple as writing `mapstructure.Decode()`. With that, I don't think we can make it easier for the developers.

docs/design-documents/20221027-builtin-config-validation.md
maha-hajja: @lovromazgon updated

docs/design-documents/20221027-builtin-config-validation.md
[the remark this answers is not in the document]
lovromazgon: Some thoughts on hitting `validate` after each config value is entered:
- In case of a standalone plugin, the `validate` endpoint will start and stop plugin each time, it's probably not the fastest endpoint (might not be an issue, but we should check how long it takes on average, just to be sure).
- The returned errors need to allow the UI to figure out which field is still wrong (we don't want to create generic pop-ups for these validations, right?).
  - We would also need to be smart about which errors to ignore, otherwise all fields would turn red once the user switches away from the first field.
- The `validate` endpoint will execute the standard validations as well. Do we even need to tell the UI about the standard validations at that point?

docs/design-documents/20221027-builtin-config-validation.md
lovromazgon: After thinking about this, it would be a nice UX if I could do something like this:
```
type Config struct {
    GreaterThan2 int `validate:"required,gt=2"`
}

func (s *Source) Configure(ctx context.Context, raw map[string]string) error {
    var cfg Config
    err := sdk.ParseConfig(raw, &cfg)
    if err != nil {
        return err
    }
    // I have a validated and parsed cfg at this point
    // Do custom validations here if needed
    // ...
}
```

Having a function that parses and validates could make it simpler for our developers (_in addition to `sdk.ValidateConfig`_, which only does the validation part, in case people want to manually parse it).

hariso: @lovromazgon agreed that it would be nice to have a function that would validate+parse.

docs/design-documents/20221027-builtin-config-validation.md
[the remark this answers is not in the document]
maha-hajja: discussed during conduit sync 22/11/2022, the UI will validate the builtlin validations for each configuration while the user is typing (validations are found in the same struct that has the list of the fields). When the user submits the form, the UI will try and create the connector (which will validate both the builtin and the custom validation), and will show an error for the user if it occurs. @jmar910

docs/design-documents/20221027-builtin-config-validation.md
```diff
@@ -0,0 +1,152 @@
...
+ 
+2. Adjust the connector SDK to give developers the ability to specify validations needed for each parameter. (manually)
+   
+Params should look something like:
+
+```go
+SourceParams: []sdk.Parameter{
+    {
+      Name: "param",
+      Type: sdk.ParameterTypeNumber,
```
hariso, line 132: This is going to change (because we have separate types for ints and floats)?

docs/design-documents/20221027-builtin-config-validation.md
```diff
@@ -0,0 +1,152 @@
...
+      string value = 2;
+    }
+
+    // Type shows the parameter type.
+    enum Type {
+      TYPE_UNSPECIFIED = 0;
+      // Parameter is a string.
+      TYPE_STRING = 1;
+      // Parameter is a number.
+      TYPE_NUMBER = 2;
```
hariso, line 56: Should this be two types now (ints and floats)?

hariso (approved):
LGTM! I left two questions, minor things.

docs/design-documents/20221027-builtin-config-validation.md
maha-hajja: it is yes, thought this design doc was before the change and doesn't have to be updated. But it's already there so why not.

Issue #681, opened 2022-10-18
Dead letter queue


uchennakevinm1:
Message rerouted to another pipeline.

Write to stdout if all paths fail. 

DevEx:

1. Error happens in main pipeline
2. If dead letter queue hasn’t been been configured, write to standard out
3. if DLQ has been configured, then messages get push to that pipeline.
4. A 3rd pipeline would have to be created to then reprocess whats in DLQ.

Error rate metrics will need to be exposed. Scope creep: acting on a certain error rate percentage to stop pipelines.

Subtasks:
- [x] https://github.com/ConduitIO/conduit/issues/712
- [x] https://github.com/ConduitIO/conduit/issues/713
- [x] https://github.com/ConduitIO/conduit/issues/714
- [x] https://github.com/ConduitIO/conduit/issues/716
- [x] https://github.com/ConduitIO/conduit/issues/715
- [x] https://github.com/ConduitIO/conduit/issues/721

Future work:
- https://github.com/ConduitIO/conduit/issues/717

PR #699, opened 2022-10-28
docs: Design dead-letter-queue

lovromazgon:
### Description

Design document for https://github.com/ConduitIO/conduit/issues/681.

### DIFF
--- a/docs/design-documents/20221028-dead-letter-queue.md
+++ b/docs/design-documents/20221028-dead-letter-queue.md
@@ -103,12 +103,10 @@ the introduction of DLQs - the first nacked message would stop the pipeline.
 
 ### API
 
-The DLQ should be defined as a property of
-the [`Pipeline.Config`](https://github.com/ConduitIO/conduit/blob/d19379efc04d20d12ab9c80df82a29fcef7e8afd/proto/api/v1/api.proto#L28)
-entity. It's a config option that can be updated even after the pipeline is
-created. We need to provide a way to choose a DLQ plugin and configure it.
+We need to provide a way to get the current DLQ configuration and to update it.
 
-One option to achieve this is to provide these fields:
+One option to achieve this is to add two methods to the `PipelineService` and
+map them to HTTP endpoints:
 
 ```protobuf
 message DLQ {
@@ -126,6 +124,40 @@ message DLQ {
   // cause the pipeline to stop
   // default = 1
   uint64 window_nack_threshold = 4;
+}
+
+service PipelineService {
+  rpc GetDLQ(GetDLQRequest) returns (GetDLQResponse) {
+    option (google.api.http) = {
+      get: "/v1/pipelines/{id}/dlq"
+      response_body: "dlq"
+    };
+  };
+
+  rpc UpdateDLQ(UpdateDLQRequest) returns (UpdateDLQResponse) {
+    option (google.api.http) = {
+      put: "/v1/pipelines/{id}/dlq"
+      body: "dlq"
+      response_body: "dlq"
+    };
+  };
+}
+
+message GetDLQRequest {
+  string id = 1;
+}
+
+message GetDLQResponse {
+  DLQ dlq = 1;
+}
+
+message UpdateDLQRequest {
+  string id = 1;
+  DLQ dlq = 2;
+}
+
+message UpdateDLQResponse {
+  DLQ dlq = 1;
 }
 ```
 
--- a/docs/design-documents/20221028-dead-letter-queue.md
+++ b/docs/design-documents/20221028-dead-letter-queue.md
@@ -43,6 +43,8 @@ The design for the DLQ should support the following:
 
 - **Should we allow the user to attach custom metadata to messages pushed to
   the DLQ?**
+  - We decided to add this as future work in the form of allowing users to
+    attach processors to a DLQ.
 
 ### Background
 
@@ -182,3 +184,5 @@ pipelines:
   error. We could give our users the option to choose what happens then - do
   they really want to stop the pipeline or ignore the error and carry on? Maybe
   even provide a fallback DLQ.
+- Since a DLQ is "just a connector" we could add the possibility of attaching
+  processors to it.
--- a/docs/design-documents/20221028-dead-letter-queue.md
+++ b/docs/design-documents/20221028-dead-letter-queue.md
@@ -92,14 +92,14 @@ the plugin name and its settings.
 
 The user can also configure a stop window and stop threshold.
 
-- `stop_window` defines how many last acks/nacks are monitored in the window
+- `window_size` defines how many last acks/nacks are monitored in the window
   that controls if the pipeline should stop. 0 disables the window.
-- `stop_threshold` defines the number of nacks in the window that would cause
-  the pipeline to stop.
+- `window_nack_threshold` defines the number of nacks in the window that would
+  cause the pipeline to stop.
 
-The default DLQ plugin should be `builtin:log` with a `stop_window` of 1 and a
-`stop_threshold` of 1. This would result in the same behavior as before the
-introduction of DLQs - the first nacked message would stop the pipeline.
+The default DLQ plugin should be `builtin:log` with a `window_size` of 1 and a
+`window_nack_threshold` of 1. This would result in the same behavior as before
+the introduction of DLQs - the first nacked message would stop the pipeline.
 
 ### API
 
@@ -118,14 +118,14 @@ message DLQ {
   // settings are the plugin settings
   map<string, string> settings = 2;
 
-  // stop_window defines how many last acks/nacks are monitored in the window
+  // window_size defines how many last acks/nacks are monitored in the window
   // that controls if the pipeline should stop (0 disables the window)
   // default = 1
-  uint64 stop_window = 3;
-  // stop_threshold defines the number of nacks in the window that would cause
-  // the pipeline to stop
+  uint64 window_size = 3;
+  // window_nack_threshold defines the number of nacks in the window that would
+  // cause the pipeline to stop
   // default = 1
-  uint64 stop_threshold = 4;
+  uint64 window_nack_threshold = 4;
 }
 ```
 
@@ -141,7 +141,7 @@ pipelines:
       [...]
     dlq:
       # disable stop window
-      stopWindow: 0
+      windowSize: 0
 
       # the next 3 lines explicitly define the log plugin
       # removing this wouldn't change the behavior, it's the default DLQ config
@@ -161,18 +161,22 @@ pipelines:
       [...]
     dlq:
       # stop when 30 of the last 100 messages are nacked
-      stopWindow: 100
-      stopThreshold: 30
-      
+      windowSize: 100
+      windowNackThreshold: 30
+
       # define file plugin used for DLQ messages
       plugin: builtin:file
       settings:
         path: ./dlq-example.dlq
 ```
 
-### UI
+### Metrics
 
-?
+We need to add metrics for monitoring the DLQ. We intend to add similar metrics
+to the ones we have for connectors:
+
+- `conduit_dlq_bytes`
+- `conduit_dlq_execution_duration_seconds`
 
 ## Future work
 
@@ -184,5 +188,5 @@ pipelines:
   error. We could give our users the option to choose what happens then - do
   they really want to stop the pipeline or ignore the error and carry on? Maybe
   even provide a fallback DLQ.
-- Since a DLQ is "just a connector" we could add the possibility of attaching
-  processors to it.
+- The DLQ will currently only be configurable through the API or pipeline
+  config file, adding a UI for this feature will be done in the future.
--- a/docs/design-documents/20221028-dead-letter-queue.md
+++ b/docs/design-documents/20221028-dead-letter-queue.md
@@ -29,6 +29,8 @@ The design for the DLQ should support the following:
   connector should be preserved, even if messages are nacked in a different
   order (this is already assured using
   a [semaphore](https://github.com/ConduitIO/conduit/blob/b6325584d51ec64bc9086faddcdaa2788a5dfa8f/pkg/pipeline/stream/source_acker.go#L35-L37)).
+  Note that the order is only preserved for messages coming from the same
+  source.
 - If a message failed to be sent to the DLQ the nack action should fail and
   stop the pipeline.
 - Allow the user to configure the window and number of messages that need to be
--- a/docs/design-documents/20221028-dead-letter-queue.md
+++ b/docs/design-documents/20221028-dead-letter-queue.md
@@ -27,12 +27,13 @@ The design for the DLQ should support the following:
 - Ensure message order is preserved when messages are sent to the DLQ.
   In other words - the order in which messages were produced by the source
   connector should be preserved, even if messages are nacked in a different
-  order.
+  order (this is already assured using
+  a [semaphore](https://github.com/ConduitIO/conduit/blob/b6325584d51ec64bc9086faddcdaa2788a5dfa8f/pkg/pipeline/stream/source_acker.go#L35-L37)).
 - If a message failed to be sent to the DLQ the nack action should fail and
   stop the pipeline.
-- Allow the user to configure the window and percentage of messages that need
-  to be nacked before the pipeline is considered broken and stops. Example: if
-  90% of the last 30 messages are nacked stop the pipeline.
+- Allow the user to configure the window and number of messages that need to be
+  nacked to consider the pipeline broken and trigger a stop. Example: if 25 of
+  the last 30 messages are nacked stop the pipeline.
 - Messages pushed to the DLQ should contain metadata describing the error and
   the source connector.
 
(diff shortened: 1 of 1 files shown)


docs/design-documents/20221028-dead-letter-queue.md
```diff
@@ -0,0 +1,144 @@
...
+- Allow the user to configure the window and percentage of messages that need
+  to be nacked before the pipeline is considered broken and stops. Example: if
+  90% of the last 30 messages are nacked stop the pipeline.
+- Messages pushed to the DLQ should contain metadata describing the error and
+  the source connector.
+
+### Questions
+
+- **Should we allow the user to attach custom metadata to messages pushed to
+  the DLQ?**
```
maha-hajja, line 42: in what case would this be useful for the user?

docs/design-documents/20221028-dead-letter-queue.md
```diff
@@ -0,0 +1,144 @@
...
+- A message sent to the DLQ should reflect the contents it had just before the
+  error happened. Example: if a message went through 2 processors and failed
+  while being processed by a 3rd processor, the DLQ message should reflect any
+  transforms done by the first 2 processors.
+- Ensure message order is preserved when messages are sent to the DLQ.
+  In other words - the order in which messages were produced by the source
+  connector should be preserved, even if messages are nacked in a different
+  order.
+- If a message failed to be sent to the DLQ the nack action should fail and
+  stop the pipeline.
```
maha-hajja, line 32: we could add another layer here, maybe create a simple file destination and write the failed records there, with a message telling the user what happened. Of course, this could fail too and then we stop the pipelines. WDYT?

docs/design-documents/20221028-dead-letter-queue.md
```diff
@@ -0,0 +1,144 @@
...
+
+- Allow user to reroute nacked messages to any Conduit connector (builtin or
+  standalone).
+- A message sent to the DLQ should reflect the contents it had just before the
+  error happened. Example: if a message went through 2 processors and failed
+  while being processed by a 3rd processor, the DLQ message should reflect any
+  transforms done by the first 2 processors.
+- Ensure message order is preserved when messages are sent to the DLQ.
+  In other words - the order in which messages were produced by the source
+  connector should be preserved, even if messages are nacked in a different
```
maha-hajja, line 29: how do we ensure this in the implementations below?

docs/design-documents/20221028-dead-letter-queue.md
```diff
@@ -0,0 +1,144 @@
...
+
+  This strategy would reroute nacked messages to a destination connector
+  plugin. This strategy needs to be able to get additional configuration
+  options defining the plugin name and its settings.
+
+### API
+
+The DLQ should be defined as a property of
+the [`Pipeline.Config`](https://github.com/ConduitIO/conduit/blob/d19379efc04d20d12ab9c80df82a29fcef7e8afd/proto/api/v1/api.proto#L28)
+entity. It's is a config option that can be updated even after the pipeline is
```
maha-hajja, line 98:
```suggestion
entity. It's a config option that can be updated even after the pipeline is
```

docs/design-documents/20221028-dead-letter-queue.md
maha-hajja: oh just noticed this in the future work, it's the same as a fallback DLQ but we decide what the destination will be, not the user.

lovromazgon: Right, that would be essentially the same as having a fallback to `builtin:file`. If we want this feature I'd rather go with the file connector than write one in Conduit for this purpose.

docs/design-documents/20221028-dead-letter-queue.md
lovromazgon: This is actually already ensured in the source acker node using the [semaphore](https://github.com/ConduitIO/conduit/blob/b6325584d51ec64bc9086faddcdaa2788a5dfa8f/pkg/pipeline/stream/source_acker.go#L35-L37).

docs/design-documents/20221028-dead-letter-queue.md
lovromazgon: Good question. Given that all DLQ messages would get the same metadata I don't know if it would be that useful.

docs/design-documents/20221028-dead-letter-queue.md
```diff
@@ -0,0 +1,192 @@
...
+
+The design for the DLQ should support the following:
+
+- Allow user to reroute nacked messages to any Conduit connector (builtin or
+  standalone).
+- A message sent to the DLQ should reflect the contents it had just before the
+  error happened. Example: if a message went through 2 processors and failed
+  while being processed by a 3rd processor, the DLQ message should reflect any
+  transforms done by the first 2 processors.
+- Ensure message order is preserved when messages are sent to the DLQ.
```
hariso, line 27: It might be useful to explicitly state this is about messages coming from the same source.

docs/design-documents/20221028-dead-letter-queue.md
maha-hajja: if we can come up with a useful use case, then sure, we'll give the user the ability to attach to the metadata! but for now I can't think of any either.

docs/design-documents/20221028-dead-letter-queue.md
```diff
@@ -0,0 +1,192 @@
...
+  In other words - the order in which messages were produced by the source
+  connector should be preserved, even if messages are nacked in a different
+  order (this is already assured using
+  a [semaphore](https://github.com/ConduitIO/conduit/blob/b6325584d51ec64bc9086faddcdaa2788a5dfa8f/pkg/pipeline/stream/source_acker.go#L35-L37)).
+- If a message failed to be sent to the DLQ the nack action should fail and
+  stop the pipeline.
+- Allow the user to configure the window and number of messages that need to be
+  nacked to consider the pipeline broken and trigger a stop. Example: if 25 of
+  the last 30 messages are nacked stop the pipeline.
+- Messages pushed to the DLQ should contain metadata describing the error and
```
hariso, line 37: Maybe also add the node in which the failure occurred, unless it's already found somewhere else?

docs/design-documents/20221028-dead-letter-queue.md
maha-hajja: Yes I meant using the builtin connector! when I mentioned

> simple file destination

I meant creating the actual output file that would have the records.

docs/design-documents/20221028-dead-letter-queue.md
```diff
@@ -0,0 +1,192 @@
...
+to the DLQ and return `nil` if that action succeeded or an error otherwise. The
+synchronous write is needed to ensure the DLQ message actually reached its
+destination, otherwise we need to consider the nack failed and need to stop the
+pipeline.
+
+### DLQ Strategies
+
+We propose to implement 2 DLQ strategies:
+
+- `log` - This strategy would cause a nacked message to be logged to the
```
hariso, line 80: Maybe we add a logging connector, and reduce this to one strategy?

docs/design-documents/20221028-dead-letter-queue.md
```diff
@@ -0,0 +1,192 @@
...
+- `plugin` - This strategy would reroute nacked messages to a destination
+  connector plugin. This strategy needs to be able to get additional
+  configuration options defining the plugin name and its settings.
+
+Regardless of the strategy the user can also configure a stop window and stop
+threshold.
+
+- `stop_window` defines how many last acks/nacks are monitored in the window
+  that controls if the pipeline should stop. 0 disables the window.
+- `stop_threshold` defines the number of nacks in the window that would cause
```
hariso, line 92: A percentage would be good too (depending on the volume of messages, the same number of nacks may be relevant or irrelevant).

docs/design-documents/20221028-dead-letter-queue.md
hariso: In case the DLQ is a connector, it's relatively easy to add some metadata through a processor.

docs/design-documents/20221028-dead-letter-queue.md
```diff
@@ -0,0 +1,192 @@
...
+  a [semaphore](https://github.com/ConduitIO/conduit/blob/b6325584d51ec64bc9086faddcdaa2788a5dfa8f/pkg/pipeline/stream/source_acker.go#L35-L37)).
+- If a message failed to be sent to the DLQ the nack action should fail and
+  stop the pipeline.
+- Allow the user to configure the window and number of messages that need to be
+  nacked to consider the pipeline broken and trigger a stop. Example: if 25 of
+  the last 30 messages are nacked stop the pipeline.
+- Messages pushed to the DLQ should contain metadata describing the error and
+  the source connector.
+
+### Questions
```
hariso, line 40: Should we allow multiple DLQs? I imagine there would be use cases such as nacked messages from one source go into the same DLQ, or that there are DLQs for records of different importance levels.

docs/design-documents/20221028-dead-letter-queue.md
lovromazgon: Right, we could add a processor in front of the DLQ connector. The problem is how we expose this functionality to the user, it might unnecessarily complicate things, so if we see no reason I'd rather skip it altogether and keep it simple.

docs/design-documents/20221028-dead-letter-queue.md
lovromazgon: This information is currently found in the error message. To me it sounds as good enough given that a human will need to read the error to figure out how to reprocess the messages.

docs/design-documents/20221028-dead-letter-queue.md
lovromazgon: I like that 👍 I'll try changing the config to see how it would turn out if we do this.

docs/design-documents/20221028-dead-letter-queue.md
lovromazgon: I started out with a percentage but switched to an absolute number. My reasoning was that "90% of 30 messages" or "27 of 30 messages" are both the same, although the second one is easier to grasp. Also if the config says "92% of 30 messages" I'm not exactly sure what the threshold is, 27 or 28? So I went with the absolute number for clarity.

I hope that's what you meant, if not let me know.

docs/design-documents/20221028-dead-letter-queue.md
lovromazgon: I honestly don't know how many users would need this, that's why I sided on the side of simplicity, so we don't work on something that potentially won't be used by anyone. That said, the workaround to achieve this behavior would be to sort records after they land in the DLQ. For example, you could push DLQ messages into a certain Kafka topic and have another pipeline that reads that topic, categorizes the messages and moves them to different DLQ topics.

docs/design-documents/20221028-dead-letter-queue.md
lovromazgon: I'm thinking what we're trying to achieve with this? Sounds like we are trying to make sure that the pipeline can keep on running even if a message fails to be pushed to the DLQ. In this case I would rather provide a generic solution for chaining DLQs as fallbacks, I wouldn't want to make the assumption that the user is fine with writing records to a file.

That said, I think even without the possibility of fallbacks we should already cover most of our users needs, so I wrote this down as future work.

docs/design-documents/20221028-dead-letter-queue.md
lovromazgon: Btw - any thoughts about the naming? I'm not sure I'm _entirely_ happy with `stop_window` and `stop_threshold`, if anyone has better suggestions I'm all ears 🙂

docs/design-documents/20221028-dead-letter-queue.md
lovromazgon: Thanks for the suggestion, I think this simplified the config a lot. WDYT?

docs/design-documents/20221028-dead-letter-queue.md
hariso: >  The problem is how we expose this functionality to the user

Maybe in the same way as a normal processor, i.e. a setting under the DLQ connector? That would make a DLQ connector behave like any connector, which would maybe simplify things (no special rules like "it mustn't have a processor").

docs/design-documents/20221028-dead-letter-queue.md
hariso: > This information is currently found in the error message.

That will probably be good enough :+1: And if we ever need it, it'll be easy to add anyway.

> To me it sounds as good enough given that a human will need to read the error to figure out how to reprocess the messages.

That's true. :+1: I was more for the option where you have a separate field, so you can more easily search the DLQ and figure out if it's a systematic failure of a node (e.g. a bad processor) and so on. But again, I don't think it's such a huge thing in this context, and it's easy to add if we ever figure out we need it.

docs/design-documents/20221028-dead-letter-queue.md
hariso: I think I should have expressed myself better in the first comment.:) The reason I suggested percentages too (in addition to what you already described for the window of last N messages) is to have a global limit (vs. local in what's in the doc).

It could be that due to a reason last N out of N messages got nacked, but because of many more ack'd messages before them, the overall success rate is very high and acceptable. That might have been a glitch and things continue to work as expected after the glitch.

On the other hand, it may be a sign of an issue which is worsening, so someone would like to focus on a sliding window.

In other words, I see two different options for a limit on the errors: global and local, and I see use cases for both.

> Btw - any thoughts about the naming? I'm not sure I'm _entirely_ happy with `stop_window` and `stop_threshold`, if anyone has better suggestions I'm all ears slightly_smiling_face

A slightly different alternative I can suggest is `sliding_window_size`/`sliding_window_threshold` or just `window_size`/`threshold` or `limit`. Adding the `sliding` prefix might  make it more clear. Since these come in the DLQ section, just saying `threshold`/`limit` is maybe enough (i.e. I think it can be easily understood that it's the DLQ threshold/limit and that it will cause a stop). I don't think it's really far from what you already suggested though.:)

docs/design-documents/20221028-dead-letter-queue.md
hariso: Even though I can imagine use cases, I can't say either how many would use, so I definitely agree: let's keep it simple.:)

docs/design-documents/20221028-dead-letter-queue.md
lovromazgon: I'm thinking that this could be a "nice to have" and not necessarily something that needs to be in the scope of the first implementation. My proposal - we write it down as future work, create an issue for it, and if time is left we add it. WDYT?

hariso: Agree 200% with that :+1:

docs/design-documents/20221028-dead-letter-queue.md
lovromazgon: Let me see if I understand this - you're saying we should also introduce a global limit of nacked messages as a percentage of all processed messages? I see a few problems with this. Let's say we start a pipeline and the first message gets nacked, globally that's 100% of all messages, so that will necessarily stop the pipeline if any global limit is configured. On the other hand, if a pipeline is running for a _long_ time, the global limit will take a long time to take effect, even if all messages start getting nacked (after 1 year a 1% limit would present roughly 3,5 days worth of nacked messages). I think the usefulness of a global limit is _limited_ (pun intended 🙃). It only works as expected within a certain frame, that's what the window is trying to address.

> It could be that due to a reason last N out of N messages got nacked /.../ That might have been a glitch and things continue to work as expected after the glitch.

Isn't that the purpose of configuring a stop window? It lets you automatically stop the pipeline when it starts misbehaving. If such a "glitch" stops a pipeline without the user wanting that to happen, then they should configure a larger window and/or threshold, because they apparently don't care about N nacked messages in a row.

---

Regarding the variable naming, you're right maybe `stop` is redundant. I'd also say that `sliding` would be redundant, I don't imagine what other window this would be if not a sliding one 😄 What about the combination `window_size` and `window_nack_threshold`?

hariso: > Isn't that the purpose of configuring a stop window?

I'd say it's for sure one of the purposes, but I imagine there are also cases where you don't care about some local peaks, as long as you're overall in some limits. However, I don't think at all it's a big thing with or without it, nor would the design radically change, so I'm fine leaving it out, especially for the sake of unblocking the design doc and progressing.

> What about the combination `window_size` and `window_nack_threshold`?

I like those!

docs/design-documents/20221028-dead-letter-queue.md
```diff
@@ -0,0 +1,188 @@
...
+- `stop_window` defines how many last acks/nacks are monitored in the window
+  that controls if the pipeline should stop. 0 disables the window.
+- `stop_threshold` defines the number of nacks in the window that would cause
+  the pipeline to stop.
+
+The default DLQ plugin should be `builtin:log` with a `stop_window` of 1 and a
+`stop_threshold` of 1. This would result in the same behavior as before the
+introduction of DLQs - the first nacked message would stop the pipeline.
+
+### API
```
hariso, line 104: Will a DLQ be accessible through the API in the same a normal connector is? Also, WDYT about exposing the state of the DLQ (current number of nacked messages in the window)?

lovromazgon: The current idea is to make the DLQ settings configurable through the pipeline endpoints and make it part of the pipeline config, so no, it wouldn't be accessible as a normal connector.

A second option would be to create another connector type for a DLQ connector, but it didn't feel entirely right, since that's still a destination connector that just acts as the DLQ connector. And I had a problem with attaching the nack window settings (we could inject them in the connector settings with a prefix like `conduit.dlq.*`). Seeing a DLQ being configured as part of the connectors in the pipeline config file also seemed off (I'd like to keep the pipeline config as close to the API representation as possible). At that point it seemed to go down the "hacky" route so I rather went with adding it to the pipeline config itself.

A third option I was thinking of is to separate the DLQ config out into a separate endpoint (e.g. `/pipelines/{id}/dlq`), but that seemed excessive, keeping the DLQ config as part of the pipeline config felt simpler.

> WDYT about exposing the state of the DLQ (current number of nacked messages in the window)?

That's a very good point, it's important to be able to monitor the DLQ. I'll add metrics to the design! 📊

hariso: > so no, it wouldn't be accessible as a normal connector.

:+1: That makes sense. That also prevents a DLQ from being used as a normal destination, which I think is good.

hariso: Last thing:) Since DLQ is still a destination a connector, adding a DLQ through the gRPC or HTTP API means that we need to stop a pipeline, add it, and restart the pipeline, correct?

hariso (approved):

PR #719, opened 2022-11-04
Stream Inspector: Proto changes for connectors

hariso:
### Description

Proto changes for the stream inspector for connectors.

### Quick checks:

### DIFF
--- a/proto/api/v1/api.proto
+++ b/proto/api/v1/api.proto
@@ -7,7 +7,6 @@ import "google/api/field_behavior.proto";
 import "google/api/annotations.proto";
 import "protoc-gen-openapiv2/options/annotations.proto";
 import "google/protobuf/timestamp.proto";
-import "google/protobuf/descriptor.proto";
 
 message Pipeline {
   // Status describes the pipeline status.
@@ -437,9 +436,10 @@ service ConnectorService {
     };
   };
 
-  rpc InspectConnector(InspectConnectorRequest) returns (stream opencdc.v1.Record) {
+  rpc InspectConnector(InspectConnectorRequest) returns (stream InspectConnectorResponse) {
     option (google.api.http) = {
       get: "/v1/connectors/{id}/inspect"
+      response_body: "record"
     };
   }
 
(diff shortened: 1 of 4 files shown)


proto/api/v1/api.proto
```diff
@@ -436,6 +437,12 @@ service ConnectorService {
     };
   };
 
+  rpc InspectConnector(InspectConnectorRequest) returns (stream opencdc.v1.Record) {
+    option (google.api.http) = {
+      get: "/v1/connectors/{id}/inspect"
```
lovromazgon, line 442: I didn't try this out, but in normal endpoints (no websocket) it is possible to set the option `response_body` and extract a field from the message that's being returned. This could possibly help us make the linter happy and keep the websocket response clean.

So having a message like this:
```
message InspectConnectorResponse {
  opencdc.v1.Record record = 1;
}
```

And add an option here:
```suggestion
      get: "/v1/connectors/{id}/inspect"
      response_body: "record"
```

Again, not sure if this works for websocket endpoints.

proto/api/v1/api.proto
```diff
@@ -1,12 +1,13 @@
...
 
 package api.v1;
-option go_package = "github.com/conduitio/conduit/proto/api/v1;api";
 
+import "opencdc/v1/opencdc.proto";
 import "google/api/field_behavior.proto";
 import "google/api/annotations.proto";
 import "protoc-gen-openapiv2/options/annotations.proto";
 import "google/protobuf/timestamp.proto";
+import "google/protobuf/descriptor.proto";
```
lovromazgon, line 10: Do we need this import? It's strange, because the linter doesn't complain that it's there, but it also doesn't complain if I remove it 🤔

hariso: TBH, I have no idea how it got here.:) I can't remember adding it myself, but could have a copy-paste mistake. I'll remove it, thanks for noticing it! And yes, definitely interesting that the linter doesn't complain either way.:S

lovromazgon (approved):
Buf is complaining a bit

> File option "go_package" changed from "github.com/conduitio/conduit/proto/api/v1;api" to "".

That's fine! Anyone using go to access the gRPC API should use the remotely generated code anyway.

> Previously present reserved range "[13]" on message "Swagger" was deleted.
> Previously present reserved range "[1]" on message "Tag" was deleted.

Not sure what buf is complaining about here, maybe something in a dependency changed. Doesn't seem like an issue for us though!

---

LGTM 💪

PR #732, opened 2022-11-09
Buf: using locally generated code

hariso:
### Description

With the changes in #719 we started facing issues using the remotely generated code. This issue has been discussed with the Buf team ([link](https://bufbuild.slack.com/archives/CRZ680FUH/p1667911231953259)) and they are working on it, but for the time being we don't have a good workaround.

With that, we're reverting to using locally generated code by `buf`.

### Quick checks:


.golangci.yml
```diff
@@ -69,7 +69,7 @@ linters:
     - goimports
     - revive
     # - gomnd
-    - gomoddirectives
+    # - gomoddirectives
```
hariso, line 72: Adding `nolint` to `go.mod` didn't help.

pkg/conduit/runtime.go
```diff
@@ -48,12 +48,12 @@ import (
 	"github.com/conduitio/conduit/pkg/web/api"
 	"github.com/conduitio/conduit/pkg/web/openapi"
 	"github.com/conduitio/conduit/pkg/web/ui"
+	apiv1 "github.com/conduitio/conduit/proto/gen/api/v1"
 	grpcruntime "github.com/grpc-ecosystem/grpc-gateway/v2/runtime"
 	"github.com/piotrkowalczuk/promgrpc/v4"
 	promclient "github.com/prometheus/client_golang/prometheus"
 	"github.com/prometheus/client_golang/prometheus/promhttp"
 	"github.com/rs/zerolog"
-	apiv1 "go.buf.build/conduitio/conduit/conduitio/conduit/api/v1"
```
hariso, line 56: Our initial plan was to keep the same imports and point them to the code in `proto/gen` using `go mod replace`. However, golangci-lint was complaining about the `replace` directive, and it wasn't possible to ignore just that line.

Makefile
```diff
@@ -9,7 +9,7 @@ VERSION=`git describe --tags --dirty`
 GO_VERSION_CHECK=`./scripts/check-go-version.sh`
 
 # The build target should stay at the top since we want it to be the default target.
-build: check-go-version pkg/web/ui/dist
+build: check-go-version pkg/web/ui/dist proto-generate
```
hariso, line 12: @lovromazgon @maha-hajja wdyt about adding this to the `test` goal too, to make sure the code is always fresh?

pkg/conduit/runtime.go
lovromazgon: Oh well 🤷 We can replace back again when/if we move back to remote generation.

Makefile
hariso: Makes sense. :+1:

.github/workflows/buf.yml
```diff
@@ -28,18 +28,3 @@ jobs:
         with:
           input: 'proto'
           against: '.git#branch=origin/main,subdir=proto'
-
-  # Push buf module to the buf schema registry, but only if the validate action succeeded and if the action is running
-  # on branch main.
```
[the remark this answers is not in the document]
hariso: Good point about other users :+1:

pkg/web/openapi/openapi.go
```diff
@@ -12,8 +12,6 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-//go:generate ./copy-swagger-api.sh
```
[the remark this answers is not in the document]
hariso: Good point, I realized I already had an up-to-date version of swagger.json, so it worked without the script.

proto/buf.gen.yaml
```diff
@@ -5,7 +5,7 @@ version: v1
 managed:
   enabled: true
   go_package_prefix:
-    default: go.buf.build/conduitio/conduit/conduitio/conduit
+    default: "github.com/conduitio/conduit/proto/gen"
     except:
       - buf.build/googleapis/googleapis
     override:
```
lovromazgon, line 11: One last thing - I propose we change the line below to this:
```
      buf.build/grpc-ecosystem/grpc-gateway: go.buf.build/protocolbuffers/go/grpc-ecosystem/grpc-gateway
```
In other words, we should use the [`protocolbuffers/go`](https://buf.build/protocolbuffers/templates/go) template instead of `conduitio/conduit`, the buf team keeps that template updated in case new versions of the protoc-gen-go plugin are released ([doc](https://docs.buf.build/bsr/remote-generation/remote-plugin-execution#protoc-based-plugins)).

We need to regenerate the proto code and run `go mod tidy` after this.

Makefile
lovromazgon: Seems like you forgot about this one.

hariso: I didn't actually, I thought your answer was about my question adding it to the `test` goal and not the `build` goal. I'm viewing it from another angle: the fact that we change it infrequently means that it's easy to forget, and a few seconds are worth preventing outdated code. I see your point though, and don't feel strongly about it.

lovromazgon: Oh now I see this is on `build` 🤦 Well I still think we shouldn't run this if the generated code is already checked in, in most cases it won't do anything and this seems something that we should catch in PRs. Essentially the same reason why we don't run `go generate` before every build.

hariso: I understand that point of view. :+1: I've removed the proto-generate before build just now.

lovromazgon (approved):

PR #737, opened 2022-11-14
docs: spell out dead-letter-queue

lovromazgon:
We are changing the user-facing naming of DLQ to `dead-letter-queue` (endpoints, config files).

I also updated the definition of the nack threshold. After starting to work with it I think it's more intuitive to define what is still tolerated rather than when the pipeline stops. For example, a threshold of `0` means no nacks are tolerated (first failed message brings the number of nacked messages to `1` which crosses the threshold and stops the pipeline).


hariso (approved):
:muscle:

PR #744, opened 2022-11-18
Parallel processors

lovromazgon:
## Description

This PR introduces a `ParallelNode` which can execute any `PubSubNode` in parallel. We expose this functionality by introducing the field `Workers` on a processor, which allows the user to define how many workers should execute a processor in parallel.

To see it in action apply [this patch](https://github.com/ConduitIO/conduit/files/10697684/parallel.patch) and run [`TestPipelineSimple`](https://github.com/ConduitIO/conduit/blob/44e05b75f1186656b8918f7dd2b6909cc898e3d0/pkg/orchestrator/orchestrator_test.go#L54):
```
go test -v -count=1 -race -test.run "TestPipelineSimple" ./pkg/orchestrator/
```
Adjust the [number of workers](https://github.com/ConduitIO/conduit/blob/44e05b75f1186656b8918f7dd2b6909cc898e3d0/pkg/orchestrator/orchestrator_test.go#L124). The processor will take 100 milliseconds to execute and it logs a message at the start and at the end. Notice how multiple processors are run in parallel if you increase the number of workers.

Log output for 1 worker:
```
2022-11-18T19:46:00+00:00 WRN 00.522: removing metadata... message_id=5bc56954-bcb5-4178-8b25-517bbde87d66/2
2022-11-18T19:46:00+00:00 WRN 00.623: removed metadata! message_id=5bc56954-bcb5-4178-8b25-517bbde87d66/2
2022-11-18T19:46:00+00:00 WRN 00.624: removing metadata... message_id=5bc56954-bcb5-4178-8b25-517bbde87d66/4
2022-11-18T19:46:00+00:00 WRN 00.725: removed metadata! message_id=5bc56954-bcb5-4178-8b25-517bbde87d66/4
2022-11-18T19:46:00+00:00 WRN 00.725: removing metadata... message_id=5bc56954-bcb5-4178-8b25-517bbde87d66/6
2022-11-18T19:46:00+00:00 WRN 00.827: removed metadata! message_id=5bc56954-bcb5-4178-8b25-517bbde87d66/6
2022-11-18T19:46:00+00:00 WRN 00.827: removing metadata... message_id=5bc56954-bcb5-4178-8b25-517bbde87d66/8
2022-11-18T19:46:00+00:00 WRN 00.928: removed metadata! message_id=5bc56954-bcb5-4178-8b25-517bbde87d66/8
2022-11-18T19:46:00+00:00 WRN 00.929: removing metadata... message_id=5bc56954-bcb5-4178-8b25-517bbde87d66/10
2022-11-18T19:46:01+00:00 WRN 01.029: removed metadata! message_id=5bc56954-bcb5-4178-8b25-517bbde87d66/10
```

Output for 2 workers:
```
2022-11-18T19:47:06+00:00 WRN 06.761: removing metadata... message_id=4a594d58-3ce8-4f4c-88a6-ada5b0e127a5/4
2022-11-18T19:47:06+00:00 WRN 06.761: removing metadata... message_id=4a594d58-3ce8-4f4c-88a6-ada5b0e127a5/2
2022-11-18T19:47:06+00:00 WRN 06.862: removed metadata! message_id=4a594d58-3ce8-4f4c-88a6-ada5b0e127a5/2
2022-11-18T19:47:06+00:00 WRN 06.862: removed metadata! message_id=4a594d58-3ce8-4f4c-88a6-ada5b0e127a5/4
2022-11-18T19:47:06+00:00 WRN 06.862: removing metadata... message_id=4a594d58-3ce8-4f4c-88a6-ada5b0e127a5/6
2022-11-18T19:47:06+00:00 WRN 06.862: removing metadata... message_id=4a594d58-3ce8-4f4c-88a6-ada5b0e127a5/8
2022-11-18T19:47:06+00:00 WRN 06.963: removed metadata! message_id=4a594d58-3ce8-4f4c-88a6-ada5b0e127a5/6
2022-11-18T19:47:06+00:00 WRN 06.963: removed metadata! message_id=4a594d58-3ce8-4f4c-88a6-ada5b0e127a5/8
2022-11-18T19:47:06+00:00 WRN 06.963: removing metadata... message_id=4a594d58-3ce8-4f4c-88a6-ada5b0e127a5/10
2022-11-18T19:47:07+00:00 WRN 07.063: removed metadata! message_id=4a594d58-3ce8-4f4c-88a6-ada5b0e127a5/10
```

### Quick checks:

- [ ] I have made sure that the PR is of reasonable size and can be easily reviewed. (the bulk of the work is in `parallel.go` and `parallel_test.go`, I can extract the API changes into a separate PR but those are pretty simple and wouldn't make this review any easier IMO)


pkg/pipeline/stream/base.go
```diff
@@ -344,13 +344,16 @@ func (n *nodeBase) Send(
 	if msg.Ctx == nil {
 		msg.Ctx = ctxutil.ContextWithMessageID(ctx, msg.ID())
 	}
+	// copy context into a local variable, we shouldn't access it anymore after
+	// we send the message to out
+	msgCtx := msg.Ctx
```
lovromazgon, line 349: This solves a race condition where `msg.Ctx` could be changed in another goroutine after it was sent to `out`.

pkg/foundation/semaphore/semaphore.go
```diff
@@ -31,7 +31,7 @@ type Simple struct {
 // Ticket reserves a place in the queue and can be used to acquire access to a
 // Lock.
 type Ticket struct {
-	Lock
+	l Lock
```
lovromazgon, line 34: This makes the semaphore use a bit safer, previously a caller could do `Ticket.Lock` to get the lock object without actually acquiring it through the semaphore. Of course that is an invalid use of the struct, but now we prevent that entirely.

pkg/pipeline/stream/logger.go
```diff
@@ -21,20 +21,39 @@ import (
 )
 
 // SetLogger figures out if the node needs a logger, sets static metadata in the
-// logger and supplies it to the node.
-func SetLogger(n Node, logger log.CtxLogger) {
+// logger and supplies it to the node. Behavior can be customized by supplying
+// custom options.
+func SetLogger(n Node, logger log.CtxLogger, options ...func(log.CtxLogger, Node) log.CtxLogger) {
```
lovromazgon, line 26: This function is backwards compatible with existing calls but adds the ability to change the behavior if needed (we want to change the behavior for parallel worker nodes).

pkg/provisioning/parse.go
```diff
@@ -28,6 +28,7 @@ import (
 type ProcessorConfig struct {
 	Type     string            `yaml:"type"`
 	Settings map[string]string `yaml:"settings"`
+	Workers  int               `yaml:"workers"`
```
maha-hajja, line 31: if no workers are defined, do we default to 1 worker? also, we should add a validation for the workers to be a positive number, in `provisioning/validate.go `

lovromazgon: Yes, we default to 1 worker, but it's a bit further down the line ([here](https://github.com/ConduitIO/conduit/blob/aa4b1587fc7cf4592166f74f132d9b9831a55ed1/pkg/pipeline/lifecycle.go#L268)). I can add a default and validation here, makes sense. Same thing in the API endpoint probably.

lovromazgon: I added validations for the workers variable in the pipeline config file (`provisioning/validate.go`) and in `processor.Service` (which will bubble up to the API). Also did some cleaning up in tests (moved tests to `matryer/is`, usage of temporary folder in runtime test).

pkg/provisioning/validate_test.go
```diff
@@ -137,6 +137,21 @@ func TestValidator_InvalidFields(t *testing.T) {
...
+	}, {
+		name: "processor workers is negative",
+		config: PipelineConfig{
+			Status:      "running",
+			Name:        "pipeline1",
+			Description: "desc1",
+			Processors: map[string]ProcessorConfig{
+				"proc1": {
+					// invalid field
+					Type:     "js",
```
hariso, line 149: Does this refer to `Type`?

pkg/pipeline/stream/message.go
```diff
@@ -272,6 +272,19 @@ func (m *Message) Status() MessageStatus {
 	}
 }
 
+// StatusError returns the error that was returned when the message was acked or
+// nacked. If the message was successfully acked/nacked or it is still open the
+// method returns nil.
+func (m *Message) StatusError() error {
+	switch m.Status() {
+	case MessageStatusAcked:
+		return m.Ack()
```
hariso, line 281: This is to cover the DLQ case?

pkg/provisioning/validate_test.go
lovromazgon: Oh, the comment was probably copied from another test case, I'll move it down to `Workers`.

pkg/pipeline/stream/message.go
lovromazgon: This specific line is to cover the case if a message is acknowledged by a processor (i.e. filtered out) and we fail to deliver the acknowledgment to the connector - in that case, the processor node will return an error, causing the whole pipeline to stop. The next case is for the DLQ, in case the processor failed to process the message and nacked it but the nack failed (i.e. it wasn't stored in the DLQ), that also causes the processor node to fail and stop the pipeline.

So essentially the result of this function tells us if the processor node is running or if it failed to process the message and it stopped. Based on this we can stop the goroutine that's wrapping the processor node.

maha-hajja (approved):
lgtm

pkg/pipeline/stream/message.go
hariso: That's some useful info, it would be nice to have a short comment about that. :+1:

lovromazgon: But is this the right place? 🤔 I mean this function can be used in other places as well. I can try to add this info somewhere in the parallel node though.

lovromazgon: WDYT? https://github.com/ConduitIO/conduit/pull/744/commits/b508934b9429b575b6973c0eb42b6b7c31f9a6f2

hariso: Excellent!

pkg/pipeline/stream/parallel.go
```diff
@@ -0,0 +1,337 @@
...
+		donePool: donePool,
+	}
+}
+
+func (c *parallelNodeCoordinator) Run(ctx context.Context) {
+	// fail toggles a short circuit that causes all future messages to be nacked
+	fail := false
+	for job := range c.jobs {
+		// wait for job to be done
+		<-job.Done
```
hariso, line 186: IIUC, this is how message order is being maintained? Every message becomes a job, as they are coming, they are pushed into the jobs "queue" (i.e. channel) and taken out in the same order.

hariso (approved):
:raised_hands:

pkg/pipeline/stream/parallel.go
lovromazgon: Exactly 👍 I'll add a comment to make this clear.

pkg/pipeline/stream/parallel.go
```diff
@@ -0,0 +1,328 @@
...
+			// message was nacked, i.e. sent to DLQ or dropped
+			w.logger.Trace(ctx).Msg("worker nacked the message")
+		case _, ok := <-out:
+			// message was processed
+			if ok {
+				w.logger.Trace(ctx).Msg("worker successfully processed the message")
+			} else {
+				w.logger.Trace(ctx).Msg("worker stopped running")
+			}
+		}
```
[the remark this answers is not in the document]
lovromazgon: That's why we add them automatically to all logs that are logged using the message context (see [`MessageIDLogCtxHook`](https://github.com/ConduitIO/conduit/blob/01366a736f5d2ece76f503e0e06a557904a6d0a0/pkg/foundation/ctxutil/messageid.go#L44)) 🙂

pkg/provisioning/validate.go
```diff
@@ -71,7 +71,10 @@ func validateProcessorsConfig(mp map[string]ProcessorConfig) error {
 	var err error
 	for k, cfg := range mp {
 		if cfg.Type == "" {
-			err = multierror.Append(err, cerrors.Errorf("processors %q: \"type\" is mandatory: %w", k, ErrMandatoryField))
+			err = multierror.Append(err, cerrors.Errorf("processor %q: \"type\" is mandatory: %w", k, ErrMandatoryField))
+		}
+		if cfg.Workers < 0 {
+			err = multierror.Append(err, cerrors.Errorf("processor %q: \"workers\" can't be negative: %w", k, ErrInvalidField))
```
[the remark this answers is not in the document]
lovromazgon: Hmm, I'm not sure, I have a few concerns:
- We would duplicate the code that's already in the SDK.
- We could only partially validate the SDK based on the specifications. In other words - the config is not necessarily "valid" if the spec validations succeed, a connector is allowed to do manual validations in `Configure`.

And another more general concern about the way we validate the config - I'm not sure the behavior would be expected in case an existing pipeline is being updated through the configuration file. Currently, we simply don't apply the configuration if it's invalid. This means that Conduit keeps the old pipeline in memory which doesn't match the provided configuration. That seems strange, I would expect that Conduit would store the _invalid_ configuration in memory and just refuse to run the pipeline. So essentially I'm inclined to do even less validation rather than more.

Issue #747, opened 2022-11-22
Conduit cannot shutdown gracefully with open inspector sessions


hariso:
### Bug description

https://github.com/ConduitIO/conduit/pull/718#discussion_r1028081348

### Steps to reproduce

1. Start Conduit with a pipeline
2. Inspect a connector
3. Shut down Conduit gracefully (e.g. Ctrl+C or `kill -2`)

**Expected behavior**:
Conduit is shut down.

**Actual behavior**:
Conduit is not shut down.

### Version

TBD

PR #752, opened 2022-11-28
WebSocket proxy for the grpc-gateway

hariso:
### Description
This PR extracts from https://github.com/tmc/grpc-websocket-proxy/ the most important parts of code to us. This library has been introduced as a dependency in #718.

**Update**: gorilla/websocket got archived: https://github.com/gorilla#gorilla-toolkit

### Quick checks:


pkg/foundation/grpcutil/websocket.go
```diff
@@ -0,0 +1,132 @@
...
+}
+
+func newWebSocketProxy(handler http.Handler, logger log.CtxLogger) *wsProxy {
+	return &wsProxy{
+		handler: handler,
+		logger:  logger.WithComponent("grpcutil.websocket"),
+		upgrader: websocket.Upgrader{
+			ReadBufferSize:  1024,
+			WriteBufferSize: 1024,
+		},
```
lovromazgon, line 71: We should allow any origin, otherwise browsers won't be able to connect to the websocket (e.g. websocket king complains it can't connect).
```suggestion
		upgrader: websocket.Upgrader{
			ReadBufferSize:  1024,
			WriteBufferSize: 1024,
			CheckOrigin:     func(r *http.Request) bool { return true },
		},
```

lovromazgon (commented):
Seems like we will still need a read loop to figure out when the client closed the connection so we can cancel our context ([this](https://github.com/tmc/grpc-websocket-proxy/blob/673ab2c3ae75cc01952b84b88590e30e75dcf395/wsproxy/websocket_proxy.go#L246-L249) and [this](https://github.com/tmc/grpc-websocket-proxy/blob/673ab2c3ae75cc01952b84b88590e30e75dcf395/wsproxy/websocket_proxy.go#L233-L235)). Right now Conduit won't stop even if the client stops the inspector session because it thinks the connection is still open.

pkg/foundation/grpcutil/websocket.go
```diff
@@ -0,0 +1,132 @@
...
+type wsProxy struct {
+	handler  http.Handler
+	logger   log.CtxLogger
+	upgrader websocket.Upgrader
+}
+
+func newWebSocketProxy(handler http.Handler, logger log.CtxLogger) *wsProxy {
+	return &wsProxy{
+		handler: handler,
+		logger:  logger.WithComponent("grpcutil.websocket"),
```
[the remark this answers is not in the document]
hariso: The reason why I put `websocket` is because wsProxy is an internal/unexported type and because it's more clear what is it about. But I don't have anything against wsProxy either.

pkg/foundation/grpcutil/websocket.go
hariso: Yeah, I noticed that in the original implementation. I've intentionally removed because checking the origin should generally protect against CSRF. CLI tools and our UI continues to work with the checked origins, so it's only web sites like WebSocketKing which have problems and I think it's an acceptable trade off for the security.

pkg/foundation/grpcutil/websocket.go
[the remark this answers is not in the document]
lovromazgon: The function itself is also unexported, so I don't see the problem 🤷 The point of the logger component is to help us find where a log message originated from, so keeping that aligned with the struct name makes it easy to find it. All in all see no upside in using other names here.

hariso:
> Seems like we will still need a read loop to figure out when the client closed the connection so we can cancel our context ([this](https://github.com/tmc/grpc-websocket-proxy/blob/673ab2c3ae75cc01952b84b88590e30e75dcf395/wsproxy/websocket_proxy.go#L246-L249) and [this](https://github.com/tmc/grpc-websocket-proxy/blob/673ab2c3ae75cc01952b84b88590e30e75dcf395/wsproxy/websocket_proxy.go#L233-L235)). Right now Conduit won't stop even if the client stops the inspector session because it thinks the connection is still open.

Can you tell me a bit more about how you tested it? I tested this a few times in the following way:

1. Start Conduit with a pipeline
2. Inspect a session with `wscat`. Wait for some records.
3. Kill `wscat`
4. Stop Conduit with Ctrl-+C or `kill -SIGINT`

This worked. There's an issue with shutting down with an **open** inspector session though (#747).

pkg/foundation/grpcutil/websocket.go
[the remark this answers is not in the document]
hariso: Gotcha, I was understanding `WithComponent` more as a logical component, not the type, but what you said makes sense. Thanks for pointing this out!

lovromazgon:
Here are steps to reproduce what I mean:
1. Terminal 1: `./conduit -pipelines.path=examples/pipelines` (provisions the example pipeline)
2. Terminal 2: `wscat -c ws://localhost:8080/v1/connectors/file-to-file:example.out/inspect`
3. Terminal 3: `echo "hi" >> example.in` (see the record appear in the inspector and in `example.out`)
4. Terminal 2: Ctrl+C (kill wscat)
5. Terminal 1: Ctrl+C (kill Conduit)

Conduit then doesn't shut down, also the message `session removed` is never logged.

Note that you can reproduce the same thing by simply navigating to http://localhost:8080/ui/pipelines/file-to-file, since the UI opens an inspector session. Even after closing the tab Conduit thinks the session is still open.

hariso:
@lovromazgon Thanks for that! I used the same steps, but after I tried it out a few times I realized it takes a bit of time for the context to be closed and the session to be removed. In other words, if you perform steps 4 and 5, you'll get to the behavior you described. But if there's a short pause, then it's all good. Thanks for noticing this, I'll look into what's causing it.

pkg/foundation/grpcutil/websocket_test.go
```diff
@@ -0,0 +1,61 @@
...
+	is       *is.I
+	response string
+}
+
+func (h *testHandler) ServeHTTP(w http.ResponseWriter, _ *http.Request) {
+	_, err := w.Write([]byte(h.response))
+	h.is.NoErr(err)
+}
+
+func TestWebSocket_NoUpgradeToWebSocket(t *testing.T) {
```
hariso, line 38: Writing tests for this turned out to be more complex than I thought. A simple http.Handler which we use in this test doesn't work because, as soon as ServeHTTP is done, the request context is cancelled, and so the WebSocket connection is closed. 

I think I'll need to mimic the setup we have when starting Conduit (grpc-gateway and buf-generated code).

pkg/foundation/grpcutil/websocket.go
```diff
@@ -0,0 +1,253 @@
...
+	}
+	p.proxy(w, r)
+}
+
+// proxy creates a "pipeline" from the underlying response
+// to a WebSocket connection. The pipeline is constructed in
+// the following way:
+//
+//	  underlying response
+//		  -> inMemoryResponseWrite
```
lovromazgon, line 102:
```suggestion
//		 -> inMemoryResponseWriter
```

pkg/foundation/grpcutil/websocket.go
```diff
@@ -0,0 +1,253 @@
...
+func (p *webSocketProxy) startWebSocketWrite(ctx context.Context, messages chan []byte, conn *websocket.Conn, cancelFn func()) {
+	defer func() {
+		for range messages {
+			// throw away
+		}
+	}()
+	defer cancelFn()
+
+	ticker := time.NewTicker(p.pingPeriod)
+	defer ticker.Stop()
```
lovromazgon, line 227: I know I have written this, but now that I look at it this could be nicer 🙂 
```suggestion
	ticker := time.NewTicker(p.pingPeriod)
	defer func() {
		ticker.Stop()
		cancelFn()
		for range messages {
			// throw away
		}
	}()
```

lovromazgon (approved):
Nice, good work!

pkg/foundation/grpcutil/websocket.go
hariso: Makes sense, having a single defer looks nicer. :+1:

PR #751, opened 2022-11-28
parse validations

maha-hajja:
### Description

Parsed validations.
+ change number type to int and float.

Fixes #746 

### Quick checks:

- [ ] I have followed the [Code Guidelines](https://github.com/ConduitIO/conduit/blob/main/docs/code_guidelines.md).
- [ ] There is no other [pull request](https://github.com/ConduitIO/conduit/pulls) for the same update/change.
- [ ] I have written unit tests.
- [ ] I have made sure that the PR is of reasonable size and can be easily reviewed.

PR #753, opened 2022-11-28
WIP: DLQ: internals (2nd part)

### LINKED ISSUE #713: DLQ: Internals
Add internal code needed for forwarding nacked messages to a connector.

lovromazgon:
### Description

TBD

Fixes #713

### Quick checks:


pkg/plugin/standalone/v1/internal/fromproto/specifier.go
```diff
@@ -58,7 +58,7 @@ func SpecifierSpecifyResponse(in *connectorv1.Specifier_Specify_Response) (plugi
 
 func SpecifierParameter(in *connectorv1.Specifier_Parameter) (plugin.Parameter, error) {
 	var validations []plugin.Validation
-	if in.Required {
+	if in.Required { //nolint:staticcheck // we use it to be backwards compatible
```
lovromazgon, line 61: This is here just to make the linter succeed on [this PR](https://github.com/ConduitIO/conduit/pull/751/files#diff-54c07ca8d7960982031b7ee7c488d0061cd54ff8c11d7b801b08032a1572fdf3R91), it will cause a conflict with the PR that properly addresses this. We will address this when merging to main.

pkg/pipeline/dlq.go
```diff
@@ -0,0 +1,99 @@
...
+	stopErr := d.Destination.Stop(ctx, d.lastPosition)
+	if stopErr != nil {
+		defer func() {
+			if err == nil {
+				// replace returned error with stop error
+				err = stopErr
+			}
+		}()
+		// log this error right away because we're not sure the connector
+		// will be able to stop right away, we might block for 1 minute
```
hariso, line 92: Out of curiosity, is the 1 minute mentioned as an example, or is that a value we set somewhere?

pkg/connector/connector.go
```diff
@@ -32,6 +32,7 @@ const (
 const (
 	ProvisionTypeAPI ProvisionType = iota
 	ProvisionTypeConfig
+	ProvisionTypeDLQ // used for provisioning DLQ connectors which are not persisted
```
hariso, line 35: WDYT about having this as a connector type (i.e. we'd have `TypeSource`, `TypeDestination`, and `TypeDLQ`)? It feels like it's more accurate to say that a DLQ is a special type of connector than a special type of provisioning. And DLQs can be provisioned through configuration files and we'll be able to provision them through the API in the future as well.

hariso (approved):
Great work as always!

pkg/pipeline/dlq.go
lovromazgon: Oh, this is referencing the behavior of the SDK ([link](https://github.com/ConduitIO/conduit-connector-sdk/blob/d18c82bdc0e546a2c902a5f1f063cf77ec95720c/destination.go#L220)).

pkg/connector/connector.go
lovromazgon: How many times you comment on something that I considered when working on my code, it's getting scary 😄

I thought about this, in fact, I think I even started with a connector type and changed it to a provision type. That's because the connector itself is still a normal destination connector, it has no idea that it's being used as a DLQ destination. What is different though is how it's provisioned/created - that happens only when the pipeline is actually started, it doesn't exist before like a normal destination connector. So in that sense, it's not provisioned by the config file or by the API, it's provisioned internally by Conduit when the pipeline starts.

lovromazgon: I know what you're saying, let me present a different view though. We introduced the field "provisioned by" to be able to know if a connector was created through the API or config file and to restrict further changes to the connector only through that mechanism. A _DLQ configuration_ works the same way - it's created, updated, deleted only through the API or config file. Notice I said _DLQ configuration_, that's not the same as the DLQ connector. When hitting the API endpoint for updating the DLQ, no connector is provisioned, it's only the configuration that changes. The DLQ connector provisioning is triggered only when the pipeline is started. We could call the type `ProvisionTypeInternal` if that makes it clearer.

The point is that the connector lifecycle is managed internally, it will be created and deleted based on the status of the pipeline. You can't change the _DLQ connector_ through the API or config file, regardless how the _DLQ configuration_ was provisioned, it's a subtle difference.

hariso: That helped shed some light on your view, thanks!

> You can't change the _DLQ connector_ through the API or config file

i.e. a DLQ connector (configuration) is immutable?

lovromazgon: The DLQ configuration is mutable, I'm saying you can't change the connector entity that's created based on that configuration, it only exists internally while the pipeline is running.

pkg/plugin/standalone/v1/internal/fromproto/specifier.go
lovromazgon: PR rebased and conflict addressed.

pkg/connector/connector.go
hariso: I can see your point more clearly now. Personally, I'm not fully on board with overlapping provisioning types, but I also recognize the points you made. :+1:

PR #763, opened 2022-12-05
Stream inspector for processors

### LINKED ISSUE #702: Stream Inspector: Processors
### Feature description

Make it possible to inspect processors. It should be possible to inspect records coming into a processor as well as transformed records.

All types of processors should be covered (source, destination, pipeline processors).

hariso:
### Description

This PR enables inspection for processors and makes the stream inspector complete. Major changes are:
1. Adds `inspect-in` and `inspect-out` endpoints.
2. Adds `InspectIn` and `InspectOut` methods to `processor.Interface`.

Fixes #702 

### Quick checks:


proto/api/v1/api.proto
```diff
@@ -794,11 +802,25 @@ service ProcessorService {
 message ListProcessorsRequest {
   repeated string parent_ids = 1;
 }
-
 message ListProcessorsResponse {
   repeated Processor processors = 1;
 }
 
+message InspectProcessorRequest {
+  enum Direction {
```
hariso, line 810: I wasn't sure what's better: direction or type, but now I'm leaning more towards "type". It doesn't feel right to say that an inspection has a direction, but it can be of different types.

pkg/web/openapi/swagger-ui/api/v1/api.swagger.json
```diff
@@ -1261,6 +1261,65 @@
           "ProcessorService"
         ]
       }
+    },
+    "/v1/processors/{id}/inspect": {
+      "get": {
+        "summary": "todo decide whether to use the direction as path param or query param",
```
hariso, line 1267: Will remove it.

proto/api/v1/api.proto
```diff
@@ -669,6 +669,20 @@ service ProcessorService {
...
+      response_body: "record"
+    };
+  }
+
+  rpc InspectProcessorOut(InspectProcessorOutRequest) returns (stream InspectProcessorOutResponse) {
+    option (google.api.http) = {
+      get: "/v1/processors/{id}/inspect-out"
+      response_body: "record"
+    };
+  }
```
hariso, line 684: We have 3 options to specify which records we want to inspect:
1. query params, e.g. `/v1/processors/{id}/inspect?type=in`
2. path params, e.g. `/v1/processors/{id}/inspect/{type}`
3. different endpoints (as implemented)

Query params would mean a sane default exists, which doesn't apply in our case. 

Path params vs different endpoints: Usage-wise, pretty similar. Different endpoints are less work (no type validation). It also makes it clear that only these two types are supported (everything else is a 404 error code). Also, path params indicate there are two `inspect` resources (with IDs `in` and `out`), but `inspect` here is meant as a verb (to start inspecting).

pkg/processor/procbuiltin/func_wrapper.go
```diff
@@ -0,0 +1,60 @@
...
+	"context"
+
+	"github.com/conduitio/conduit/pkg/foundation/log"
+	"github.com/conduitio/conduit/pkg/inspector"
+	"github.com/conduitio/conduit/pkg/record"
+	"github.com/rs/zerolog"
+)
+
+// FuncWrapper is an adapter allowing use of a function as an Interface.
+type FuncWrapper struct {
```
hariso, line 27: Previously `InterfaceFunc`. I've moved it to the `procbuiltin` package, because this is virtually the only place where it's used (plus a couple of tests).

docs/processors.md
```diff
@@ -163,3 +163,15 @@ function process(record) {
...
 }
 ```
+
+## Inspecting a processor
+Records entering a processor and the resulting records can be inspected using 
+[Conduit's stream inspector](/docs/design-documents/20221024-stream-inspector.md). This makes it possible to "debug" a 
+processor.
+
+The records are made available via a WebSocket connection. To inspect the input records, following endpoint needs to be 
+used: `ws://host:port/v1/processors/{processor-id}/inspect-in`. Similarly, to inspect the output records, following
```
lovromazgon, line 173: This sounds better to me (disclaimer: not a native english speaker 🥴)
```suggestion
The records are made available via a WebSocket connection. To inspect input records, the following endpoint needs to be 
used: `ws://host:port/v1/processors/{processor-id}/inspect-in`. Similarly, to inspect output records, the following 
```

proto/api/v1/api.proto
lovromazgon: I'm not sure which approach is better, none of them strike me as particularly nice tbh, but I don't have a better proposal. I slightly lean towards option 2, although I have nothing to back up my preference except it "looks nicer" 😅

Another thing to think about is changing in/out to incoming/outgoing, although I don't feel strongly about that either. I don't mind if you leave everything as it is now, just throwing out more ideas as food for thought 😉

proto/api/v1/api.proto
lovromazgon: Looks simpler now without the direction enum.

hariso: Yup!

proto/api/v1/api.proto
hariso: I do agree that option 2 looks nicer, but I think that it makes it less REST-like. It would indicate there's an `inspect` resource type and two different resources with IDs `in` and `out` (or `incoming` and `outgoing`). But again, I find it definitely prettier.:)

`incoming` and `outgoing` also sound good to me. I lean a bit more towards `in` and `out` because it's shorter, but still clear what it's about.

pkg/processor/procjs/processor.go
```diff
@@ -46,10 +48,15 @@ type jsRecord struct {
...
 	runtime  *goja.Runtime
 	function goja.Callable
+	inInsp   *inspector.Inspector
+	outInsp  *inspector.Inspector
 }
 
 func New(src string, logger zerolog.Logger) (*Processor, error) {
-	p := &Processor{}
+	p := &Processor{
+		inInsp:  inspector.New(log.New(logger), 1000),
```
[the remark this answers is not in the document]
hariso: Agreed!

pkg/processor/procbuiltin/func_wrapper.go
```diff
@@ -0,0 +1,60 @@
...
+		inInsp:  inspector.New(log.New(zl), 1000),
+		outInsp: inspector.New(log.New(zl), 1000),
+	}
+}
+
+func (f FuncWrapper) Process(ctx context.Context, inRec record.Record) (record.Record, error) {
+	// todo same behavior as in procjs, probably can be enforced
+	f.inInsp.Send(ctx, inRec)
+	outRec, err := f.f(ctx, inRec)
+	f.outInsp.Send(ctx, outRec)
```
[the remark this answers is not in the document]
hariso: Yup, that makes sense.

pkg/processor/procbuiltin/func_wrapper.go
```diff
@@ -38,23 +38,25 @@ func NewFuncWrapper(f func(context.Context, record.Record) (record.Record, error
...
 }
 
 func (f FuncWrapper) Process(ctx context.Context, inRec record.Record) (record.Record, error) {
 	// todo same behavior as in procjs, probably can be enforced
 	f.inInsp.Send(ctx, inRec)
 	outRec, err := f.f(ctx, inRec)
-	f.outInsp.Send(ctx, outRec)
+	if err == nil {
+		f.outInsp.Send(ctx, outRec)
+	}
```
lovromazgon, line 52: Nitpick: I'd prefer keeping the happy flow to the left, do you mind turning the condition around? 🙏

lovromazgon (approved):
🚀

pkg/processor/procbuiltin/func_wrapper.go
hariso: Good one!

PR #771, opened 2022-12-07
DLQ config parsing

### LINKED ISSUE #716: DLQ: Pipeline config file
Update pipeline config parser to parse and provision DLQ. Bump config version because of the change.

lovromazgon:
### Description

This PR changes the config parser to also parse DLQ configs.

Fixes #716

### Quick checks:


pkg/pipeline/service.go
```diff
@@ -193,6 +193,36 @@ func (s *Service) Update(ctx context.Context, pipelineID string, cfg Config) (*I
...
+	pl, err := s.Get(ctx, pipelineID)
+	if err != nil {
+		return nil, err
+	}
+
+	if cfg.Plugin == "" {
+		return nil, cerrors.New("DLQ plugin must be provided")
+	}
+	if cfg.WindowSize < 0 {
+		return nil, cerrors.New("DLQ window size must be non-negative")
```
maha-hajja, line 207: what does it mean if window size is 0?

pkg/provisioning/parse.go
```diff
@@ -44,6 +44,14 @@ type PipelineConfig struct {
 	Description string                     `yaml:"description"`
 	Connectors  map[string]ConnectorConfig `yaml:"connectors,omitempty"`
 	Processors  map[string]ProcessorConfig `yaml:"processors,omitempty"`
+	DLQ         DLQConfig                  `yaml:"dead-letter-queue"`
+}
+
+type DLQConfig struct {
+	Plugin              string            `yaml:"plugin"`
+	Settings            map[string]string `yaml:"settings"`
+	WindowSize          *int              `yaml:"window-size"`
```
maha-hajja, line 53: is there a specific reason to have it as a pointer?

pkg/pipeline/service.go
lovromazgon: Window size 0 means that no records are monitored, all nacked records are rerouted to the DLQ without a limit so the pipeline doesn't stop.

pkg/provisioning/parse.go
lovromazgon: Yes, setting the window size to 0 is a valid thing to do, while if it's not set the default window size is 1, so we have to be able to differentiate if the window size was explicitly set to 0 or if it is empty.

maha-hajja: oh makes sense! thanks for the explanation

PR #772, opened 2022-12-07
Pipeline config changelog and linter

lovromazgon:
### Description

This PR adds a way to define the changelog for the pipeline config file that is used in a linter that produces warnings based on deprecated fields and/or new fields that officially don't exist in that config version yet.

Right now the following warnings are produced:
- warn when version of file is unknown
- warn when a field is found that is unknown (field will be ignored)
- warn when a field is defined that was introduced in a higher version than what is specified in the file (even if parser knows how to use it)
- warn when a field is deprecated

Depends on https://github.com/ConduitIO/conduit/pull/771 ~and https://github.com/ConduitIO/yaml/pull/3 (tests won't pass until this is merged)~.

### Quick checks:

### DIFF
--- a/pkg/provisioning/changelog.go
+++ b/pkg/provisioning/changelog.go
@@ -71,6 +71,12 @@ func expandChangelog(changelog map[string][]change) map[string]map[string]interf
 		knownChanges[v.Original()] = make(map[string]interface{})
 	}
 
+	// addChange stores change c in map m by splitting c.field into multiple
+	// tokens and storing it in m[token1][token2][...][tokenN]. If any map in
+	// that hierarchy does not exist it is created. If any value in that
+	// hierarchy exists and is _not_ a map it is _not_ replaced. This means that
+	// changes related to parent fields take precedence over changes related to
+	// child fields.
 	addChange := func(c change, m map[string]interface{}) {
 		tokens := strings.Split(c.field, ".")
 		curMap := m
@@ -99,12 +105,14 @@ func expandChangelog(changelog map[string][]change) map[string]map[string]interf
 		for _, v2 := range versions {
 			switch {
 			case !v.GreaterThan(v2):
+				// warn about deprecated fields in future versions
 				for _, c := range changes {
 					if c.changeType == fieldDeprecated {
 						addChange(c, knownChanges[v2.Original()])
 					}
 				}
 			case v.GreaterThan(v2):
+				// warn about introduced fields in older versions
 				for _, c := range changes {
 					if c.changeType == fieldIntroduced {
 						addChange(c, knownChanges[v2.Original()])
--- a/pkg/provisioning/changelog.go
+++ b/pkg/provisioning/changelog.go
@@ -1,0 +1,117 @@
+// Copyright © 2022 Meroxa, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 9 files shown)


pkg/provisioning/changelog.go
```diff
@@ -0,0 +1,117 @@
...
+		versions = append(versions, semver.MustParse(k))
+	}
+	sort.Sort(versions)
+
+	knownChanges := make(map[string]map[string]interface{})
+	for _, v := range versions {
+		knownChanges[v.Original()] = make(map[string]interface{})
+	}
+
+	addChange := func(c change, m map[string]interface{}) {
```
hariso, line 74: Nitpick: When reading this for the first time, it would help to explain what `m` is.

pkg/provisioning/changelog.go
```diff
@@ -0,0 +1,117 @@
...
+// changeType defines the type of the change introduced in a specific version.
+type changeType int
+
+const (
+	fieldDeprecated changeType = iota
+	fieldIntroduced
+)
+
+// expandedChangelog is populated in init based on changelog. This map contains
+// the changelog in a structure that is useful for traversing in configLinter.
```
hariso, line 55: Maybe mention an example of how it should look like? I think it would help understand the code better.

pkg/provisioning/changelog.go
```diff
@@ -0,0 +1,117 @@
...
+	// that uses this field with an unsupported version.
+	message string
+}
+
+// changeType defines the type of the change introduced in a specific version.
+type changeType int
+
+const (
+	fieldDeprecated changeType = iota
+	fieldIntroduced
```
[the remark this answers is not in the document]
maha-hajja: isn't deleted the same as deprecated? or do you mean deleted fields that are not supported anymore?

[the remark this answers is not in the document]
lovromazgon: Deleting a field would be a breaking change, ideally we shouldn't do that. We can deprecate a field and warn the user if they use it, but at the same time keep the behavior of that field if we encounter it. That's at least the current idea.

pkg/provisioning/changelog.go
lovromazgon: I should probably write a test for this, I somehow missed it. That's at least what I would write in a review comment if I saw this 😄 I hope this will count as providing an example.

pkg/provisioning/changelog.go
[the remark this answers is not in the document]
hariso: Agreed that deleted configuration fields are not nice to work with. It would be nice if we write down our philosophy somewhere, even if it's a comment in the code.

pkg/provisioning/changelog_test.go
```diff
@@ -0,0 +1,129 @@
...
+			// pipelines was deprecated in 1.4 so it takes precedence
+			field:      "pipelines.*.name",
+			changeType: fieldDeprecated,
+			message:    "field name was deprecated in version 1.2, please use \"title\" instead",
+		}},
+		"1.3": {}, // here just to test deprecated field
+		"1.4": {{
+			field:      "pipelines",
+			changeType: fieldDeprecated,
+			message:    "field pipelines was deprecated in version 1.4, please don't create pipelines anymore",
```
hariso, line 57: Is this a hint at something? :smile:

hariso (approved):

pkg/provisioning/changelog_test.go
lovromazgon: Who knows 🔮 You'll just have to wait and see!

pkg/provisioning/changelog.go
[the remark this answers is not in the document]
lovromazgon: I'd argue that this is general advice we should follow in any publicly facing interface, from the API, the SDK interface, connector protocol, pipeline config files, CLI flags etc. For instance we did the same with the [required field](https://github.com/ConduitIO/conduit-connector-sdk/blob/e74f759f151935a9d6215eca3cef4fa7d503964f/specifier.go#L45-L47). It's just the way semantic versioning works, so I'm reluctant to add this comment here, and I'm not sure where the best place for this would be.

PR #774, opened 2022-12-08
Add conduit-pipeline-check utility

### LINKED ISSUE #773: Feature: command line tool for health checking pipelines
### Feature description

Command line tool which checks the health of an existing pipeline. 
The tool will return non-zero status when the pipeline is not running or does not exist.

It can be used in automation systems, such as kubernetes to health check conduit deployments.

lyuboxa:
### Description

The command line utility allows users or automation to request status of a pipeline and return a non-zero code when the pipeline is not running. The check happens over gRPC.

Additionally include a .dockerignore file to speed up docker builds.

Fixes #773 

### Quick checks:

### DIFF
--- a/cmd/conduit-pipeline-check/main.go
+++ b/cmd/conduit-pipeline-check/main.go
@@ -1,0 +1,118 @@
+// Copyright © 2022 Meroxa, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package main
+
+import (
+	"context"
+	"flag"
+	"fmt"
+	"os"
+	"os/signal"
+	"time"
+
+	apiv1 "github.com/conduitio/conduit/proto/gen/api/v1"
+	"google.golang.org/grpc"
+	"google.golang.org/grpc/credentials/insecure"
+)
+
+const (
+	exitCodeOk = iota
+	exitCodeErr
+	exitCodeInterrupt
+)
+
+func main() {
+	ctx := cancelOnInterrupt(context.Background())
+
+	exitcode, err := checkPipeline(ctx)
+	if err != nil {
+		fmt.Fprintln(os.Stderr, err)
+	}
+	os.Exit(exitcode)
+}
+
+func checkPipeline(ctx context.Context) (int, error) {
+	flags := flag.NewFlagSet(os.Args[0], flag.ExitOnError)
+	var (
+		grpcAddress  = flags.String("grpc.address", ":8084", "address of the Conduit server gRPC API")
+		pipelineName = flags.String("pipeline.name", "", "name of the target pipeline")
+		timeout      = flags.Duration("timeout", 10*time.Second, "timeout duration")
+		verbose      = flags.Bool("verbose", false, "return non-zero code on failure instead of logging")
+	)
+
+	_ = flags.Parse(os.Args[1:])
+
+	if pipelineName == nil || *pipelineName == "" {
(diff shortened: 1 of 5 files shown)


cmd/conduit-pipeline-check/main.go
```diff
@@ -0,0 +1,118 @@
...
+	opts := []grpc.DialOption{
+		grpc.WithTransportCredentials(insecure.NewCredentials()),
+	}
+
+	dialCtx, dialCancel := context.WithTimeout(ctx, *timeout)
+	defer dialCancel()
+
+	c, err := grpc.DialContext(dialCtx, *grpcAddress, opts...)
+	if err != nil {
+		return exitCodeErr, fmt.Errorf("failed to connect to conduit grpc server: %v", err)
```
hariso, line 70: Nitpick: %v should be %w

cmd/conduit-pipeline-check/main.go
```diff
@@ -0,0 +1,118 @@
...
+	defer c.Close()
+
+	pipeline := apiv1.NewPipelineServiceClient(c)
+	p, err := pipeline.GetPipeline(
+		ctx,
+		&apiv1.GetPipelineRequest{
+			Id: *pipelineName,
+		})
+	if err != nil {
+		return exitCodeErr, fmt.Errorf("failed to find pipeline %q: %v", *pipelineName, err)
```
hariso, line 81: %v -> %w

lyuboxa: fixed in [5cb1a8](https://github.com/ConduitIO/conduit/pull/774/commits/75cb1a895fb83bebefc230d4ef24f7d73e4ab118)

cmd/conduit-pipeline-check/main.go
lyuboxa: fixed in [5cb1a8](https://github.com/ConduitIO/conduit/pull/774/commits/75cb1a895fb83bebefc230d4ef24f7d73e4ab118)

cmd/conduit-pipeline-check/main.go
```diff
@@ -0,0 +1,123 @@
...
+// * missing or invalid flags
+// * gRPC connection failures
+// * operation takes too long to execute
+func checkPipeline(ctx context.Context) (int, error) {
+	flags := flag.NewFlagSet(os.Args[0], flag.ExitOnError)
+	var (
+		grpcAddress  = flags.String("grpc.address", ":8084", "address of the Conduit server gRPC API")
+		pipelineName = flags.String("pipeline.name", "", "name of the target pipeline")
+		timeout      = flags.Duration("timeout", 10*time.Second, "timeout duration")
+		verbose      = flags.Bool("verbose", false, "return non-zero code on failure instead of logging")
```
lovromazgon, line 57: IIUC `verbose` will make sure a log is printed _in addition_ to the exit code, right? The message seems a bit misleading.

cmd/conduit-pipeline-check/main.go
```diff
@@ -0,0 +1,123 @@
...
+)
+
+func main() {
+	ctx := cancelOnInterrupt(context.Background())
+
+	exitcode, err := checkPipeline(ctx)
+	if err != nil {
+		fmt.Fprintln(os.Stderr, err)
+	}
+	os.Exit(exitcode)
```
lovromazgon, line 43: Here `exitcode` will always be `exitCodeErr` if an error is returned or `exitCodeOk` if it's not, right? We could simplify this:
```suggestion
	err := checkPipeline(ctx)
	if err != nil {
		fmt.Fprintln(os.Stderr, err)
		os.Exit(exitCodeErr)
	}
```
The nice thing about this is we don't call `os.Exit(0)` which just terminates the program right away without calling any deferred functions. Not that it matters here, but you know, good practice 🙃

lovromazgon (approved):
LGTM, just a few nitpicks 😉

cmd/conduit-pipeline-check/main.go
lyuboxa: yup this makes sense. will make the change, originally i started with just an error as well.

cmd/conduit-pipeline-check/main.go
lyuboxa: ugh yea, will update

cmd/conduit-pipeline-check/main.go
lyuboxa: fixed in bc01d7c425731ccb18893dd3447adcec34012f64

PR #785, opened 2022-12-15
Skip connector validation when initializing connector service

lovromazgon:
### Description

This fixes the bug where an invalid connector prevents Conduit from starting.

This is a backport to v0.2.1, a similar fix will be created for the latest version.


samirketema (approved):
looks great, thanks for the speedy fix!

PR #790, opened 2022-12-21
DLQ API

### LINKED ISSUE #715: DLQ: API
Update the pipeline grpc service to include methods for retrieving and updating the DLQ config. The new methods should be mapped to these endpoints:

- `PUT /v1/pipelines/{id}/dead-letter-queue`
- `GET /v1/pipelines/{id}/dead-letter-queue`

Make sure that the DLQ config is validated, including plugin and settings (same as in a normal connector).

lovromazgon:
### Description

Adds a way to get info about a DLQ and update it through the API.

Depends on #789.

Fixes #715

### Quick checks:


pkg/orchestrator/pipelines.go
```diff
@@ -80,5 +81,37 @@ func (s *PipelineOrchestrator) Delete(ctx context.Context, id string) error {
...
+		return nil, cerrors.Errorf("pipeline %q cannot be updated: %w", pl.ID, ErrImmutableProvisionedByConfig)
+	}
+	// TODO lock pipeline
+	if pl.Status == pipeline.StatusRunning {
+		return nil, pipeline.ErrPipelineRunning
+	}
+
+	// cast orchestrator to a ConnectorOrchestrator to get access to the plugin
+	// config validate function
+	err = (*ConnectorOrchestrator)(po).Validate(
```
hariso, line 103: Why does the cast work btw?

pkg/orchestrator/pipelines_test.go
```diff
@@ -302,5 +313,140 @@ func TestPipelineOrchestrator_Delete_PipelineDoesntExist(t *testing.T) {
...
+
+	orc := NewOrchestrator(db, log.Nop(), plsMock, consMock, procsMock, pluginMock)
+	plsMock.EXPECT().
+		Get(gomock.AssignableToTypeOf(ctxType), plBefore.ID).
+		Return(plBefore, nil)
+
+	got, err := orc.Pipelines.UpdateDLQ(ctx, plBefore.ID, pipeline.DLQ{})
+	is.Equal(got, nil)
+	is.True(err != nil)
+	is.Equal(pipeline.ErrPipelineRunning, err)
```
hariso, line 390: The check `is.Equal(pipeline.ErrPipelineRunning, err)` also implies `is.True(err != nil)`, right?

hariso (commented):
As always, great work!

pkg/orchestrator/pipelines.go
lovromazgon: All orchestrators have the same underlying type [`base`](https://github.com/ConduitIO/conduit/blob/54025b01872bb2dccc5e133fd6ae8f0d7e719552/pkg/orchestrator/orchestrator.go#L62), they just define different methods on top. See how the orchestrators are created [here](https://github.com/ConduitIO/conduit/blob/54025b01872bb2dccc5e133fd6ae8f0d7e719552/pkg/orchestrator/orchestrator.go#L55-L58).

Here's a simplified example of what's happening: https://go.dev/play/p/aQQRxZaMnHC

pkg/orchestrator/pipelines_test.go
lovromazgon: It does, it's a bit redundant 🤫

pkg/orchestrator/pipelines_test.go
```diff
@@ -132,11 +137,12 @@ func TestPipelineOrchestrator_Update_Success(t *testing.T) {
 		Return(want, nil)
 
 	got, err := orc.Pipelines.Update(ctx, plBefore.ID, newConfig)
-	assert.Equal(t, got, want)
-	assert.Ok(t, err)
+	is.Equal(got, want)
+	is.NoErr(err)
```
maha-hajja, line 141: I think it's better to check for the error first, so that if it fails, it shows that there's an error, instead of saying that got!=want. WDYT?

pkg/orchestrator/pipelines_test.go
```diff
@@ -302,5 +313,140 @@ func TestPipelineOrchestrator_Delete_PipelineDoesntExist(t *testing.T) {
...
+	newDLQ := pipeline.DLQ{
+		Plugin:              "new-plugin",
+		Settings:            map[string]string{"baz": "qux"},
+		WindowSize:          2,
+		WindowNackThreshold: 0,
+	}
+	want := &pipeline.Instance{
+		ID:     plBefore.ID,
+		Status: pipeline.StatusSystemStopped,
+		Config: pipeline.Config{Name: "test-pipeline"},
```
maha-hajja, line 348:
```suggestion
		Status: plBefore.Status,
		Config: plBefore.Config,
```

pkg/orchestrator/pipelines.go
```diff
@@ -80,5 +81,37 @@ func (s *PipelineOrchestrator) Delete(ctx context.Context, id string) error {
...
+		ctx,
+		connector.TypeDestination,
+		dlq.Plugin,
+		connector.Config{
+			Name:     "temporary-dlq",
```
maha-hajja, line 108: can we add a clarifying comment here plz? about why this name is not important, cuz I was confused at first when I saw it 👀

maha-hajja (approved):
lgtm

PR #794, opened 2022-12-23
DLQ Metrics

### LINKED ISSUE #721: DLQ: Metrics
Add metrics to monitor a pipeline DLQ behavior:

- `conduit_dlq_bytes` - same as `conduit_connector_bytes` but for DLQs
- `conduit_dlq_execution_duration_seconds` - same as `conduit_connector_execution_duration_seconds` but for DLQs

lovromazgon:
### Description

This PR introduces metrics `conduit_dlq_execution_duration_seconds` and `conduit_dlq_bytes`. Internally it introduces `metrics.RecordBytesHistogram` which is a utility to observe record sizes (previously this code was in `stream.MetricsNode`).

Fixes #721 

### Quick checks:

- [ ] I have written unit tests. (well existing tests pass)


pkg/foundation/metrics/measure/measure.go
```diff
@@ -41,6 +41,12 @@ var (
 		// buckets from 1KiB to 2MiB
 		prometheus.HistogramOpts{Buckets: []float64{1024, 1024 << 1, 1024 << 2, 1024 << 3, 1024 << 4, 1024 << 5, 1024 << 6, 1024 << 7, 1024 << 8, 1024 << 9, 1024 << 10, 1024 << 11}},
 	)
+	DLQBytesHistogram = metrics.NewLabeledHistogram("conduit_dlq_bytes",
+		"Number of bytes a DLQ connector processed per pipeline and plugin.",
```
hariso, line 45: Nitpick: it would be nice to have a few words about what these bytes are (similar to what you already did in RecordBytesHistogram).

hariso (approved):
Looks great to me! I have one question and one suggestion:
1. WDYT about adding metrics related to the number of records which went through a DLQ (if we don't have that somewhere)?
2. It would be nice to mention these metrics in docs/metrics.md and maybe also docs.conduit.io

pkg/foundation/metrics/measure/measure.go
lovromazgon: Do you mean the help text? I wouldn't add too much info there, looking at existing metrics these help texts are kept very brief. If someone is interested in the details they can have a look at the docs.

That said, I agree that we should mention how we calculate the record bytes in our metrics docs, I'll do that in a separate PR!

lovromazgon: https://github.com/ConduitIO/conduit/pull/829

Issue #793, opened 2022-12-22
Bug: UI does not populate enum parameter


lovromazgon:
### Bug description

If a parameter is an enum (i.e. contains an inclusion validation) the current value is not populated correctly.

### Steps to reproduce

https://user-images.githubusercontent.com/8320753/209168820-a0e69081-7212-49c3-a043-38bdca5dcfb1.mov

### Version

v0.4.0

PR #805, opened 2023-01-06
fix(ui): use field value w inclusion when present

jmar910:
### Description
Fixes https://github.com/ConduitIO/conduit/issues/793

Ensures if a value is already set in the configurable object, we populate it in select input config fields

### Quick checks:


jayjayjpg (approved):
This reads great and has nice test coverage as well! :sparkles:

PR #811, opened 2023-01-09
[UI] Ember 4 Deprecations

jmar910:
### Description

Clears...
+ mirage config deprecation 
+ replace route deprecation
+ runloop deprecation via flash messages upgrade
+ implicit serve injections deprecation
+ ember keyboard deprecation via mx-ui-components upgrade
+ unload record deprecation 
+ modifier arg deprecation

### Quick checks:

### Verification
+ Tests pass
+ No deprecation log messages in test output
+ General smoke test of app (confirmed locally in my testing)

### DIFF
--- a/ui/config/deprecation-workflow.js
+++ b/ui/config/deprecation-workflow.js
@@ -2,18 +2,5 @@
 self.deprecationWorkflow = self.deprecationWorkflow || {};
 
 self.deprecationWorkflow.config = {
-  workflow: [
-    { handler: 'silence', matchId: 'ember-keyboard.first-responder-inputs' },
-    { handler: 'silence', matchId: 'ember-test-waiters-legacy-module-name' },
-    { handler: 'silence', matchId: 'ember-keyboard.old-propagation-model' },
-    {
-      handler: 'silence',
-      matchId: 'ember.built-in-components.legacy-arguments',
-    },
-
-    {
-      handler: 'silence',
-      matchId: 'ember-modifier.function-based-options',
-    },
-  ],
+  workflow: [],
 };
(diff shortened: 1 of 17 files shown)


ui/config/deprecation-workflow.js
```diff
@@ -3,20 +3,17 @@ self.deprecationWorkflow = self.deprecationWorkflow || {};
...
-    { handler: 'silence', matchId: 'ember-global' },
-    { handler: 'silence', matchId: 'ember.component.reopen' },
     { handler: 'silence', matchId: 'ember-keyboard.first-responder-inputs' },
-    { handler: 'silence', matchId: 'implicit-injections' },
     { handler: 'silence', matchId: 'ember-test-waiters-legacy-module-name' },
     { handler: 'silence', matchId: 'ember-keyboard.old-propagation-model' },
     {
       handler: 'silence',
-      matchId: 'deprecated-run-loop-and-computed-dot-access',
+      matchId: 'ember.built-in-components.legacy-arguments',
```
brettgoulder, line 11: should this still be silenced?

brettgoulder (approved):
approving with just one small comment

ui/config/deprecation-workflow.js
brettgoulder: I meant, should we throw here instead of silence?

jmar910: I wouldnt recommend setting these to throw unless you are working locally and trying to see the errors halt your test suite which is what I did. If you miss a spot with a deprecation and set it to throw, it can blow the app on code that works perfectly fine (the case here since I am not upgrading to ember 4 in this PR)

Deprecations will still be logged from ember even with these entries set to silence or not even here, they will just be spewed and not grouped.

PR #826, opened 2023-01-19
add parsejson and unwraprecord processors

### LINKED ISSUE #817: Feature: Record debezium unwrapping processor
### Feature description

We need to be able to unwrap a Debezium record into an opencdc record.

This issue has two parts - one processor needs to be able to parse [`RawData`](https://github.com/ConduitIO/conduit/blob/309373cb54f320baab8e6235f53d0fb71f6bac55/pkg/record/record.go#L197) and convert it into [`StructuredData`](https://github.com/ConduitIO/conduit/blob/309373cb54f320baab8e6235f53d0fb71f6bac55/pkg/record/record.go#L184), a second one needs to be able to take the contents of [`Record.Payload.After`](https://github.com/ConduitIO/conduit/blob/309373cb54f320baab8e6235f53d0fb71f6bac55/pkg/record/record.go#L160) and assign create a `Record` out of them.

### Part 1

When the Kafka connector reads a message it stuffs the key into [`Record.Key`](https://github.com/ConduitIO/conduit/blob/309373cb54f320baab8e6235f53d0fb71f6bac55/pkg/record/record.go#L88) and payload into [`Record.Payload.After`](https://github.com/ConduitIO/conduit/blob/309373cb54f320baab8e6235f53d0fb71f6bac55/pkg/record/record.go#L160), both as [`RawData`](https://github.com/ConduitIO/conduit/blob/309373cb54f320baab8e6235f53d0fb71f6bac55/pkg/record/record.go#L197). Other connectors reading from an unstructured source work the same.

We need a transform that is able to parse that data as JSON and create structured data out of it.

### Part 2

When we have structured data in a payload it might be that the payload itself contains a record that has its own metadata, payload etc. For now we are focusing on a Debezium record, but we need the same for OpenCDC in the future as well.

We need a processor that can "unwrap" that record, i.e. take the contents of [`Record.Payload.After`](https://github.com/ConduitIO/conduit/blob/309373cb54f320baab8e6235f53d0fb71f6bac55/pkg/record/record.go#L160) and create a `Record` out of them. There are different possible approaches, right now we are looking to simply extract the record and return it instead of the wrapped record.

**Important!** The processor should specifically work on a Debezium record and use the same mapping we use in the SDK when creating a Debezium record ([link](https://github.com/ConduitIO/conduit-connector-sdk/blob/6cd5101ba73f7388dfea47dc188692d27845d455/record.go#L252-L259)), but in the other direction. The schema can be removed from the record (Conduit currently doesn't use schemas).

In the future we might want to give the user more flexibility on how the record gets unwrapped, since the approach above causes us to lose any info we had in the wrapped record (e.g. metadata). Thinking out loud - we could give the user fine-grained control over which payload field gets written into which Record field (e.g. `record.payload.after.foo>record.metadata`), the separator between them could control if the target field gets overwritten (`>`) or appended (`>>`). But this should not be part of this issue.

maha-hajja:
### Description

Fixes #817

### Quick checks:


pkg/processor/procbuiltin/unwraprecord.go
```diff
@@ -0,0 +1,171 @@
...
+
+		// convert json data into a record
+		before := convertJSONToStructuredData(payload[DebeziumFieldBefore])
+		after := convertJSONToStructuredData(payload[DebeziumFieldAfter])
+		metadata := convertJSONToMetadata(payload[DebeziumFieldSource])
+
+		r.Payload.Before = before
+		r.Payload.After = after
+		r.Metadata = metadata
+		r.Operation = getRecordOp(payload[DebeziumFieldOp].(string))
```
samirketema, line 96: just leaving a note here from our pairing:
- need to include the `r.Position`
- need to include the `r.Key`, pending the approach we decide on for the key

pkg/processor/procbuiltin/unwraprecord.go
```diff
@@ -0,0 +1,174 @@
...
+
+func unwrapRecord(
+	processorType string,
+	getSetter recordDataGetSetter,
+	config processor.Config,
+) (processor.Interface, error) {
+	var (
+		err    error
+		format string
+		r      record.Record
```
lovromazgon, line 54: If the processor is called from multiple goroutines we will have issues with concurrent writes to `r` and `err` (this might happen if we introduce [parallel processors](https://github.com/ConduitIO/conduit/pull/744)). That's why we should remove these two variables and rather make them local to the processor function.

lovromazgon (commented):
Most of my comments are regarding the unwrap processor, which is not that easy to get right. A couple of thoughts that came to my mind while reviewing:
- The processor always works on the field `payload.after`. I was thinking - is there a case when the wrapped record will actually be in `payload.before`? Given that Debezium records are also CDC records we don't expect them to get updated or deleted, they are simply created, which means that the OpenCDC record carrying a Debezium record is expected to have the operation "created" and contain data in `payload.after` so this seems fine. I'd just like us to document that this is a conscious decision.
- The processor will throw away metadata in the original record. I think it would be better to keep the metadata and copy over the Debezium metadata. If the user wants to drop the metadata they can use another processor (we need to create such a processor, but that's a separate issue).
- Do we need a similar processor for the key? The key can also contain a wrapped debezium record, although in our use case we will probably just need to use `extractfieldkey` and extract `payload` so this might not be needed.

pkg/processor/procbuiltin/unwraprecord.go
```diff
@@ -0,0 +1,174 @@
...
+	return processor.InterfaceFunc(func(_ context.Context, rec record.Record) (record.Record, error) {
+		data := getSetter.Get(rec) // record.payload.after
+		var jsonData record.StructuredData
+		switch d := data.(type) {
+		case record.RawData:
+			// unmarshal raw data to structured
+			err = json.Unmarshal(data.Bytes(), &jsonData)
+			if err != nil {
+				return record.Record{}, cerrors.Errorf("%s: %w", processorType, err)
+			}
```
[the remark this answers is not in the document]
maha-hajja: but this would mean that processors order is important, which is hard for platform to do, that's why I have this here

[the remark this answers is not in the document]
lovromazgon: Hmm, right, I forgot about that. Then let's leave it like this for now, I still think we should remove it once platform can properly order processors and use two separate processors.

pkg/processor/procbuiltin/unwraprecord.go
```diff
@@ -0,0 +1,174 @@
...
+	return multiErr
+}
+
+func convertJSONToStructuredData(mp any) record.StructuredData {
+	if mp == nil {
+		return nil
+	}
+
+	structured := record.StructuredData{}
+	for k, v := range mp.(map[string]any) {
```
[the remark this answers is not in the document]
maha-hajja: I tried casting it directly, but it complains about the left side not being an interface, because it doesn't implement `Process`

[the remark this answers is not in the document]
lovromazgon: `Process`? 🤔 I must be missing something, I'm saying you could do something like this:
```
if mpMap, ok := mp.(map[string]any); ok {
	mpStructuredData := record.StructuredData(mpMap)
}
```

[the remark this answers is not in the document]
maha-hajja: oh thought you meant casting the whole map into StructuredData without looping through it.. this is helpful thanks!

pkg/processor/procbuiltin/parsejson.go
```diff
@@ -0,0 +1,72 @@
...
+		case record.RawData:
+			var jsonData record.StructuredData
+			err := json.Unmarshal(data.Bytes(), &jsonData)
+			if err != nil {
+				return record.Record{}, cerrors.Errorf("%s: %w", processorType, err)
+			}
+			r = getSetter.Set(r, jsonData)
+
+		case record.StructuredData:
+			return record.Record{}, cerrors.Errorf("%s: the data is already structured", processorType)
```
hariso, line 64: I'm wondering if we should have an error here. The goal of `parsejson` is to have structured data. If it's already there, we're already at the goal.

pkg/processor/procbuiltin/parsejson_test.go
```diff
@@ -0,0 +1,139 @@
...
+	"github.com/google/go-cmp/cmp"
+	"github.com/matryer/is"
+)
+
+func TestParseJsonKey_Process(t *testing.T) {
+	is := is.New(t)
+
+	type args struct {
+		r record.Record
+	}
```
hariso, line 33: If it's only one field (and we do not plan to add more), it's simpler to just use a record directly.

pkg/processor/procbuiltin/parsejson_test.go
```diff
@@ -0,0 +1,139 @@
...
+	tests := []struct {
+		name    string
+		args    args
+		want    record.Record
+		wantErr bool
```
hariso, line 38: It would be good to also check what the error is (not only if the error is present).

pkg/processor/procbuiltin/parsejson_test.go
```diff
@@ -0,0 +1,139 @@
...
+				t.Errorf("process() error = %v, wantErr = %v", err, tt.wantErr)
+				return
+			}
+			if diff := cmp.Diff(tt.want, got); diff != "" {
+				t.Errorf("process() diff = %s", diff)
+			}
+		})
+	}
+}
+func TestParseJsonPayload_Process(t *testing.T) {
```
hariso, line 80: Nitpick: Newline above the test method.

pkg/processor/procbuiltin/parsejson_test.go
```diff
@@ -0,0 +1,139 @@
...
+	"fmt"
+	"testing"
+
+	"github.com/conduitio/conduit/pkg/processor"
+	"github.com/conduitio/conduit/pkg/record"
+	"github.com/google/go-cmp/cmp"
+	"github.com/matryer/is"
+)
+
+func TestParseJsonKey_Process(t *testing.T) {
```
hariso, line 28: Nitpick: Json -> JSON

pkg/processor/procbuiltin/parsejson_test.go
```diff
@@ -0,0 +1,139 @@
...
+	tests := []struct {
+		name    string
+		args    args
+		want    record.Record
+		wantErr bool
+	}{{
+		name: "raw payload",
+		args: args{r: record.Record{
+			Payload: record.Change{
+				Before: nil,
```
hariso, line 95: Would be nice to have a test where `Before` is set.

pkg/processor/procbuiltin/parsejson_test.go
```diff
@@ -0,0 +1,139 @@
...
+	tests := []struct {
+		name    string
+		args    args
+		want    record.Record
+		wantErr bool
+	}{{
+		name: "raw key",
+		args: args{r: record.Record{
+			Key: record.RawData{
+				Raw:    []byte("{\"after\":{\"data\":4,\"id\":3}}"),
```
hariso, line 43: It would be nice to have a negative test case, with some invalid JSON.

hariso (commented):
Great work! :1st_place_medal:

pkg/processor/procbuiltin/unwraprecord.go
```diff
@@ -0,0 +1,185 @@
...
+	}
+	if _, ok := data[DebeziumFieldAfter]; !ok {
+		multiErr = multierror.Append(multiErr, cerrors.Errorf("the %q field is missing from debezium payload", DebeziumFieldAfter))
+	}
+	if _, ok := data[DebeziumFieldSource]; !ok {
+		multiErr = multierror.Append(multiErr, cerrors.Errorf("the %q field is missing from debezium payload", DebeziumFieldSource))
+	}
+	if _, ok := data[DebeziumFieldOp]; !ok {
+		multiErr = multierror.Append(multiErr, cerrors.Errorf("the %q field is missing from debezium payload", DebeziumFieldOp))
+	}
```
hariso, line 158: The op field may not be present. A Debezium connector can also include schema changes, in which case the op field is not present. As for `before` and `after`, there's virtually no difference between them being `null` and not being present (especially because some JSON serialization frameworks will simply skip them if they're `null`). With that, IMHO, returning errors in these cases is a bit too strict. 

I'm not sure about the source field. It looks like Debezium's source code says that `source` is optional too: https://github.com/debezium/debezium/blob/8ca5e0a9b51e5a14ac9e8fd342ffae3e12052428/debezium-core/src/main/java/io/debezium/data/Envelope.java#L243-L251

pkg/processor/procbuiltin/unwraprecord.go
```diff
@@ -0,0 +1,185 @@
...
+	"github.com/conduitio/conduit/pkg/foundation/multierror"
+	"github.com/conduitio/conduit/pkg/processor"
+	"github.com/conduitio/conduit/pkg/record"
+)
+
+const (
+	unwrapRecordPayloadProcType = "unwraprecord"
+
+	// todo: support openCDC format
+	unwrapRecordConfigFormat = "format" // only debezium for now
```
hariso, line 31: >  // only debezium for now

But also kafka-connect?

pkg/processor/procbuiltin/unwraprecord.go
```diff
@@ -0,0 +1,185 @@
...
+	format = config.Settings[unwrapRecordConfigFormat]
+	if format == "" {
+		return nil, cerrors.Errorf("%s: record format type not specified", processorType)
+	}
+	if format != RecordFormatDebezium && format != RecordFormatKafkaConnect {
+		return nil, cerrors.Errorf("%s: record format should be one of {%s, %s}", processorType, RecordFormatDebezium, RecordFormatKafkaConnect)
+	}
+
+	return processor.InterfaceFunc(func(_ context.Context, rec record.Record) (record.Record, error) {
+		data := getSetter.Get(rec) // record.payload.after
```
hariso, line 66: What does the comment indicate?

pkg/processor/procbuiltin/unwraprecord_test.go
```diff
@@ -0,0 +1,228 @@
...
+	}
+	tests := []struct {
+		name    string
+		args    args
+		want    record.Record
+		config  processor.Config
+		wantErr bool
+	}{
+		{
+			name: "raw payload",
```
hariso, line 104: Would be nice to have a negative test case, e.g. with invalid JSON in raw data.

pkg/processor/procbuiltin/unwraprecord.go
```diff
@@ -0,0 +1,174 @@
...
+	"encoding/json"
+
+	"github.com/conduitio/conduit/pkg/foundation/cerrors"
+	"github.com/conduitio/conduit/pkg/foundation/multierror"
+	"github.com/conduitio/conduit/pkg/processor"
+	"github.com/conduitio/conduit/pkg/record"
+)
+
+const (
+	unwrapRecordPayloadProcType = "unwraprecord"
```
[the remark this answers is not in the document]
maha-hajja: I think `convert` is more about converting the whole record to a new format (which is what we do in the middleware), but here, we ARE taking out something and removing layers, we're taking out the record from the `rec.Payload.After `and returning that as the new record, the input record is openCDC, so is the output record. so for me `unwrap` makes sense

pkg/processor/procbuiltin/unwraprecord_test.go
```diff
@@ -0,0 +1,195 @@
...
+	"context"
+	"testing"
+
+	"github.com/conduitio/conduit/pkg/processor"
+	"github.com/conduitio/conduit/pkg/record"
+	"github.com/google/go-cmp/cmp"
+	"github.com/matryer/is"
+)
+
+var dbz = "{" +
```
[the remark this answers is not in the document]
maha-hajja: for some reason the json parser doesn't like it and the parsing fails, unless it's all in one line! I'll investigate it more

pkg/processor/procbuiltin/unwraprecord.go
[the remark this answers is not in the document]
lovromazgon: I think both of you make good points and both names would work. I'm leaning towards `unwrap` though, because it will be easier for us to distinguish them when we talk about "converting" and "unwrapping".

pkg/processor/procbuiltin/unwraprecord.go
maha-hajja: that we're getting `record.payload.after` from this method, but wasn't that helpful honestly.
either way, it's gone with the refactor 👀

pkg/processor/procbuiltin/parsejson.go
maha-hajja: agreed!

pkg/processor/procbuiltin/unwrap.go
```diff
@@ -0,0 +1,310 @@
...
+
+type unwrapProcessor struct {
+	unWrapper unWrapper
+
+	inInsp  *inspector.Inspector
+	outInsp *inspector.Inspector
+}
+
+// unWrapper unwraps the formatted record from the openCDC record
+type unWrapper interface {
```
lovromazgon, line 38: Unwrap is a single word, I think we should use `unwrapper` instead of `unWrapper`.

pkg/processor/procbuiltin/unwrap.go
```diff
@@ -0,0 +1,310 @@
...
+	DebeziumOpUpdate = "u"
+	DebeziumOpDelete = "d"
+	DebeziumOpRead   = "r" // snapshot
+
+	DebeziumFieldBefore  = "before"
+	DebeziumFieldAfter   = "after"
+	DebeziumFieldSource  = "source"
+	DebeziumFieldOp      = "op"
+	DebeziumFieldPayload = "payload"
+)
```
lovromazgon, line 182: I'd make unwrapper structs and constants private, no need to expose them.

pkg/processor/procbuiltin/unwrap.go
```diff
@@ -0,0 +1,310 @@
...
+		// unmarshal raw data to structured
+		err := json.Unmarshal(data.Bytes(), &structData)
+		if err != nil {
+			return record.Record{}, cerrors.Errorf("failed to unmarshal raw data as JSON: %w", unwrapPayloadProcType, err)
+		}
+	case record.StructuredData:
+		structData = d
+	default:
+		return record.Record{}, cerrors.Errorf("unexpected data type %T", unwrapPayloadProcType, data)
+	}
```
lovromazgon, line 199: This code is the same in both unwrappers, we could move it to `unwrapProcessor.Process`. Unwrappers could then expect `rec.Payload.After` to be of type `record.StructuredData` and return an error if it's not.

pkg/processor/procbuiltin/unwrap.go
```diff
@@ -0,0 +1,310 @@
...
+
+	var before record.StructuredData
+	if b, ok := structPayload[DebeziumFieldBefore].(map[string]any); ok {
+		before = b
+	}
+
+	var after record.StructuredData
+	if a, ok := structPayload[DebeziumFieldAfter].(map[string]any); ok {
+		after = a
+	}
```
lovromazgon, line 224: ❕ What if `ok` is `false`? I'd expect to see an error, same as we do for operation.

pkg/processor/procbuiltin/unwrap.go
```diff
@@ -0,0 +1,310 @@
...
+
+	var op string
+	if str, ok := structPayload[DebeziumFieldOp].(string); ok {
+		op = str
+	} else {
```
lovromazgon, line 229: I suggest a simplification.
```suggestion
	op, ok := structPayload[DebeziumFieldOp].(string)
	if !ok {
```

pkg/processor/procbuiltin/unwraprecord.go
maha-hajja: we're unwrapping the record that was created here https://github.com/ConduitIO/conduit-connector-sdk/blob/6cd5101ba73f7388dfea47dc188692d27845d455/record.go#L252-L259 , so all these fields should be there.
I'm not sure if this is too specific or if we should make it more generic tbh@samirketema @lovromazgon what do you think? is it possible that platform will have a record that doesn't contain these fields? if so, should we still support it?

pkg/processor/procbuiltin/unwrap.go
```diff
@@ -0,0 +1,310 @@
...
+	 example of kafka-connect record:
+		`{
+		 "payload": {
+		     "description": "desc",
+		     "id": 20
+		   },
+		 "schema": {} // will be ignored
+		}`
+*/
+type KafkaConnectUnWrapper struct{}
```
lovromazgon, line 109: ❕ Thinking out loud - should we be more strict in this unwrapper and check that the record only contains 2 fields, `payload` and `schema`? Otherwise some data might get thrown away without the user knowing. I wonder if Kafka Connect allows records to have fields beside `payload` and `schema` (e.g. `{"payload":{...},"schema":{...},"foo":"bar"}` 🤔

pkg/processor/procbuiltin/unwrap.go
```diff
@@ -0,0 +1,310 @@
...
+		if str, ok := v.(string); ok {
+			meta[k] = str
+		} else {
+			return nil, cerrors.Errorf("the value %q from the field %q is not a string", v, DebeziumFieldSource)
+		}
+	}
+	// append original record's metadata to meta
+	for k, v := range rec.Metadata {
+		meta[k] = v
+	}
```
lovromazgon, line 292: ❕If the OpenCDC record and Debezium record contain the same metadata key, the OpenCDC one takes precedence. I would expect the Debezium value to take over tbh, same as the operation is overwritten or payload before.

pkg/processor/procbuiltin/unwrap.go
```diff
@@ -0,0 +1,310 @@
...
+		multiErr = multierror.Append(multiErr, cerrors.Errorf("the %q field is missing from debezium payload", DebeziumFieldSource))
+	}
+	if _, ok := data[DebeziumFieldOp]; !ok {
+		multiErr = multierror.Append(multiErr, cerrors.Errorf("the %q field is missing from debezium payload", DebeziumFieldOp))
+	}
+	// ts_ms and transaction can be empty
+	return multiErr
+}
+
+func (d *DebeziumUnWrapper) unwrapMetadata(rec record.Record, payload record.StructuredData) (record.Metadata, error) {
```
lovromazgon, line 273: The record carries the payload as well, no need to pass it separately.
```suggestion
func (d *DebeziumUnWrapper) unwrapMetadata(rec record.Record) (record.Metadata, error) {
```

pkg/processor/procbuiltin/unwraprecord.go
```diff
@@ -0,0 +1,185 @@
...
+}
+
+// UnwrapRecordPayload unwraps the record format from the payload and returns the record as openCDC format
+func UnwrapRecordPayload(config processor.Config) (processor.Interface, error) {
+	return unwrapRecord(unwrapRecordPayloadProcType, recordPayloadGetSetter{}, config)
+}
+
+func unwrapRecord(
+	processorType string,
+	getSetter recordDataGetSetter,
```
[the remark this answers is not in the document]
maha-hajja: agreed! I think accessor makes more sense for what the interface does.
we could have the interface renamed to `recordFieldAccessor` and other structs that implement it to be `recordPayloadAccessor`, WDYT?

lovromazgon (commented):
First of all, I like the refactoring, this will make it easy to add more unwrap formats in the future! 👏 

I posted a bunch of nitpicks, I marked the important comments with ❕. And here are two more things we should probably do:

1. The Debezium format contains the field `ts_ms` that contains info about when the record was read, we should parse it and set it via [`Metadata.SetReadAt`](https://github.com/ConduitIO/conduit/blob/b7eccf8536f1d47c13683e1c37c9649e09e6b2b7/pkg/record/metadata.go#L127).
2. If we get a Debezium or Kafka connect format we should probably also try to unwrap the key, as it will most likely also contain a Kafka Connect compatible JSON.

pkg/processor/procbuiltin/unwrap.go
maha-hajja: goLand complained about it not being a word, so this one had no squiggly lines under it, that's why I renamed it. But I agree that it should be `unwrapper`

pkg/processor/procbuiltin/unwrap.go
maha-hajja: hmm I took it out because if `!ok` then `after` is nil, and will be assigned to the record as nil, but I should've checked that alone

pkg/processor/procbuiltin/unwrap.go
maha-hajja: I passed it because the payload could be raw data and I would've needed to parse it again, but now since we expect it to only be structured, then it will work

pkg/processor/procbuiltin/unwrap.go
maha-hajja: agreed

pkg/processor/procbuiltin/unwrap.go
lovromazgon: Those cases are different, if the Debezium `after` field is `nil` we should set the payload to `nil` as well, but if it's not a map we should return an error.

pkg/processor/procbuiltin/unwrap.go
lovromazgon: Time to update the dictionary, Goland! 🤓

pkg/processor/procbuiltin/unwraprecord.go
lovromazgon: The unwrap should work in the opposite way as the record formatter you linked to @maha-hajja, but it's not necessarily always going to be a record that comes from Conduit, it could be a record that is actually produced by Debezium.

That said, I thought we are focusing on unwrapping CDC records, as in [create](https://debezium.io/documentation/reference/2.1/connectors/postgresql.html#postgresql-create-events), [update](https://debezium.io/documentation/reference/2.1/connectors/postgresql.html#postgresql-update-events) and [delete](https://debezium.io/documentation/reference/2.1/connectors/postgresql.html#postgresql-delete-events) events. What's left are [truncate](https://debezium.io/documentation/reference/2.1/connectors/postgresql.html#postgresql-truncate-events), [message](https://debezium.io/documentation/reference/2.1/connectors/postgresql.html#postgresql-message-events) and [transaction](https://debezium.io/documentation/reference/2.1/connectors/postgresql.html#postgresql-transaction-metadata) events, but those would be much harder to parse into OpenCDC, we would need to first think about how we want to properly do that. Let's see what @samirketema says and if we can live without support for those events.

pkg/processor/procbuiltin/unwrap.go
maha-hajja: done

pkg/processor/procbuiltin/unwraprecord.go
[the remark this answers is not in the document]
hariso: I think that makes sense. But IMHO no need to change anything here, I wanted to use the chance and get the team's opinion about this.

pkg/processor/procbuiltin/unwraprecord.go
hariso: > it could be a record that is actually produced by Debezium.

I was about to say the same thing. A "real" Debezium connector may have created this record and we got it like that. If this processor is about unwrapping Debezium records, then it should rely on the Debezium record "specification".

> Let's see what @samirketema says and if we can live without support for those events.

IIUC, we actually have 2 options. One is to let the Debezium un-wrapper drop the "non-row" events and another is to let a user do that (maybe via a built-in processor?)

pkg/processor/procbuiltin/unwrap.go
lovromazgon: Let's not block because of this and not be too strict for no reason 😉 Resolving this comment.

pkg/processor/procbuiltin/unwraprecord.go
lovromazgon: I have a bad feeling about an unwrap processor that is dropping records 😕 Right now if we got such a record the pipeline would stop because of an error. This feels safer to me compared to dropping the record and assuming the user doesn't care about it. I'd rather annoy the user and be loud about the fact that we don't support that record than drop it.

pkg/processor/procbuiltin/unwraprecord.go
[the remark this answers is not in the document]
lovromazgon: I think accessor makes sense! I also wouldn't add it into this PR though.

pkg/processor/procbuiltin/unwraprecord.go
hariso: Same here, I believe simply dropping records would be a surprise (unless we make it completely explicit through the processor name and the docs that the processor does so). IMHO it would be nice to have a processor which keeps only "row records" and drops "non-row" records.

lovromazgon: Ok then let's leave it like it is and see how far we get with this solution.

---

Regarding the processor that drops "non-row" records - this is doable right now by using two processors, first `parsejsonpayload` and then `filterfieldpayload` which can be configured to drop records where the field `op` doesn't contain `c`, `d`, `u` or `r`. I assume you are talking about creating a dedicated processor that does this out of the box - we need to figure out what our philosophy regarding that will be. I'm worried that if we start creating dedicated processors for things like this we risk creating a long list of processors and confusing our users about which one to use.

That's a conversation for another time and place, let's get this one shipped now 🙂

hariso: I agree with leaving things as they are when it comes to "non-row" records. But IMHO we still shouldn't validate Debezium records in the way we do it now (existence of before, after, op fields).

pkg/processor/procbuiltin/unwrap.go
```diff
@@ -0,0 +1,350 @@
...
+		return record.Record{}, cerrors.Errorf("error unwrapping operation: %w", err)
+	}
+
+	metadata, err := d.unwrapMetadata(rec)
+	if err != nil {
+		return record.Record{}, cerrors.Errorf("error unwrapping metadata: %w", err)
+	}
+
+	// create a kafka connect unwrapper to unwrap the key
+	kafkaConnectUnwrapper := &KafkaConnectUnwrapper{}
```
lovromazgon, line 258: Nitpick: To prevent us creating a new struct for each record we could have a field of type `KafkaConnectUnwrapper` in `DebeziumUnwrapper` and use that like `d.kafkaConnectUnwrapper.UnwrapKey(key)`.

pkg/processor/procbuiltin/unwrap.go
```diff
@@ -0,0 +1,350 @@
...
+	// record must be structured
+	structPayload, ok := rec.Payload.After.(record.StructuredData)
+	if !ok {
+		return record.Record{}, cerrors.Errorf("record payload data must be structured data")
+	}
+	// get payload
+	payload := structPayload["payload"] // the payload has the debezium record
+	if p, ok := payload.(map[string]interface{}); ok {
+		structPayload = p
+	} else {
```
lovromazgon, line 216: Nitpick: I see indented happy path, I nitpick 🙃 We should generally strive to keep the happy path out of if statements.
```suggestion
	structPayload, ok = structPayload["payload"].(map[string]interface{}) // the payload has the debezium record
	if !ok {
```

pkg/processor/procbuiltin/unwrap.go
```diff
@@ -0,0 +1,350 @@
...
+
+	// check fields under payload
+	err := d.validateRecord(structPayload)
+	if err != nil {
+		return record.Record{}, err
+	}
+
+	var before record.StructuredData
+	beforeData := structPayload[debeziumFieldBefore]
+	b, ok := beforeData.(map[string]any)
```
lovromazgon, line 228: Nitpick: We can get away without `b`. Same with `after`/`a`.
```suggestion
	beforeData := structPayload[debeziumFieldBefore]
	before, ok := beforeData.(map[string]any)
```

pkg/processor/procbuiltin/unwrap.go
```diff
@@ -0,0 +1,350 @@
...
+	for k, v := range rec.Metadata {
+		meta[k] = v
+	}
+
+	var structPayload record.StructuredData
+	if p, ok := rec.Payload.After.(record.StructuredData)["payload"].(map[string]interface{}); ok {
+		structPayload = p
+	} else {
+		return nil, cerrors.Errorf("payload doesn't contain a record")
+	}
```
lovromazgon, line 303: Nitpick: Line 302 is unreachable, if that would be the case the unwrapper would already throw an error in line 217. We could simplify.
```
structPayload := rec.Payload.After.(record.StructuredData)["payload"].(map[string]interface{})
```

pkg/processor/procbuiltin/unwrap_test.go
```diff
@@ -0,0 +1,246 @@
...
+								"id":          27,
+							},
+							"before": interface{}(nil),
+							"op":     "u",
+							"source": map[string]interface{}{
+								"opencdc.readAt":  "1674061777225877000",
+								"opencdc.version": "v1",
+							},
+							"transaction": interface{}(nil),
+							"ts_ms":       float64(1674061777225),
```
lovromazgon, line 155: If `opencdc.readAt` wouldn't be set in `source` then this value would be used to set `opencdc.readAt` in metadata, right? Can we change the test to test this please?

lovromazgon (approved):
LGTM 🚀  I posted another list of nitpicks, but those are really only that - nitpicks 🙃 Nice work @maha-hajja 👏

pkg/processor/procbuiltin/unwrap.go
maha-hajja: we would still need the line `var before record.StructuredData`
without it, "before" would be a map type, but we need it to be structured data

pkg/processor/procbuiltin/unwrap.go
maha-hajja: It just felt weird to ignore a casting error 😅 but yeah it shouldn't get here enyways

pkg/processor/procbuiltin/unwrap_test.go
maha-hajja: `opencdc.readAt` should be a nano second value, right?

pkg/processor/procbuiltin/unwrap.go
lovromazgon: That's right, and we could cast it to structured data when we assign it to the record that we return.

pkg/processor/procbuiltin/unwrap_test.go
lovromazgon: Yeah, I expect the metadata field to contain `1674061777225000000`.

pkg/processor/procbuiltin/unwrap.go
lovromazgon: You're right, maybe it would be nice to add a comment why we can safely do this.

pkg/processor/procbuiltin/unwraprecord.go
lovromazgon: According to the Debezium specs I linked in a previous comment these 4 fields will be present in all records that the processor targets right now, why would we relax the condition?

hariso: You're right about the `op` field. But the `before` and `after` fields are optional. For example:

Screenshot from 2023-01-31 17-22-18

lovromazgon: Maybe we understand "optional" differently 😄 I thought it means that the field can be `null`, not that it might be missing entirely. For instance in a "create" event the field `before` will always be set to `null` since there was nothing there before, but they don't remove the field, they just set it to `null`. I'd expect the same behaviour for all change data events.

hariso: FTR we discussed at the standup. In addition to the comment above, we may want to discuss the validation a bit more (how to make sure this processor is actually called on a Debezium record? what if we have unknown fields?), but the team agrees none of this is a blocker.

PR #827, opened 2023-01-23
docs: Pipeline config files ADR

lovromazgon:
Gathers decisions around expected pipeline config file provisioning behavior.


docs/architecture-decision-records/20230123-pipeline-config-files.md
```diff
@@ -0,0 +1,60 @@
+# Pipeline Config Files
+
+## Summary
+
+This document describes the behavior of provisioning pipelines using pipeline
```
hariso, line 5: Really a nitpick: "config" has been around in our vocabulary for some time. I'd like to suggest using the complete word ("configuration") in "official" communication (docs, GitHub, web site, blog posts, etc.).

docs/architecture-decision-records/20230123-pipeline-config-files.md
```diff
@@ -0,0 +1,60 @@
...
+and tedious.
+
+## Decision
+
+Pipeline config files are loaded at the start of Conduit. The end result is a
+provisioned pipeline that matches whatever is defined in the config files (as
+long as the config files are valid).
+
+In a pipeline config files no changes are forbidden, config files are treated as
+the source of truth. This means that if you can change a config file between
```
hariso, line 23: "This means that if you can change" -- maybe without the "if"?

docs/architecture-decision-records/20230123-pipeline-config-files.md
```diff
@@ -0,0 +1,60 @@
...
+the state of that entity gets erased. That shouldn't be unexpected though,
+because you changed the plugin anyway, so you can't really expect to start from
+where you left off.
+
+Entities created through config files are stored in the Conduit store so that
+state can be preserved between runs.
+
+If a pipeline config file is loaded and we detect that Conduit already contains
+that pipeline in the store, the existing pipeline is updated to match whatever
+is in the config file.
```
hariso, line 36: > If a pipeline config file is loaded and we detect that Conduit already contains
that pipeline in the store

Is the following correct?

1. If the already existing pipeline has been created through the API, then provisioning fails.
2. If the already existing pipeline has been created through a pipeline config file, then the pipeline is updated.

docs/architecture-decision-records/20230123-pipeline-config-files.md
```diff
@@ -0,0 +1,60 @@
...
+the invalid change and recreate it from scratch. The consequence here is that
+any state of that entity will be deleted (e.g. starting position of the source
+connector).
+
+When creating entities we mark them with a flag if they are created from a
+config file - if they are not found in a config file at startup, those entities
+get deleted. Entities that were provisioned through a config file also can't be
+mutated through the API, they need to be changed in the corresponding config
+file.
+
```
hariso, line 50: It might be good to mention that config files are treated individually and one invalid config file won't prevent other pipelines to be loaded. Also, it's worth mentioning how we handle duplicate pipelines.

hariso (commented):
Great work, Lovro! :rocket: I left a few minor comments and questions.

docs/architecture-decision-records/20230123-pipeline-config-files.md
lovromazgon: 1. Good questions, I looked at the code and it seems like we don't check how the pipeline was provisioned before updating it, so we would overwrite it even if it was previously created through the API. Not really the behavior I'd expect 🤔
2. Yes, this case is true (since it apparently doesn't matter how the pipeline was provisioned in the first place).

hariso: 1. same here, I wouldn't expect that.

lovromazgon: I created a bug report for point 1: https://github.com/ConduitIO/conduit/issues/837

hariso: So regarding #1, I created a pipeline through the API and then restarted Conduit with a pipeline config file that contains a pipeline with the same name. I got "pipeline name already exists", no updates were done.

hariso (approved):
:zap:

PR #833, opened 2023-01-26
Force stop pipeline + Go 1.20 + internal improvements

### LINKED ISSUE #825: Feature: Force stop pipeline
There’s two asks per this:

1. Force Stop - Option to stop  
2. Cascading Delete - for all of the objects attached to a pipeline, you would be able to delete pipeline and all those objects would get deleted too.

lovromazgon:
### Description

This PR started out focused on force stopping a pipeline but later grew to address other concerns, here is a summary of everything in this PR, I will break them out into separate PRs:

- https://github.com/ConduitIO/conduit/pull/856 Mainly it's focusing on providing a way to force stop a pipeline. This includes the proto files and API endpoints, orchestrator, the force stop functionality itself as well as changes in the nodes to ensure a node actually stops on a force stop, regardless of what the connector is doing.
- https://github.com/ConduitIO/conduit/pull/855 Force stopping a pipeline is now implemented as an idempotent operation - it will cancel the tomb context that is shared by all nodes and call the function `ForceStop` on nodes that implement this function (currently `SourceNode` and `DestinationNode`).
- https://github.com/ConduitIO/conduit/pull/852 Conduit is currently built using Go 1.18. As part of the implementation I used [`atomic.Bool`](https://pkg.go.dev/sync/atomic#Bool) which was introduced in Go 1.19. Since the latest Go version is 1.20 I upgraded the Go version to 1.20 and golangci-lint to 1.51.0 (needed to make it work with the upgraded Go version).
- https://github.com/ConduitIO/conduit/pull/853 In a test I saw the need for an utility to send an object to a channel and timeout if the operation doesn't succeed. I extended the `cchan` package with utilities for output channels, input channels and two-way channels. This means that the existing `cchan.Chan` object was renamed to `cchan.ChanOut` and additionally `cchan.ChanIn` and `cchan.Chan` were introduced.
- https://github.com/ConduitIO/conduit/pull/857 Some pipeline lifecycle tests were randomly failing (also in the [`main` branch](https://github.com/ConduitIO/conduit/actions/runs/3989729342/jobs/6842512996#step:4:859)). The problem was a race condition in the connector persister - once the source connector is terminated it calls `Persister.ConnectorStopped` which triggers a flush. If a flush is already happening the function blocks until that flush is done before it returns. The problem is that the connector is locked while `ConnectorStopped` is called and the flush routine also tries to acquire the lock of the same connector to persist it, which results in a deadlock. I fixed this by introducing `connector.Store.PrepareSet` which encodes the connector and returns a function that, when called, persists the connector to the database. When a connector now calls `Persister.Persist` the connector is encoded by the store and pushed onto the next batch. At that time the connector is already locked so there is no race condition with any other goroutine. Once the batch is flushed to the database the connector doesn't have to be locked again, because it's already encoded, the changes are simply written.
- https://github.com/ConduitIO/conduit/pull/852 A minor change in `scripts/check-go-version.sh` means that we don't have a hardcoded go version in that file anymore, it's read from the go module itself using `go list`.
- https://github.com/ConduitIO/conduit/pull/854 Buf was added back as a tool dependency into `tools.go` since we are using local generation again. This means that running `make install-tools` will install `buf` as well.

Future work:
- Right now we ensure that a force stop will stop a pipeline in case the connector is stuck. However, the same thing can happen in a processor, especially a custom JS processor. In the future it would be nice to add the same functionality to the processor node.

Fixes #825

### Quick checks:

- [ ] I have made sure that the PR is of reasonable size and can be easily reviewed. 👎


lovromazgon:
Closing in favor of other PRs linked in the description.

PR #839, opened 2023-01-30
Implement health check

### LINKED ISSUE #64: Health check
We have a health check endpoint that always returns a healthy status. We need to implement an actual health check that makes sure Conduit is running correctly. Right now the only thing to check is if the DB is working correctly (e.g. ping the DB). Once we tackle clustering this check should (probably) be improved to definitively indicate if a node in the cluster is healthy.

hariso:
### Description

This PR adds health checking to Conduit. It's possible to check the health of Conduit as a whole or check individual gRPC services (for more info, see [this](https://github.com/grpc/grpc/blob/master/doc/health-checking.md)).

Fixes #64 

### Quick checks:


pkg/conduit/runtime.go
```diff
@@ -326,8 +326,17 @@ func (r *Runtime) serveGRPCAPI(ctx context.Context, t *tomb.Tomb) (net.Addr, err
...
+	healthServer := api.NewHealthServer(
+		map[string]api.Checker{
+			"PipelineService":  r.pipelineService,
+			"ConnectorService": r.connectorService,
+			"ProcessorService": r.processorService,
+			"PluginService":    r.pluginService,
+		},
+		r.logger,
+	)
+	grpc_health_v1.RegisterHealthServer(grpcServer, healthServer)
```
hariso, line 339: Because of this I renamed `HealthChecker` to `HealthServer`.

pkg/foundation/database/badger/db.go
```diff
@@ -47,6 +48,15 @@ func New(l zerolog.Logger, path string) (*DB, error) {
 	}, nil
 }
 
+func (d *DB) Ping(ctx context.Context) error {
+	key := "test-" + uuid.NewString()
+	defer d.Set(ctx, key, nil) //nolint:errcheck // clean-up
+
+	// todo ideally, we find a way to test the connection
+	// without inserting data
```
lovromazgon, line 56: There is no connection, badger is embedded in the process and flushes data to the file system from time to time.

I tried running Conduit and deleting `conduit.db` while it was running. The health check was saying everything is fine, and indeed the service was running fine, even mutating and running pipelines was still possible. But once I wanted to stop Conduit it threw an error saying that it's unable to open `conduit.db/...`. I tried changing the health check to execute `d.db.Sync()` and not even that threw an error, Conduit was "healthy" and proceeded to normally work until it was stopped.

Long story short - I think apart from checking if the db is closed there's not much we can check here 😕

pkg/web/api/health_checker.go
```diff
@@ -17,28 +17,76 @@ package api
...
 }
 
-func NewHealthChecker() *HealthChecker {
-	return &HealthChecker{}
+// checkAll checks returns `true` if all service are healthy,
+// returns `false` otherwise (one or more services are not healthy).
+func (h *HealthServer) checkAll(ctx context.Context) bool {
+	ok := true
+	for service := range h.checkers {
+		ok = ok && h.check(ctx, service)
```
lovromazgon, line 72: Nitpick: ~we could cut the loop short if `h.check` returns `false`, and then if we get to the return after the loop we know it will be `true`.~ Nope, now I get it, I assume we do this to get the log message for all unhealthy services. If that's the case I'd just ask for a clarifying comment explaining why we are deliberately looping through all services when we already know the result.

pkg/web/api/health_checker.go
```diff
@@ -17,28 +17,76 @@ package api
...
+
+// HealthServer implements the gRPC health service.
+// Using the HealthServer, it's possible to check
+// the gRPC server's overall health, but also the
+// health of the individual gRPC services.
+// For more information, see: https://github.com/grpc/grpc/blob/master/doc/health-checking.md
+type HealthServer struct {
 	grpc_health_v1.UnimplementedHealthServer
+	checkers map[string]Checker
+	log      log.CtxLogger
```
lovromazgon, line 38: Nitpick: throughout the codebase we use the var name `logger` for this type, would be nice to have this consistent with the rest.

pkg/web/api/health_checker.go
```diff
@@ -17,28 +17,76 @@ package api
...
+	err := h.checkers[service].Check(ctx)
+	if err != nil {
+		h.log.Err(ctx, err).
+			Str("service_name", service).
+			Msg("health check not OK")
+	}
+	return err == nil
+}
+
+func NewHealthServer(checkers map[string]Checker, log log.CtxLogger) *HealthServer {
```
lovromazgon, line 87: Nitpick: normally I'd expect the constructor to be placed directly after the type declaration, I think [this is a good guideline](https://github.com/uber-go/guide/blob/master/style.md#function-grouping-and-ordering).

pkg/web/api/health_checker_test.go
```diff
@@ -0,0 +1,91 @@
...
+
+type testChecker struct {
+	err error
+}
+
+func (t *testChecker) Check(ctx context.Context) error {
+	return t.err
+}
+
+func TestHealthChecker_Check_OK(t *testing.T) {
```
lovromazgon, line 35: Seems like a leftover from renaming (same for other tests).
```suggestion
func TestHealthServer_Check_OK(t *testing.T) {
```

pkg/web/api/health_checker.go
hariso: Definitely agree. I normally do it but as new methods were added through the IDE they kept appearing at the top.:)

pkg/foundation/database/badger/db.go
hariso: > There is no connection, badger is embedded in the process and flushes data to the file system from time to time.

Yup, yup, I didn't mean a "permanent" connection, as we have it in DBs, but more like a "connection" between BadgerDB and the file system. 

Besides a method in BadgerDB's StreamWriter (which also deletes all the data:) ) I didn't find a good way to do this check. The only thing which we could potentially do is to check if the dir is there and if we can list the contents.

PR #848, opened 2023-02-01
Parse built-in plugin versions from build info

lovromazgon:
### Description

Standalone plugins have the correct version because we set it with `-ldflags` when building the binary. Built-in plugins on the other hand currently report their version using a constant that we need to remember to update before every release ([example](https://github.com/ConduitIO/conduit-connector-kafka/blob/0808ebd29f7135d989ffdf51c9c8b0962eefc4ab/spec.go#L24)). This PR makes this step unnecessary, because Conduit will overwrite the version with the version reported by go.mod.

I wanted to add unit tests but `debug.BuildInfo` is [not populated in tests](https://github.com/golang/go/issues/33976) 😕

### Quick checks:

### DIFF
--- a/pkg/plugin/builtin/registry.go
+++ b/pkg/plugin/builtin/registry.go
@@ -16,6 +16,7 @@ package builtin
 
 import (
 	"context"
+	"runtime/debug"
 
 	file "github.com/conduitio/conduit-connector-file"
 	generator "github.com/conduitio/conduit-connector-generator"
@@ -24,7 +25,7 @@ import (
 	postgres "github.com/conduitio/conduit-connector-postgres"
 	"github.com/conduitio/conduit-connector-protocol/cpluginv1"
 	s3 "github.com/conduitio/conduit-connector-s3"
-	"github.com/conduitio/conduit-connector-sdk"
+	sdk "github.com/conduitio/conduit-connector-sdk"
 	"github.com/conduitio/conduit/pkg/foundation/cerrors"
 	"github.com/conduitio/conduit/pkg/foundation/log"
 	"github.com/conduitio/conduit/pkg/plugin"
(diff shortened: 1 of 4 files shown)


pkg/plugin/builtin/registry.go
```diff
@@ -24,27 +25,29 @@ import (
...
+	// built-in plugins. The key of the map is the import path of the module
+	// containing the connector implementation.
+	DefaultDispenserFactories = map[string]DispenserFactory{
+		"github.com/conduitio/conduit-connector-file":      sdkDispenserFactory(file.Connector),
+		"github.com/conduitio/conduit-connector-kafka":     sdkDispenserFactory(kafka.Connector),
+		"github.com/conduitio/conduit-connector-generator": sdkDispenserFactory(generator.Connector),
+		"github.com/conduitio/conduit-connector-s3":        sdkDispenserFactory(s3.Connector),
+		"github.com/conduitio/conduit-connector-postgres":  sdkDispenserFactory(postgres.Connector),
+		"github.com/conduitio/conduit-connector-log":       sdkDispenserFactory(connLog.Connector),
 	}
```
hariso, line 46: Since all of our built-in connectors follow the same naming convention, we can determine the module name from the connector name.

hariso (approved):
LGTM! :1st_place_medal:

pkg/plugin/builtin/registry.go
lovromazgon: I was thinking of ways to determine the module (reflection, naming convention) and went with the dumb solution at the end. I'm not convinced it's worth changing it. I think your proposal would be a good fit if we were adding built-in plugins on a more regular basis, but for now I'd leave it like this.

PR #849, opened 2023-02-01
Deep-copy records when copying messages

### LINKED ISSUE #748: Possible concurrency issues with shallow record copies
### Bug description

Destination connectors mutate a record's metadata after they receive it. This can cause issues with the stream inspector (see [this comment](https://github.com/ConduitIO/conduit/pull/718#discussion_r1024054550)) or in pipelines with multiple destinations (see [this comment](https://github.com/ConduitIO/conduit/pull/718#discussion_r1028437875)).

### Steps to reproduce

TBD

### Version

v0.4.0-nightly.20221122

hariso:
### Description

Fixes #748 

### Quick checks:


pkg/record/record.go
```diff
@@ -147,6 +148,19 @@ func (r Record) mapData(d Data) interface{} {
 	return nil
 }
 
+func (r Record) Clone() (Record, error) {
+	clone := Record{}
+	err := copier.CopyWithOption(
```
lovromazgon, line 153: I have an issue with using a library with reflection for this 😕 It's a function that will be called at least once for every record flowing through the pipeline. I don't have benchmarks to prove it, but I imagine this being far slower than just manually copying fields (including deep copying maps). Sure, it's easier with this library, but since this is the hot path I wouldn't want to take shortcuts here.

hariso: Yeah, I considered that as well. My first version was in fact copying fields "manually". However, the after and before fields can be `StructuredData` which is `map[string]interface{}`. Implementing a deep copy option for `interface{}` would take some time, so for this PR I chose to first get the functionality fixed (especially because it's easy to reproduce) versus optimizing (for high volume streams).

lovromazgon: I see, good point, `interface{}` does indeed make things harder, it's not a trivial thing to get right, we'll need to recursively traverse it if it's a map or a slice. And we'll need to extensively test this for all types. You convinced me, let's do this separately, but we should not forget about it, please create an issue for this 🙏

Regarding that issue - let's consider what can be in that `interface{}`. We get that from the connector so it can't just be _anything_, it has to be something that gRPC can handle - it's either a basic type, a slice or a map (did I forget anything?). We plan to add [`time.Time`](https://github.com/ConduitIO/conduit/issues/866) to the list of supported types, but that's also a type that's used as a value and easily copied. So sounds like if we can copy a map with these values we should be good, and that should be doable without reflection.

pkg/record/record.go
```diff
@@ -147,6 +148,19 @@ func (r Record) mapData(d Data) interface{} {
 	return nil
 }
 
+func (r Record) Clone() (Record, error) {
```
lovromazgon, line 151: What needs to be in `Record` for this method to return an error? 🤔

hariso: The copier will return an error if there's something wrong with the source and destination of copying (source is a zero value, destination not addressable). IIRC, I got a panic with a struct with an unexported field in it. Now that I think of it, that could be a problem if we ever add unexported fields to `Record`.

lovromazgon: The source struct can't be zero since this method is declared on a value, and we are in charge of creating the destination value so we can guarantee it is addressable.

I tested copying a struct with an [unexported field](https://go.dev/play/p/y4aqr8SAam0) and it didn't error out, but I found some other strange behavior - the unexported field was _changed_ 🤨 This makes me a bit uneasy.

Anyway, my point is, if we are in charge of all the conditions that could make this fail, maybe we should just catch the error early and panic rather than return the error, since the caller can do nothing about it. Otherwise I challenge you to write a test that makes this function return an error 😛

hariso: In the test for clone I changed `Name` to be unexported ([here](https://github.com/ConduitIO/conduit/pull/849/files#diff-7ac1828e3d69c68456d56641b56ffda6e981673226b9ea12cf895277ca6f582fR26)) and it failed... But now I see it's because of go-cmp.:D 



> but I found some other strange behavior - the unexported field was _changed_ raised_eyebrow This makes me feel a bit uneasy.

Not sure if you noticed, but the first argument in `copier.CopyWithOption` is the destination. You may have expected a source there.

> Anyway, my point is, if we are in charge of all the conditions that could make this fail, maybe we should just catch the error early and panic rather than return the error, since the caller can do nothing about it.

I actually had that in mind too. Panicking felt a bit too brutal, but I can definitely implement it that way.:)

lovromazgon: Oh my bad, I didn't RTFM 🤦‍♂️ It works like a charm, thanks for pointing that out. I see that even unexported fields are clones.

Well yeah, panicking feels brutal, but I think that if we can guarantee that the contents of `Record` don't make a difference and the error could only happen if we are using the library wrong, then I'd rather panic and write this down as an error that we intend to catch in tests. It makes the usage simpler.

hariso: You're 200% right about the usage!

pkg/record/record.go
hariso: Forgot to mention the issue for the optimizing record cloning: https://github.com/ConduitIO/conduit/issues/885.

hariso: > Regarding that issue - let's consider what can be in that `interface{}`. We get that from the connector so it can't just be _anything_, it has to be something that gRPC can handle - it's either a basic type, a slice, or a map (did I forget anything?).

IIUC, the built-in connectors, in theory, can return pretty much anything, but the Conduit protocol is based on gRPC, so we need to be aligned with it?

pkg/record/record.go
hariso: Implemented as discussed above. Thanks for the suggestion @lovromazgon!

pkg/record/record.go
lovromazgon: That's right, we need to design for the common denominator, so gRPC and standalone connectors.

pkg/record/record.go
```diff
@@ -147,6 +148,32 @@ func (r Record) mapData(d Data) interface{} {
...
+		// At the moment, a clone error cannot happen because of the input data
+		// (i.e. the record itself).
+		// It can only happen if the clone destination is not addressable
+		// or if reflect.ValueOf(&r) is invalid.
+		// The first should never happen, because &Record{} is always addressable.
+		// The second, reflect.ValueOf(&r) is invalid if &r is nil, which in our case
+		// is also not possible.
+		// Hence, if the copier returns an error, it's a bug related to how we use copier
+		// which would cause all pipelines to fail anyway, so we panic here.
+		// This also makes the method signature simpler.
```
lovromazgon, line 171: ❤️

lovromazgon (approved):
Nice work 🚀

PR #853, opened 2023-02-03
Add channel utilities for input channels

lovromazgon:
### Description

In a test I saw the need for an utility to send an object to a channel and timeout if the operation doesn't succeed. I extended the `cchan` package with utilities for output channels, input channels and two-way channels. This means that the existing `cchan.Chan` object was renamed to `cchan.ChanOut` and additionally `cchan.ChanIn` and `cchan.Chan` were introduced.

### Quick checks:

### DIFF
--- a/pkg/foundation/cchan/chan_test.go
+++ b/pkg/foundation/cchan/chan_test.go
@@ -22,7 +22,7 @@ import (
 	"github.com/matryer/is"
 )
 
-func TestChan_Recv_Success(t *testing.T) {
+func TestChanOut_Recv_Success(t *testing.T) {
 	is := is.New(t)
 	ctx := context.Background()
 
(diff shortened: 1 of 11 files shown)


pkg/foundation/cchan/chan_test.go
```diff
@@ -80,3 +80,60 @@ func TestChan_RecvTimeout_DeadlineExceeded(t *testing.T) {
...
+func TestChanIn_Send_Closed(t *testing.T) {
+	is := is.New(t)
+	ctx := context.Background()
+	c := make(chan int)
+
+	close(c)
+
+	defer func() {
+		r := recover()
+		is.True(r != nil)
```
hariso, line 112: It would be nice to check what the error is.

hariso (approved):
LGTM! :zap:

PR #855, opened 2023-02-03
Force stop pipeline internals

lovromazgon:
### Description

Force stopping a pipeline is implemented as an idempotent operation by cancelling the connector context a node uses to communicate with the connector. Currently this is done in `SourceNode`, `DestinationNode`, `DestinationAckerNode` and `DLQHandlerNode`.

Depends on https://github.com/ConduitIO/conduit/pull/852 and https://github.com/ConduitIO/conduit/pull/853

### Quick checks:

- [X] I have followed the [Code Guidelines](https://github.com/ConduitIO/conduit/blob/main/docs/code_guidelines.md).
- [X] There is no other [pull request](https://github.com/ConduitIO/conduit/pulls) for the same update/change.
- [X] I have written unit tests.
- [X] I have made sure that the PR is of reasonable size and can be easily reviewed.

PR #856, opened 2023-02-03
Force stop pipeline API

lovromazgon:
### Description

The API endpoint for stopping a pipeline now accepts a request body with a JSON `{"force":true}` to force stop a pipeline. Force stopping a pipeline is implemented as an idempotent operation - it will cancel the tomb context that is shared by all nodes and call the function `ForceStop` on nodes that implement this function.

Depends on https://github.com/ConduitIO/conduit/pull/855, https://github.com/ConduitIO/conduit/pull/852 and https://github.com/ConduitIO/conduit/pull/853.

### Quick checks:


pkg/orchestrator/orchestrator.go
```diff
@@ -74,7 +74,7 @@ type PipelineService interface {
 	// Stop initiates a graceful shutdown of the given pipeline.
```
maha-hajja, line 74: we should update the docs for this method, it only talks about the graceful shutdown

pkg/pipeline/lifecycle.go
```diff
@@ -142,21 +152,35 @@ func (s *Service) stopWithReason(ctx context.Context, pl *Instance, reason error
...
-		err := s.stopWithReason(ctx, pl, reason)
+		err := s.stopGraceful(ctx, pl, reason)
 		if err != nil {
 			s.logger.Warn(ctx).
 				Err(err).
 				Str(log.PipelineIDField, pl.ID).
 				Msg("could not stop pipeline")
 		}
 	}
+	// TODO stop pipelines forcefully after timeout if they are still running
```
maha-hajja, line 183: is this out of scope? or is it not that high of a priority right now?

pkg/pipeline/lifecycle.go
```diff
@@ -474,19 +498,21 @@ func (s *Service) runPipeline(ctx context.Context, pl *Instance) error {
...
 	var nodesWg sync.WaitGroup
+	var isGracefulShutdown atomic.Bool
 	for id := range pl.n {
 		nodesWg.Add(1)
 		node := pl.n[id]
 
-		nodesTomb.Go(func() (errOut error) {
-			// If any of the nodes stops, the nodesTomb will be put into a dying state
+		pl.t.Go(func() (errOut error) {
+			// If any of the nodes stops, the tomb will be put into a dying state
```
maha-hajja, line 511: nitpick: 
```suggestion
			// If any of the nodes stop, the tomb will be put into a dying state
```

pkg/pipeline/lifecycle.go
```diff
@@ -528,28 +550,32 @@ func (s *Service) runPipeline(ctx context.Context, pl *Instance) error {
...
+
+		nodesWg.Wait()
+		err := pl.t.Err()
 
 		measure.PipelinesGauge.WithValues(strings.ToLower(pl.Status.String())).Dec()
 
-		switch {
-		case cerrors.Is(err, ErrGracefulShutdown):
-			// not an actual error, it was a graceful shutdown, that is expected
+		if err == tomb.ErrStillAlive {
```
maha-hajja, line 565: I think this check would be nicer as a switch, instead of nested if-else statements so that it would be a switch statement with if-else inside

pkg/pipeline/lifecycle.go
```diff
@@ -112,21 +113,30 @@ func (s *Service) Start(
...
 	}
 
-	s.logger.Debug(ctx).Str(log.PipelineIDField, pl.ID).Msg("stopping pipeline")
+	switch force {
+	case false:
+		return s.stopGraceful(ctx, pl, nil)
+	case true:
+		return s.stopForceful(ctx, pl)
+	}
+	panic("unreachable code")
```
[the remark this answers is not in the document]
lovromazgon: Just in case, so this doesn't go unnoticed if we somehow hit this.

pkg/pipeline/lifecycle.go
lovromazgon: I didn't want to increase the scope of this issue since the PR was already quite big. We can tackle this in a separate PR, we will need to rethink how we want this to work, e.g. do we want to force stop pipelines after a timeout? Should that timeout be configurable? Or maybe it should be triggered by the second SIGINT signal (currently Conduit is just killed by the second SIGINT).

pkg/pipeline/lifecycle.go
lovromazgon: Nice suggestion, thanks 👍

hariso (approved):

maha-hajja (approved):

PR #857, opened 2023-02-03
Fix race condition in connector persister

lovromazgon:
### Description

Some pipeline lifecycle tests were randomly failing (also in the [`main` branch](https://github.com/ConduitIO/conduit/actions/runs/3989729342/jobs/6842512996#step:4:859)). The problem was a race condition in the connector persister - once the source connector is terminated it calls `Persister.ConnectorStopped` which triggers a flush. If a flush is already happening the function blocks until that flush is done before it returns. The problem is that the connector is locked while `ConnectorStopped` is called and the flush routine also tries to acquire the lock of the same connector to persist it, which results in a deadlock. I fixed this by introducing `connector.Store.PrepareSet` which encodes the connector and returns a function that, when called, persists the connector to the database. When a connector now calls `Persister.Persist` the connector is encoded by the store and pushed onto the next batch. At that time the connector is already locked so there is no race condition with any other goroutine. Once the batch is flushed to the database the connector doesn't have to be locked again, because it's already encoded, the changes are simply written.

### Quick checks:


hariso:
Thanks for the detailed explanation of the bug! Did you maybe find a way to reproduce the bug? A similar bug we fixed was reproducible with a performance test: https://github.com/ConduitIO/conduit/issues/570.

lovromazgon:
@hariso What I did was run that failing test (`TestServiceLifecycle_PipelineStop`) repeatedly in a loop and it resulted in a deadlock eventually, after this change it's not happening anymore though.

pkg/connector/source.go
```diff
@@ -201,16 +207,20 @@ func (s *Source) Ack(ctx context.Context, p record.Position) error {
 		return err
 	}
 
-	// lock to prevent race condition with connector persister
+	// lock as we are updating the state and leave it locked so the persister
+	// can safely prepare the connector before it stores it
```
hariso, line 211: >  before it stores it

This is in reference to the persisted batch being flushed, and not that the connector will necessarily be stored here in Ack()?

hariso (approved):
LGTM! Had one question, a clarification, not a blocker!

pkg/connector/source.go
lovromazgon: Yeah, by "prepare the connector" I mean that the connector will be encoded into bytes, and since it's locked already it should be safe to do so. The persister will then only store the encoded bytes in the batch and flush it at a later point in time.

PR #869, opened 2023-02-09
Parallel processor JavaScript

lovromazgon:
### Description

The `js` processor needed some changes so we can run it in parallel. Goja runtime is not safe for concurrent use so now we create a new goja runtime whenever it's needed.

This PR also updates the dependency to Goja, since they are not releasing tagged versions and dependabot does not update it.

Depends on https://github.com/ConduitIO/conduit/pull/744.

### Quick checks:


maha-hajja (approved):
lgtm

hariso (approved):

PR #870, opened 2023-02-09
Support debezium records with raw data (MongoDB records)

lovromazgon:
### Description

This PR contains logic to unwrap a Debezium record coming from MongoDB, i.e. a record with raw data in fields `after` and `before`.

### Quick checks:


pkg/processor/procbuiltin/unwrap.go
```diff
@@ -281,44 +291,60 @@ func (d *debeziumUnwrapper) validateRecord(data record.StructuredData) error {
...
+			continue // ignore
+		case debeziumFieldTimestamp:
+			tsMs, ok := val.(float64)
+			if !ok {
+				return nil, cerrors.Errorf("%s is not a float", debeziumFieldTimestamp)
+			}
+			readAt := time.UnixMilli(int64(tsMs))
+			rec.Metadata.SetReadAt(readAt)
+		case debeziumFieldSource:
+			// don't add prefix for source fields
```
hariso, line 309: What's the reason for this btw?

lovromazgon: For compatibility with the [SDK](https://github.com/ConduitIO/conduit-connector-sdk/blob/059da633a9feb696d0c61a27e806ed96252d1d29/record_formatter.go#L180). There we put metadata fields into `source`, so if another Conduit instance processes such a record it unwraps back the original OpenCDC record.

hariso: Gotcha, thanks for the clarification! :raised_hands: Having a comment about that would be very useful IMO. :+1:

lovromazgon: Added the comment here: https://github.com/ConduitIO/conduit/pull/877/commits/30d21e0a651ede80eafe04cce9f3973d971ac6ce

PR #877, opened 2023-02-14
Graceful stop panic fix

lovromazgon:
### Description

This fixes a panic that can happen when stopping a pipeline gracefully.

### Quick checks:

### DIFF
--- a/pkg/pipeline/stream/source.go
+++ b/pkg/pipeline/stream/source.go
@@ -177,11 +177,11 @@ func (n *SourceNode) Stop(ctx context.Context, reason error) error {
 		// only execute stop once, more calls won't make a difference
 		err = n.stop(ctx, reason)
 		stopExecuted = true
-		if err != nil {
-			// an error happened, allow stop to be executed again
-			n.stopOnce = sync.Once{}
-		}
 	})
+	if err != nil {
+		// an error happened, allow stop to be executed again
+		n.stopOnce = sync.Once{}
+	}
 	if !stopExecuted {
 		n.logger.Warn(ctx).Msg("source connector stop already triggered, " +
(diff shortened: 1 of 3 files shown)


pkg/processor/procbuiltin/unwrap.go
```diff
@@ -306,7 +306,9 @@ func (d *debeziumUnwrapper) unwrapMetadata(rec record.Record) (record.Metadata,
 			readAt := time.UnixMilli(int64(tsMs))
 			rec.Metadata.SetReadAt(readAt)
 		case debeziumFieldSource:
-			// don't add prefix for source fields
+			// don't add prefix for source fields to be consistent with the
+			// behavior of the debezium converter in the SDK - it puts all
+			// metadata fields into the `source` field
```
hariso, line 311: Thanks for this! :D

pkg/pipeline/stream/source.go
```diff
@@ -177,11 +177,11 @@ func (n *SourceNode) Stop(ctx context.Context, reason error) error {
 		// only execute stop once, more calls won't make a difference
 		err = n.stop(ctx, reason)
 		stopExecuted = true
-		if err != nil {
-			// an error happened, allow stop to be executed again
-			n.stopOnce = sync.Once{}
```
lyuboxa, line 182: is that allowed in any circumstances? replacing the receiver in execution?

hariso (approved):

pkg/pipeline/stream/source.go
hariso: YOLO:)

lovromazgon: Don't do this at home, kids!

PR #880, opened 2023-02-16
Health check docs

hariso:
### Description

Adds a few words about the health check.

### Quick checks:


docs/health_check.md
```diff
@@ -0,0 +1,18 @@
...
+can successfully connect to the database which it was configured with (which can be BadgerDB, PostgreSQL or the 
+in-memory one). The health check is available at the `/healthz` path. Here’s an example:
+
+```bash
+$ curl "http://localhost:8080/healthz"
+{"status":"SERVING"}
+```
+
+You can also check individual services within Conduit. The following example checks if the PipelineService is running:
+```bash
```
lovromazgon, line 12:
```suggestion

```bash
```

lovromazgon (approved):
Nice 👍  Just some formatting stuff, we should introduce a linter for markdown files, shouldn't be too hard 🙂

Btw, should we add this to conduit.io docs as well?

PR #882, opened 2023-02-16
 Stream Inspector: Close inspector when connector/processor is deleted

### LINKED ISSUE #743: Stream Inspector: Close inspector when connector/processor is deleted
### Feature description

Currently, when a connector/processor is deleted, attached inspector sessions continue living.

hariso:
### Description

Fixes #743 

### Quick checks:


pkg/inspector/inspector.go
```diff
@@ -134,6 +144,12 @@ func (i *Inspector) NewSession(ctx context.Context) *Session {
 	return s
 }
 
+func (i *Inspector) Close() {
+	for _, s := range i.sessions {
+		s.close()
+	}
```
lovromazgon, line 150: If we are strict, we should be locking `i.lock` before we loop through `i.sessions`, but then `s.close()` also tries to lock it and we have a deadlock 😕 For correctness' sake we might want to lock first, then copy `i.sessions` into a separate map/slice, then unlock and then loop through the copied map/slice to close the sessions. It still doesn't guarantee that a new session won't be created while this is happening.

I know, it's an edge case, but I'm trying to say that something feels off regarding this design 😕

pkg/processor/service.go
```diff
@@ -162,6 +162,7 @@ func (s *Service) Delete(ctx context.Context, id string) error {
 		return err
 	}
 
+	instance.Processor.Close()
```
lovromazgon, line 165: If the delete in the next line fails we will drop inspector instances even though the processor won't be deleted. Again, an edge case, but given that this operation can't fail we might want to execute it after the store delete.

pkg/inspector/inspector.go
hariso: I wasn't too happy with these locks from the beginning, and even less now. The thing is that the lock handles concurrent access to the sessions' map, but what we actually need to do is to make sure that the inspector will refuse to create a new session ever again after it's been closed.

IMHO, the edge case is rare enough that working more on this design doesn't return much, especially after we move `inspector.Close()` after connector/processor deletion. At that point, starting a new session won't be possible at all (the API will complain that the processor/connector doesn't exist).

pkg/processor/service.go
hariso: Makes sense!

pkg/inspector/inspector.go
hariso: > Still, it doesn't feel right to have a component that can behave unexpectedly if it's used the wrong way and we rely on using it correctly elsewhere confused

I definitely agree with that and my point was that it reduces the changes of something which is an edge case already.

> I gave it a go and spiked out a slightly different approach to managing sessions in the inspector using a channel (see [2796796](https://github.com/ConduitIO/conduit/commit/279679648b58f5537246dbc00eb25388449d1478)).

Really nice! :+1: It solves some of the annoyances we currently have. I was initially thinking about something like a slice or a channel to store the sessions, but didn't go with it because I wanted the sessions to be removed and not stay around after they are not needed anymore.

> The nice thing about this approach is that we don't need to lock anything to check for open sessions

When you say "to check for open sessions" which part do you refer to?

lovromazgon: I'm talking about [this part](https://github.com/ConduitIO/conduit/commit/279679648b58f5537246dbc00eb25388449d1478#diff-3282116ee9a0d26e5209ed21cdbe482ca8df07ca06624b806262952789dd6ec2R90-R93) where we return early if there are no open sessions. If we did the same thing with a map we would need to acquire a lock first.

PR #905, opened 2023-02-23
Housekeeping: Markdownlint

lovromazgon:
### Description

Adds the CI action https://github.com/DavidAnson/markdownlint-cli2-action that executes https://github.com/DavidAnson/markdownlint when markdown files are updated. Settings are stored in `.markdownlint.yml` and taken mostly from [here](https://github.com/DavidAnson/markdownlint/blob/main/.markdownlint.json) with some changes (i.e. line length 120, tables excluded from line length checks, hard tabs are allowed in `go` code blocks, Conduit needs to be written with an uppercase C).

The CI action lints all markdown files (i.e. files that end with `.md`) except `LICENSE.md`, anything in `pkg/web/openapi`, anything in `ui/node_modules` and markdown files in `.github`.

I also ran `markdown-cli2-fix` that fixed most of the issues, the rest were fixed manually.

### Quick checks:


docs/architecture-decision-records/20220121-conduit-plugin-architecture.md
```diff
@@ -86,7 +86,7 @@ Functions will be called in the order in which they are defined.
   and* all remaining acknowledgments are received (this is handled by the SDK).
 - `Stop` - Conduit signals to the plugin it should stop fetching new records from the origin. The plugin is still
   allowed to produce records in the response stream of `Run` in case there are any cached records in memory, but it
-  should not go and fetch new ones from the origin.
+  should not Go and fetch new ones from the origin.
```
jayjayjpg, line 89: I think this one would still be `go` since the verb is used here, instead of the noun referring to the programming language:

```suggestion
  should not go and fetch new ones from the origin.
```

docs/code_guidelines.md
```diff
@@ -7,78 +7,78 @@ Unless specified otherwise, we should follow the guidelines outlined in
...
 
 Additionally, we encourage the usage of the `internal` package to hide complex internal implementation details of a
 package and enforce a better separation of concerns between packages.
 
-### Logging
+## Logging
 
 We want to keep our logs as minimal as possible and reserve them for actionable messages like warnings and errors when
-something does not go as expected. Info logs are fine when booting up the app, all other successful operations should be
+something does not Go as expected. Info logs are fine when booting up the app, all other successful operations should be
```
jayjayjpg, line 44: I think this would also be lowercase `go`:

```suggestion
something does not go as expected. Info logs are fine when booting up the app, all other successful operations should be
```

jayjayjpg (approved):
This reads great, thank you for adding this useful tool to the CI workflow and updating the repo docs so quickly! :sparkles:

docs/architecture-decision-records/20220121-conduit-plugin-architecture.md
lovromazgon: Lol, didn't notice that 😅 Good points, I'll remove that rule and revert these words.

.github/workflows/markdown-lint.yml
```diff
@@ -0,0 +1,21 @@
+name: markdown-lint
```
maha-hajja, line 1: I just have one question, so this linter will be executed every time a commit is pushed right? should we maybe add a command to the Makefile so we can test it locally before pushing our changes? or should we just run it locally like we do with `golangci-lint run`

maha-hajja (commented):
lovely work! and super useful :D

.github/workflows/markdown-lint.yml
lovromazgon: Sure, I can add a make target for linting markdown files 👍 It won't run in every PR, only in PRs that change markdown files.

lovromazgon: Done 👍 I also updated the pre-push hook to lint markdown files.

maha-hajja: perfect! thank you

maha-hajja (approved):

PR #909, opened 2023-02-24
Fix: state not set in file-provided pipelines

hariso:
### Description

This PR fixes the following bug.

**Bug description**
When Conduit is restarted with a pipeline, that was provided from a configuration file, the state is **not** correctly set after restart and lost.

**Steps to reproduce**
1. Start Conduit with a pipeline configuration file
2. Shut down Conduit gracefully
3. Start Conduit again

**Expected behavior**
After the restart, Conduit should start from the previous source position. Only new records are read.

**Actual behavior**
After the restart, Conduit re-reads all the records from source. Following can be found in logs:
```
could not apply old state, the connector will start from the beginning
```

### Quick checks:


pkg/connector/service_test.go
```diff
@@ -388,3 +389,91 @@ func TestService_UpdateInstanceNotFound(t *testing.T) {
...
+		{
+			name:     "completely wrong state",
+			connType: TypeSource,
+			state:    testCase{name: "completely wrong state"},
+			wantErr:  ErrInvalidConnectorStateType,
+		},
+	}
+
+	for _, tc := range testCases {
+		tc := tc
```
lovromazgon, line 442: I think we don't need this, do we?

lovromazgon (approved):
Good catch! 👍

pkg/connector/service_test.go
hariso: I don't think so.:)

Issue #847, opened 2023-02-01
Bug: Non-deterministic order of processors in file provisioned pipelines


lovromazgon:
### Bug description

When a pipeline is provisioned using a configuration file the processor order is not guaranteed because of the use of a [map](https://github.com/ConduitIO/conduit/blob/eacafff5bc575f14396a95c8fd402b8316c1dfb2/pkg/provisioning/parse.go#L38). We have two options to mitigate this:
- Change parsing to decode the contents of the map into a `yaml.Node`, then iterate through it manually and build an ordered slice of processors. The benefit is that the pipeline configuration file structure doesn't change, but the code won't be nice and people will have headaches ensuring the order when creating pipeline config files programatically (at least those using Go).
- Change the pipeline config file structure so that processors are stored in an array instead of a map. The benefit is that it's a cleaner solution and clearly indicates that processors are ordered, the drawback is that it would break existing pipeline config files (unless we add backwards compatibility support and bump the major version of pipeline config files). Possible structure:
```yaml
...
processors:
  - name: proc1
    type: js
    settings: ...
  - name: proc2
    type: filterfieldpayload
    settings: ...
...
```

### Steps to reproduce

1. Provision a pipeline config file with multiple processors.
2. Repeat multiple times and see the order change.

### Version

v0.5.0

PR #926, opened 2023-03-03
Refactor pkg/provisioning

lovromazgon:
### Description

This PR refactors the package `pkg/provisioning` and pushes code into packages `config` and `config/yaml`. Now we have an abstraction layer between the actual parser implementation (e.g. `yaml.Parser`) and the parser that is used by the service (interface `config.Parser`). The types that the provisioning service is working with are now in `config`, while the parser implementation declares its own types. This means that we don't have a direct dependency between the structure of the config file and the provisioning service.

The goal is to make it possible to introduce more config parsers. In the next PR we plan to introduce a new parser for a new version of the pipeline config file. This structure will also allow us to introduce parsers for different formats in the future (e.g. JSON, TOML, ...).

There are a few semantic changes:
- The YAML parser now doesn't care anymore if there are duplicated pipeline IDs in a file, that's checked by the provisioning service.
- In the case of duplicated pipeline IDs we previously provisioned the first one we encountered and skipped all others, the order was not deterministic. Now we skip _all_ pipelines that share the same ID and log an error on how many were skipped.
- We don't append the parent ID to the name of a connector or processor anymore (we still append it to their IDs to ensure their uniqueness).

The rest of the functionality should have stayed the same, although some of it was shuffled around.

Related to https://github.com/ConduitIO/conduit/issues/847.

### Quick checks:

- [X] I have made sure that the PR is of reasonable size and can be easily reviewed. (It seems big, but it's mostly just moving things around)


pkg/foundation/ctxutil/filepath.go
```diff
@@ -0,0 +1,52 @@
...
+	filepath := ctx.Value(filepathCtxKey{})
+	if filepath != nil {
+		return filepath.(string)
+	}
+	return ""
+}
+
+// FilepathLogCtxHook fetches the record filepath from the context and if it
+// exists it adds the filepath to the log output.
+type FilepathLogCtxHook struct{}
```
lovromazgon, line 44: This log hook was introduced so that a `config.Parser` can operate on an `io.Reader` and don't care about where a config comes from, while still retaining the filepath in log messages.

pkg/provisioning/config/yaml/parser.go
```diff
@@ -0,0 +1,109 @@
...
+		e := p.logger.Warn(ctx).
+			Int("line", uerr.Line()).
+			Int("column", uerr.Column())
+		e.Msg(uerr.Error())
+	}
+	return nil
+}
+
+// configurationsToConfig transforms all configurations to config.Pipeline types.
+func (p *Parser) configurationsToConfig(in []Configuration) []config.Pipeline {
```
hariso, line 99: Nitpick: Maybe `toConfig` is enough, as it's pretty much everything the method is doing (calling `ToConfig` on all yaml Configurations). I see the reasoning behind the method name (yaml configurations to the `config` structs), but it's still a bit unclear (configurations to configs). However, I'm not too sure what would be a better name, besides just shortening it.

hariso (approved):
Great work! :1st_place_medal:

pkg/provisioning/service.go
```diff
@@ -72,32 +76,47 @@ func (s *Service) Init(ctx context.Context) error {
...
+	var allPls []string
+
+	// delete duplicate pipelines (if any)
+	for duplicateID, duplicateIndexes := range s.findDuplicateIDs(configs) {
+		multierr = cerrors.Errorf("%d pipelines with ID %q will be skipped: %w", len(duplicateIndexes), duplicateID, ErrDuplicatedPipelineID)
+		configs = s.deleteIndexes(configs, duplicateIndexes)
+
+		// duplicated IDs should still count towards all encountered pipeline IDs
+		allPls = append(allPls, duplicateID)
+	}
```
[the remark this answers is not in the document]
lovromazgon: I think we didn't discuss it explicitly and just said it's _good enough_. Given that the behavior is non-deterministic (i.e. you don't know which pipeline will be provisioned and which one won't be), I think it's better to change it and not provision anything in case of a conflict.

Pipelines in the configuration file are in a list, yes. Here we detect duplicates even across files.

pkg/provisioning/service.go
```diff
@@ -72,32 +76,47 @@ func (s *Service) Init(ctx context.Context) error {
...
-		allPls = append(allPls, all...)
-		if len(provPipelines) != 0 {
-			successPls = append(successPls, provPipelines...)
+		configs = append(configs, cfg...)
+	}
+
+	// contains pipelineIDs of all the pipelines in all the configuration files, either successfully provisioned or not.
+	var allPls []string
+
+	// delete duplicate pipelines (if any)
```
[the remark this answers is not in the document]
lovromazgon: Maybe "delete" is not the right word, we detect duplicates and delete them from the list of pipelines that will be provisioned. I'm not sure if it's documented, I'll add a comment to https://github.com/ConduitIO/conduit/issues/939 so we don't forget to document it 👍

pkg/provisioning/config/yaml/parser.go
lovromazgon: Naming is hard 😄 I agree, this name is bad, I'll change it to `toConfig` unless I think of an even better one.

lovromazgon: I solved it by introducing the struct `Configurations` which is just a `[]Configuration` that implements the method `ToConfig`, so it follows the same approach as all other model types.

PR #928, opened 2023-03-06
Refactor pkg/provisioning/config/yaml

lovromazgon:
### Description

Another refactoring PR related to config files. This one further refactors the YAML config parser and separates code related to version `v1` into a separate package. After this, we are ready to tackle the last two things to fix the ordering of processors:
- create types for parsing version `v2`
- change the parser to choose the correct types based on the yaml version (right now it simply uses [`v1` types](https://github.com/ConduitIO/conduit/blob/30081d9fb82111ebcaa6b9d4e361b09369f52b63/pkg/provisioning/config/yaml/parser.go#L58-L59))

Depends on:
- https://github.com/ConduitIO/conduit/pull/926
- https://github.com/ConduitIO/yaml/pull/4
  - This was needed so that we can output a warning for inner nodes, not only leaf nodes (e.g. `processors` is an inner node in our configurations). This makes warnings more targeted so they point directly to the offending line.

Related to https://github.com/ConduitIO/conduit/issues/847.

### Quick checks:


pkg/provisioning/config/yaml/linter.go
```diff
@@ -49,16 +57,23 @@ func (cl *configLinter) init() {
...
 	}
 
-	curMap := expandedChangelog[cl.version]
-	for _, field := range path {
+	if c, ok := cl.findChange(path); ok {
+		cl.addWarning(path[len(path)-1], node, c.Message)
+	}
+}
+
+func (cl *configLinter) findChange(path []string) (internal.Change, bool) {
```
lovromazgon, line 73: This logic was adjusted a bit so that we don't output all warnings that match anything in `path` but just output the one warning that matches the whole `path`.

For example, if we previously had a change in the changelog tied to the field `foo.bar` and the path contained `[foo,bar,another,field]` we produced a warning for this line because of `foo.bar`. This also meant that we produced a warning for each descendant element that was a leaf (has no children). The reason was that the `yaml` library triggered the hook only for leaf nodes. Now the `yaml` lib triggers the hook for all nodes (including inner nodes), so we can target the warnings better.

pkg/provisioning/config/yaml/parser_test.go
```diff
@@ -135,19 +139,19 @@ func TestParser_Warnings(t *testing.T) {
...
+	want := `{"level":"warn","component":"yaml.Parser","line":5,"column":5,"message":"field unknownField not found in type v1.Pipeline"}
+{"level":"warn","component":"yaml.Parser","line":18,"column":11,"field":"processors","value":"","message":"the order of processors is non-deterministic in configuration files with version 1.x, please upgrade to version 2.x"}
+{"level":"warn","component":"yaml.Parser","line":24,"column":7,"field":"processors","value":"","message":"the order of processors is non-deterministic in configuration files with version 1.x, please upgrade to version 2.x"}
+{"level":"warn","component":"yaml.Parser","line":31,"column":7,"field":"dead-letter-queue","value":"","message":"field dead-letter-queue was introduced in version 1.1, please update the pipeline config version"}
+{"level":"warn","component":"yaml.Parser","line":52,"column":11,"field":"processors","value":"","message":"the order of processors is non-deterministic in configuration files with version 1.x, please upgrade to version 2.x"}
```
lovromazgon, line 146: Related to https://github.com/ConduitIO/conduit/pull/928#discussion_r1126720243: notice how we previously had 4 warnings related to `dead-letter-queue` and we now only have 1.

go.mod
```diff
@@ -15,7 +15,7 @@ require (
 	github.com/conduitio/conduit-connector-protocol v0.4.0
 	github.com/conduitio/conduit-connector-s3 v0.3.0
 	github.com/conduitio/conduit-connector-sdk v0.5.1
-	github.com/conduitio/yaml/v3 v3.2.0
+	github.com/conduitio/yaml/v3 v3.2.1-0.20230306160152-c42fb08ec7ff
```
lovromazgon, line 18: Note to self - merge https://github.com/ConduitIO/yaml/pull/4, release `v3.2.1` and update the dependency before merging this PR.

pkg/provisioning/config/yaml/linter.go
```diff
@@ -21,23 +21,31 @@ import (
...
 	"github.com/conduitio/yaml/v3"
 )
 
 type configLinter struct {
-	version  string
-	warnings warnings
+	changelog internal.Changelog
+
+	expandedChangelog map[string]map[string]any
+	version           string
```
hariso, line 32: `expandedChangelog` contains all versions, and `version` is just the version being "linted"?

hariso (approved):

pkg/provisioning/config/yaml/linter.go
lovromazgon: Yes! I'll add a comment 😅

lovromazgon: I added a comment in the last PR to prevent more rebases. `version` was actually removed in the next PR because the linter gets the version from the parser, I added a comment for `expandedChangelog` though (https://github.com/ConduitIO/conduit/pull/929/commits/1dfbabeabfc1fd8f2915fc1b87afe56188296eb4).

PR #929, opened 2023-03-07
YAML Configuration v2

### LINKED ISSUE #847: Bug: Non-deterministic order of processors in file provisioned pipelines

lovromazgon:
### Description

This PR introduces a new configuration version `2.0` and the corresponding types for parsing that version. The parser was changed to first figure out the version and then choose the correct types for decoding the configuration. If the version is empty we default to `2.0`, if it's specified and we recognize the version we use it, if we don't recognize it we check the major version and if we recognize it (i.e. `1` or `2`) we assume it's the latest known version for that major version (i.e. `1.1` or `2.0`), otherwise we fail and refuse to parse the configuration file. This means that Conduit will be forwards compatible with newer `2.x` versions, and if there are unrecognized fields we will output warnings (as we already do).

The tests for the new parser version are essentially copies of existing `1.x` tests. Tests for the service were adjusted to use version `2.0` and they worked out of the box, which is a good sign.

The linter was changed a bit so it gets the version information from the caller (parser), since it already has that information. Previously the line numbers of warnings didn't always line up with the offending line, this was fixed in https://github.com/ConduitIO/yaml/pull/4.

Fixes #847

### Quick checks:


pkg/provisioning/config/yaml/linter.go
```diff
@@ -16,62 +16,52 @@ package yaml
...
+		expandedChangelog: expandedChangelog,
 	}
-	sort.Sort(versions)
-
-	// latest version is the default version
-	cl.version = versions[len(versions)-1].Original()
 }
 
-func (cl *configLinter) InspectNode(path []string, node *yaml.Node) {
-	if len(path) == 1 && path[0] == "version" {
```
lovromazgon, line 58: The linter doesn't have to figure out the version itself anymore. We parse the version before actually decoding the config since we need to choose the types based on the version, so we supply it to the linter as well.

pkg/provisioning/config/yaml/parser_test.go
```diff
@@ -236,7 +241,295 @@ func TestParser_EnvVars(t *testing.T) {
...
+func TestParser_V2_ParseV1Config(t *testing.T) {
+	is := is.New(t)
+	parser := NewParser(log.Nop())
+	filepath := "./v1/testdata/pipelines1-success.yml"
+
+	file, err := os.Open(filepath)
+	is.NoErr(err)
+	defer file.Close()
+
+	// replace major version so that the v2 parser is chosen for a v1 config
```
hariso, line 503: Why do we want to force a v2 parser btw?

hariso (approved):

pkg/provisioning/test/pipelines1/provision1-pipeline-with-processors.yml
```diff
@@ -1,29 +1,29 @@
-version: 1.0
+version: 2.0
```
maha-hajja, line 1: why did we change all the test files to version 2.0? aren't we still supporting version 1.0 too?

pkg/provisioning/config/yaml/parser.go
```diff
@@ -36,69 +47,171 @@ type Configuration interface {
...
-		dec.WithHook(multiDecoderHook(
-			envDecoderHook,     // replace environment variables with their values
-			linter.InspectNode, // register fresh linter hook
-		))
-		err := dec.Decode(&cfg)
+		version, w, err := p.parseVersion(versionDecoder)
+		warn = append(warn, w...)
 		if err != nil {
-			// we reached the end of the document
+			// we probably reached the end of the document
```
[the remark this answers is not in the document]
lovromazgon: `io.EOF` means we reached the end of the document, but in this line we are not sure it's an `io.EOF` yet, that gets decided in the next line. I assume you thought this comment documents the line below, it's actually meant to document why we ended up in the current `if` clause. I hope that's not too confusing x)

pkg/provisioning/config/yaml/parser_test.go
lovromazgon: In this test we want to see how the v2 parser behaves if it encounters a v1 config. In other words, what happens if we take a v1 configuration and simply write `version: 2.0` without changing the structure of the configuration.

pkg/provisioning/test/pipelines1/provision1-pipeline-with-processors.yml
lovromazgon: To be honest, if I had the time I would go a step further, remove the test files and test this with a mocked parser that returns the configuration (we can do this now that we have the interface `config.Parser`). This would ensure we are testing the service code in isolation. After all, the service doesn't care about how the parser gets the configuration (from yaml, json, or if it makes it up 🙂), so it's not the job of these tests to ensure the parser works correctly. But I don't think it's worth it at this point.

We do support both versions, and that's still tested in the tests for `yaml.Parser`. It's just that for service tests we use the latest version now.

PR #951, opened 2023-03-17
 Add CI action for generated code

### LINKED ISSUE #45: Add CI action for generated code
We should add a Github action that runs `make generate` and succeeds if there is no diff. It should run on each push to a PR.

~We could do the same with proto files (generate them and make sure there's no diff).~ Proto files aren't generated in the repo anymore, we use Buf remote code generation now.

hariso:
### Description

Fixes #45 

### Quick checks:

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -21,6 +21,7 @@ jobs:
           export PATH=$PATH:$(go env GOPATH)/bin
           make install-tools
           make generate
+          make proto-generate
           git diff --exit-code --numstat
 
       - name: Test
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -16,5 +16,12 @@ jobs:
         with:
           go-version: '1.20'
 
+      - name: Generated files
+        run: |
+          export PATH=$PATH:$(go env GOPATH)/bin
+          make install-tools
+          make generate
+          git diff --exit-code --numstat
+
       - name: Test
         run: make test-integration GOTEST_FLAGS="-v -count=1"


.github/workflows/build.yml
```diff
@@ -16,5 +16,29 @@ jobs:
...
+
+      - name: Cache Go dependencies
+        uses: actions/cache@v3
+        with:
+          path: |
+            ~/.cache/go-build
+            ~/go/pkg/mod
+          key: ${{ runner.os }}-go-${{ hashFiles('**/go.sum') }}
+          restore-keys: |
+            ${{ runner.os }}-go-
```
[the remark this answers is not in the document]
hariso: Hmm, without this the action ran for about 2 minutes, with this caching it ran for about 1m. I can double check.

maha-hajja (approved):

.github/workflows/build.yml
```diff
@@ -16,5 +16,19 @@ jobs:
         with:
           go-version: '1.20'
 
+      - name: Install Protoc
```
lovromazgon, line 19: Not sure if you've seen my main comment in the previous review, I'll copy it here so we can also discuss it if needed.

I'd rather separate this action into its own workflow, it serves a different purpose.

hariso: Hey @lovromazgon! Sure, no problems. :+1:

.github/workflows/test.yml
```diff
@@ -0,0 +1,20 @@
+name: test
+
+on:
+  push:
+    branches: [ main ]
+  pull_request:
+
+jobs:
+  test:
```
lovromazgon, line 9: Better 👍  When this is merged please update the branch protection rules to require the "test" action to succeed instead of "build".

.github/workflows/generated-files.yml
```diff
@@ -1,12 +1,12 @@
-name: build
+name: validate-generated-files
```
lovromazgon, line 1: Nitpick: Can we use the same name for the workflow as the filename please? 🙏 Makes it easier to find it when debugging the action.

lovromazgon (approved):
Thanks for doing this 👍

PR #953, opened 2023-03-20
Glob patterns exclude double digits in tags

hariso:
### Description

Nightly build tags for releases where a version is >= 10 are wrong, because of the sorting used. This is a port of https://github.com/meroxa/cli/pull/656. Credits go to @janelletavares. 

### Quick checks:


lovromazgon (approved):
Nice catch. Not an imminent issue for us, but good thing we don't have to care about it 👍

hariso:
> Nice catch. Not an imminent issue for us, but good thing we don't have to care about it +1

Exactly!

PR #955, opened 2023-03-20
Pipeline configuration v2

### LINKED ISSUE #847: Bug: Non-deterministic order of processors in file provisioned pipelines

lovromazgon:
### Description

This PR introduces a new pipeline configuration format version v2. Includes work from 3 PRs which were reviewed separately:
- https://github.com/ConduitIO/conduit/pull/926
- https://github.com/ConduitIO/conduit/pull/928
- https://github.com/ConduitIO/conduit/pull/929

The new format fixes #847.

### Quick checks:


hariso (approved):
:rocket:

Issue #962, opened 2023-03-23
CLE: Trigger correct connector lifecycle events when provisioning configuration files


lovromazgon:
### Feature description

Correct connector lifecycle events are triggered when a connector is created, updated or deleted. The issue is that pipelines provisioned from configuration files are not updated. Instead, the pipeline is deleted and recreated with the new configuration, which triggers events `deleted` and `created` instead of `updated`, even if the configuration didn't actually change.

We need to make the provisioning logic smarter, detect if a configuration actually changed and then actually update the entity instead of recreating it.

PR #963, opened 2023-03-23
Trigger Lifecycle Events

### LINKED ISSUE #938: CLE: Trigger lifecycle events in Conduit
### Feature description

Trigger lifecycle methods introduced in #937. The actual moments when each lifecycle method should be triggered are described in the [design document](https://github.com/ConduitIO/conduit/blob/main/docs/design-documents/20230228-connector-lifecycle-methods.md). The triggers need to be backwards compatible with older plugins that don't implement connector lifecycle methods.

lovromazgon:
### Description

This PR adds triggers for lifecycle events in connectors. Note that when provisioning an existing pipeline from a configuration file (i.e. one that was already provisioned in a previous run) will not trigger the correct lifecycle events, this will be addressed in #962.

Fixes #938.

### Quick checks:


hariso (approved):
Great work!

PR #964, opened 2023-03-24
Exit Conduit if pipeline config file fails to be provisioned

### LINKED ISSUE #850: Feature: Exit Conduit if pipeline config file fails to be provisioned (flag)
### Feature description

If a Conduit instance is deployed with the purpose to run only pipelines provisioned through pipeline config files, it would be nice if Conduit would stop right away if it detects that a pipeline config file is invalid (an invalid connector plugin config probably also makes the file invalid), otherwise it will seem like the server is healthy while some pipeline might actually not work.

We could add a flag that enables this functionality, e.g. `conduit -pipelines.exit-on-provision-error` (naming is not final of course).

hariso:
### Description
TBD

Fixes #850.

### Quick checks:

### DIFF
--- a/cmd/conduit/main.go
+++ b/cmd/conduit/main.go
@@ -71,7 +71,16 @@ func parseConfig() conduit.Config {
 
 		connectorsDir = flags.String("connectors.path", "./connectors", "path to standalone connectors directory")
 
-		pipelinesDir = flags.String("pipelines.path", "./pipelines", "path to the directory that has the yaml pipeline configuration files, or a single pipeline configuration file")
+		pipelinesDir = flags.String(
+			"pipelines.path",
+			"./pipelines",
+			"path to the directory that has the yaml pipeline configuration files, or a single pipeline configuration file",
+		)
+		pipelinesStrictMode = flags.Bool(
+			"pipelines.strictMode",
+			false,
+			"enable strict mode for pipelines",
+		)
 	)
 
 	// flags is set up to exit on error, we can safely ignore the error
@@ -94,7 +103,7 @@ func parseConfig() conduit.Config {
 	cfg.Log.Format = strings.ToLower(stringPtrToVal(logFormat))
 	cfg.Connectors.Path = stringPtrToVal(connectorsDir)
 	cfg.Pipelines.Path = strings.ToLower(stringPtrToVal(pipelinesDir))
-
+	cfg.Pipelines.StrictMode = *pipelinesStrictMode
 	return cfg
 }
 
--- a/pkg/conduit/config.go
+++ b/pkg/conduit/config.go
@@ -58,7 +58,8 @@ type Config struct {
 	}
 
 	Pipelines struct {
-		Path       string
+		Path string
+		// todo there's probably a better name
 		StrictMode bool
 	}
 }
--- a/pkg/conduit/config.go
+++ b/pkg/conduit/config.go
@@ -58,7 +58,8 @@ type Config struct {
 	}
 
 	Pipelines struct {
-		Path string
+		Path       string
+		StrictMode bool
 	}
 }
 
--- a/pkg/conduit/runtime.go
+++ b/pkg/conduit/runtime.go
@@ -220,7 +220,8 @@ func (r *Runtime) Run(ctx context.Context) (err error) {
 
 	if r.Config.Pipelines.StrictMode {
 		r.pipelineService.OnFailure(func(e pipeline.FailureEvent) {
-			r.logger.Err(ctx, e.Cause).
+			r.logger.Warn(ctx).
+				Err(e.Cause).
 				Str(log.PipelineIDField, e.ID).
 				Msg("Conduit will be shut down due to a pipeline failure and strict mode enabled")
 			cancel()
@@ -236,6 +237,8 @@ func (r *Runtime) Run(ctx context.Context) (err error) {
 		multierror.ForEach(err, func(err error) {
 			r.logger.Err(ctx, err).Msg("provisioning failed")
 		})
+		r.logger.Warn(ctx).
+			Msg("Conduit will be shut down due to a pipeline failure and strict mode enabled")
 		cancel()
 	}
 
--- a/pkg/conduit/runtime.go
+++ b/pkg/conduit/runtime.go
@@ -232,7 +232,7 @@ func (r *Runtime) Run(ctx context.Context) (err error) {
 	if r.Config.Pipelines.StrictMode {
 		r.pipelineService.OnFailure(func(e pipeline.FailureEvent) {
 			r.logger.Err(ctx, e.Cause).
-				Str(log.PipelineIDField, e.Pipeline.ID).
+				Str(log.PipelineIDField, e.ID).
 				Msg("Conduit will be shut down due to a pipeline failure and strict mode enabled")
 			cancel()
 		})
--- a/pkg/conduit/runtime.go
+++ b/pkg/conduit/runtime.go
@@ -236,6 +236,7 @@ func (r *Runtime) Run(ctx context.Context) (err error) {
 		multierror.ForEach(err, func(err error) {
 			r.logger.Err(ctx, err).Msg("provisioning failed")
 		})
+		cancel()
 	}
 
 	err = r.pipelineService.Run(ctx, r.connectorService, r.processorService, r.pluginService)
--- a/pkg/conduit/runtime.go
+++ b/pkg/conduit/runtime.go
@@ -217,6 +217,15 @@ func (r *Runtime) Run(ctx context.Context) (err error) {
 	if err != nil {
 		return cerrors.Errorf("failed to init connector service: %w", err)
 	}
+
+	if r.Config.Pipelines.StrictMode {
+		r.pipelineService.OnFailure(func(e pipeline.FailureEvent) {
+			r.logger.Err(ctx, e.Cause).
+				Str(log.PipelineIDField, e.ID).
+				Msg("Conduit will be shut down due to a pipeline failure and strict mode enabled")
+			cancel()
+		})
+	}
 	err = r.pipelineService.Init(ctx)
 	if err != nil {
 		return cerrors.Errorf("failed to init pipeline service: %w", err)
@@ -229,14 +238,6 @@ func (r *Runtime) Run(ctx context.Context) (err error) {
 		})
 	}
 
-	if r.Config.Pipelines.StrictMode {
-		r.pipelineService.OnFailure(func(e pipeline.FailureEvent) {
-			r.logger.Err(ctx, e.Cause).
-				Str(log.PipelineIDField, e.ID).
-				Msg("Conduit will be shut down due to a pipeline failure and strict mode enabled")
-			cancel()
-		})
-	}
 	err = r.pipelineService.Run(ctx, r.connectorService, r.processorService, r.pluginService)
 	if err != nil {
 		multierror.ForEach(err, func(err error) {
--- a/pkg/conduit/runtime.go
+++ b/pkg/conduit/runtime.go
@@ -191,6 +191,7 @@ func newServices(
 // HTTP APIs. This function blocks until the supplied context is cancelled or
 // one of the services experiences a fatal error.
 func (r *Runtime) Run(ctx context.Context) (err error) {
+	ctx, cancel := context.WithCancel(ctx)
 	t, ctx := tomb.WithContext(ctx)
 	defer func() {
 		if err != nil {
@@ -228,6 +229,14 @@ func (r *Runtime) Run(ctx context.Context) (err error) {
 		})
 	}
 
+	if r.Config.Pipelines.StrictMode {
+		r.pipelineService.OnFailure(func(e pipeline.FailureEvent) {
+			r.logger.Err(ctx, e.Cause).
+				Str(log.PipelineIDField, e.Pipeline.ID).
+				Msg("Conduit will be shut down due to a pipeline failure and strict mode enabled")
+			cancel()
+		})
+	}
 	err = r.pipelineService.Run(ctx, r.connectorService, r.processorService, r.pluginService)
 	if err != nil {
 		multierror.ForEach(err, func(err error) {
--- a/pkg/orchestrator/orchestrator.go
+++ b/pkg/orchestrator/orchestrator.go
@@ -80,6 +80,9 @@ type PipelineService interface {
 	// requested.
 	Stop(ctx context.Context, pipelineID string, force bool) error
 
+	// OnFailure registers a handler for a pipeline.FailureEvent.
+	OnFailure(handler pipeline.FailureHandler)
+
 	List(ctx context.Context) map[string]*pipeline.Instance
 	Get(ctx context.Context, id string) (*pipeline.Instance, error)
 	Create(ctx context.Context, id string, cfg pipeline.Config, p pipeline.ProvisionType) (*pipeline.Instance, error)
--- a/pkg/pipeline/lifecycle.go
+++ b/pkg/pipeline/lifecycle.go
@@ -93,6 +93,20 @@ func (s *Service) Start(
 
 	s.logger.Debug(ctx).Str(log.PipelineIDField, pl.ID).Msg("starting pipeline")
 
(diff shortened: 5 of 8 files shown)


cmd/conduit/main.go
```diff
@@ -71,7 +71,16 @@ func parseConfig() conduit.Config {
...
-		pipelinesDir = flags.String("pipelines.path", "./pipelines", "path to the directory that has the yaml pipeline configuration files, or a single pipeline configuration file")
+		pipelinesDir = flags.String(
+			"pipelines.path",
+			"./pipelines",
+			"path to the directory that has the yaml pipeline configuration files, or a single pipeline configuration file",
+		)
+		pipelinesStrictMode = flags.Bool(
+			"pipelines.strictMode",
+			false,
+			"strict mode for pipelines (Conduit shuts down if a pipeline fails)",
```
maha-hajja, line 82: is strict mode only for the pipelines provisioned by config? if so, we should mention it here

pkg/conduit/config.go
```diff
@@ -59,6 +59,8 @@ type Config struct {
 
 	Pipelines struct {
 		Path string
+		// todo there's probably a better name
+		StrictMode bool
```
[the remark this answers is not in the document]
maha-hajja: I like `exitOnError` more tbh, i think `failFast` works too

pkg/conduit/runtime.go
```diff
@@ -217,6 +217,18 @@ func (r *Runtime) Run(ctx context.Context) (err error) {
...
 		return cerrors.Errorf("failed to init connector service: %w", err)
 	}
+
+	if r.Config.Pipelines.StrictMode {
+		r.pipelineService.OnFailure(func(e pipeline.FailureEvent) {
+			if e.ProvisionType == pipeline.ProvisionTypeConfig {
+				r.logger.Warn(ctx).
+					Err(e.Cause).
+					Str(log.PipelineIDField, e.ID).
+					Msg("Conduit will be shut down due to a pipeline failure and strict mode enabled")
```
maha-hajja, line 227: nitpick:
```suggestion
					Msg("Conduit will shut down due to a pipeline failure and strict mode enabled")
```

pkg/pipeline/lifecycle.go
```diff
@@ -93,6 +93,20 @@ func (s *Service) Start(
...
 	s.logger.Debug(ctx).Str(log.PipelineIDField, pl.ID).Msg("starting pipeline")
 
+	err = s.startInternal(ctx, pl, connFetcher, procFetcher, pluginFetcher)
+	if err != nil {
+		s.notify(pl.ID, pl.ProvisionedBy, err)
+	}
+	return err
+}
+
+func (s *Service) startInternal(
```
maha-hajja, line 103: can we add a one like comment to talk about this method plz?

pkg/pipeline/service.go
```diff
@@ -25,6 +25,15 @@ import (
 	"github.com/conduitio/conduit/pkg/foundation/metrics/measure"
 )
 
+type FailureEvent struct {
+	// ID is the ID of the pipeline which failed.
+	ID            string
+	ProvisionType ProvisionType
+	Cause         error
```
maha-hajja, line 32: `Cause` makes sense, but why not call it `Error`? I'm not that attached to it! just a suggestion

pkg/conduit/runtime.go
```diff
@@ -191,7 +191,7 @@ func newServices(
 // Run initializes all of Conduit's underlying services and starts the GRPC and
 // HTTP APIs. This function blocks until the supplied context is cancelled or
 // one of the services experiences a fatal error.
-func (r *Runtime) Run(ctx context.Context) (err error) {
+func (r *Runtime) Run(ctx context.Context, cancel context.CancelFunc) (err error) {
 	t, ctx := tomb.WithContext(ctx)
```
lovromazgon, line 195: The first thing we do is instantiate a tomb, so instead of requiring the caller to send us a cancel function we can simply kill the tomb with `t.Kill(reason)`, that will cancel the context. A nice thing is also that we can supply the error that caused the death of the tomb, which is then returned by `Run` and logged by Conduit as the cause for why it stopped.

pkg/conduit/runtime.go
```diff
@@ -227,6 +239,9 @@ func (r *Runtime) Run(ctx context.Context) (err error) {
 		multierror.ForEach(err, func(err error) {
 			r.logger.Err(ctx, err).Msg("provisioning failed")
 		})
+		r.logger.Warn(ctx).
+			Msg("Conduit will be shut down due to a pipeline failure and strict mode enabled")
+		cancel()
```
lovromazgon, line 244: I think last time we spoke about this we didn't quite decide on the default mode, but it seems harsh to have Conduit fail because of provisioning as the default and only behavior. I'll bring it up at standup.

pkg/pipeline/service.go
```diff
@@ -25,6 +25,15 @@ import (
...
+}
+
+type FailureHandler func(FailureEvent)
```
lovromazgon, line 35: I like the handler pattern, it keeps the logic for what to do on failures outside of the pipeline service 👍

pkg/pipeline/service.go
```diff
@@ -102,6 +112,15 @@ func (s *Service) Get(ctx context.Context, id string) (*Instance, error) {
 // Create will create a new pipeline instance with the given config and return
 // it if it was successfully saved to the database.
 func (s *Service) Create(ctx context.Context, id string, cfg Config, p ProvisionType) (*Instance, error) {
+	pl, err := s.createInternal(ctx, id, cfg, p)
+	if err != nil {
+		s.notify(id, p, err)
```
lovromazgon, line 117: I don't think we should notify on creation, this error will be returned by the provisioning service.

pkg/pipeline/lifecycle.go
```diff
@@ -623,6 +637,7 @@ func (s *Service) runPipeline(ctx context.Context, pl *Instance) error {
 			Str(log.PipelineIDField, pl.ID).
 			Msg("pipeline stopped")
 
+		s.notify(pl.ID, pl.ProvisionedBy, err)
```
lovromazgon, line 640: Maybe the logger statement confused you, but at this point err could be `nil`. We should only notify if `err` is not `nil`. The logger statement works, because it chooses the level based on the error - if `err` is `nil` it's INFO, otherwise it's ERROR.

pkg/orchestrator/orchestrator.go
```diff
@@ -80,6 +80,9 @@ type PipelineService interface {
 	// requested.
 	Stop(ctx context.Context, pipelineID string, force bool) error
 
+	// OnFailure registers a handler for a pipeline.FailureEvent.
+	OnFailure(handler pipeline.FailureHandler)
```
lovromazgon, line 84: Wait, do we need this? The orchestrator is not using this function, is it?

pkg/conduit/config.go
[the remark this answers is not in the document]
hariso: `exit-on-error` got most votes. I put `StrictMode` because it started to appear in the last call we had, so I thought it would be good to discuss it. Thanks for the inputs!

cmd/conduit/main.go
hariso: Agreed!

pkg/orchestrator/orchestrator.go
hariso: Not really, you're right. Tbh I didn't even check how's the pipeline service declared in `runtime.go`, I _assumed_ it's an interface (Java gets me from time to time), and this was the only one. I'll remove. :+1:

pkg/pipeline/lifecycle.go
hariso: I think I skimmed through the code and saw the two cases... I think I'll build the nil check into `notify`, it's easier to use.

pkg/pipeline/service.go
hariso: Yeah, let's let the pipeline service's boss decide.:D

pkg/pipeline/service.go
hariso: This is to cover the API-created pipelines as well.

pkg/pipeline/service.go
hariso: Error sounds good as well. :+1:

pkg/pipeline/service.go
lovromazgon: If we _really_ want to do that we should also make sure that we catch errors in the connectors and processors service. That said, I honestly think it's going too far if we do this, if I fail to _create_ a pipeline, connector or processor for some reason I really don't think Conduit should crash 😕 If I manage to start the pipeline and it fails because of an error that's one thing, but not if I make a mistake while creating the pipeline.

pkg/conduit/runtime.go
hariso: That was actually my first option. Then I realized that `tomb.WithContext` returns a new context, so the original one itself (given to `Run()`) cannot be canceled. But _now_ I see we're actually overwriting the input context. I guess that's done so that whoever is using the "main Conduit context" also gets notified of the cancellation when this tomb is killed. I had a concern that `Run` can mutate `ctx` in a way in which the original cancellation is lost (e.g. `tomb.WithContext(context.Background())`). But it looks like there's a test for this, so I think your suggestion makes a lot of sense.

pkg/conduit/runtime_test.go
hariso, line 1: Adding a unit test for the flag on its own was easy, but at the moment we can't have multiple unit tests for the runtime because of the way the Prometheus client registers collectors. It has an internal, global, registry of collectors, to which our own stats handler is added. Doing so twice is not allowed in the Prometheus library.

pkg/provisioning/config/parser.go
```diff
@@ -12,8 +12,6 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-//go:generate mockgen -destination=mock/parser.go -package=mock -mock_names=Parser=Parser . Parser
-
```
hariso, line 16: @lovromazgon I recall that you wanted to remove this in another PR. I've removed it here because it became a bit annoying seeing the generated but unused mock from time to time.:)

pkg/pipeline/service.go
ahmeroxa: I'd say this is taking it a step to far and is a pretty bad UX. I'd say we should stop if something fails to create, since that doesn't really address the _spirit_ of the initial request.

simonl2002: +1 on validate errors in creation via the API shouldn't just kill the process.

hariso: It's no problem to change, and I don't feel strongly about this. I do agree that this is pretty "brutal". 

However, WDYT about the same for pipeline config files? Using this example: because of the pipeline name, one of the pipelines won't be provided (current behavior in the main branch). That, together with the `exit-on-error` flag, makes Conduit crash.

Personally, I think the behavior as it is in this PR, is consistent (though very strict). I'm also assuming that this flag will typically be used in production environments where pipeline config files will mostly be used.

pkg/provisioning/config/parser.go
lovromazgon: Sure, it will also be removed in https://github.com/ConduitIO/conduit/pull/968 but I don't expect this to create a conflict.

pkg/pipeline/service.go
lovromazgon: I think these two ways to create the pipeline are fundamentally different in the sense of the feedback you get. The API gives you immediate feedback about the success of the actions you are performing, while provisioning the pipeline through a config file is "asynchronous", you need to check the logs or the API to see what happened. And as far as I understand, this issue should be addressing the second part + any errors that happen if a running pipeline fails.

pkg/pipeline/lifecycle.go
```diff
@@ -93,6 +94,20 @@ func (s *Service) Start(
 
 	s.logger.Debug(ctx).Str(log.PipelineIDField, pl.ID).Msg("starting pipeline")
 
+	err = s.startInternal(ctx, pl, connFetcher, procFetcher, pluginFetcher)
+	s.notify(pl.ID, pl.ProvisionedBy, err)
+	return err
```
lovromazgon, line 99: Connected to the discussion about stopping Conduit if a pipeline fails to be created, I'll argue we shouldn't stop if the pipeline is not _started_* successfully. This code will cause Conduit to stop if I create a pipeline without any connectors and attempt to start it. Since I'd get immediate feedback about this through the API response, I don't think this should crash Conduit.

*Note that "started" in this context only means that the pipeline nodes are successfully created, connected together with channels, and all goroutines are started. If one of those goroutines stops _right after_ the start because of an error (e.g. can't establish the connection to the database), that's when Conduit should stop (that's done down in line 641).

pkg/pipeline/service.go
hariso: The feedback perspective you're mentioning is a good one and I agree with that and is a good argument against being consistent in all aspects.:)

With everything mentioned above, I'll change this **not** to fail for API pipelines.

hariso:
@lovromazgon @maha-hajja @simonl2002 @ahmeroxa @lyuboxa 

I've changed the behavior of Conduit, so that with `--pipelines.exit-on-error=true` the following happens:

| scenario                      | API shuts down Conduit | configuration file shuts down Conduit |
|-------------------------------|------------------------|---------------------------------------|
| pipeline cannot be created(1) | false                  | true                                  |
| pipeline cannot be started(2) | false                  | true                                  |
| error while running(3)        | true                   | true                                  |
| malformed yaml file           | not applicable         | true                                  |


(1) e.g. duplicate pipeline name
(2) e.g. pipeline has no connectors
(3) e.g. record cannot be written to a destination

Please let me know if you have any objections to this.

cmd/conduit/main.go
```diff
@@ -71,7 +71,16 @@ func parseConfig() conduit.Config {
...
-		pipelinesDir = flags.String("pipelines.path", "./pipelines", "path to the directory that has the yaml pipeline configuration files, or a single pipeline configuration file")
+		pipelinesDir = flags.String(
+			"pipelines.path",
+			"./pipelines",
+			"path to the directory that has the yaml pipeline configuration files, or a single pipeline configuration file",
+		)
+		pipelinesExitOnError = flags.Bool(
+			"pipelines.exit-on-error",
+			false,
+			"Exit Conduit if any pipeline provisioned through a file experiences an error while running.",
```
lovromazgon, line 82: The description doesn't fit now, we should mention that Conduit also exits if any pipeline stops because of an error.

Also a nitpick: the description should start with a lowercase "e" and not end with "." to keep it consistent with other flags.

pkg/conduit/runtime.go
```diff
@@ -227,6 +237,12 @@ func (r *Runtime) Run(ctx context.Context) (err error) {
 		multierror.ForEach(err, func(err error) {
 			r.logger.Err(ctx, err).Msg("provisioning failed")
 		})
+		if r.Config.Pipelines.ExitOnError {
+			r.logger.Warn(ctx).
+				Err(err).
+				Msg("Conduit will shut down due to a pipeline provisioning failure and 'exit on error' enabled")
+			t.Kill(cerrors.Errorf("shut down due to 'exit on error' enabled: %w", err))
```
lovromazgon, line 244: Is there a reason why we don't just return the error at this point? We know we want Conduit to stop at this point.

pkg/pipeline/service.go
```diff
@@ -25,6 +25,15 @@ import (
 	"github.com/conduitio/conduit/pkg/foundation/metrics/measure"
 )
 
+type FailureEvent struct {
+	// ID is the ID of the pipeline which failed.
+	ID            string
+	ProvisionType ProvisionType
```
lovromazgon, line 31: `ProvisionType` seems to be redundant now.

pkg/conduit/runtime.go
hariso: I'm not sure I understand... Return the error from the handler function?

lovromazgon: This code is not in the handler.

hariso: Ah, right. Sure, we can do a return here.

PR #968, opened 2023-03-28
Update entities when provisioning config files

### LINKED ISSUE #962: CLE: Trigger correct connector lifecycle events when provisioning configuration files

lovromazgon:
### Description

This PR changes the behavior of the provisioning service to update entities instead of recreating them. This is needed to ensure that the correct connector lifecycle events are triggered.

A big chunk of the code changes in this PR constitutes new tests and refactoring of existing tests.

Fixes #962 

### Quick checks:


maha-hajja (commented):
lovely work with all the refactoring and test simplifying! most of these comments are nitpicks and method names suggestions, so nothing huge.
note: haven't reviewed the `import.go` file and its tests yet

pkg/provisioning/config/enrich.go
```diff
@@ -23,11 +25,33 @@ func Enrich(cfg Pipeline) Pipeline {
...
+	}
+	if dlq.WindowSize == nil {
+		tmp := pipeline.DefaultDLQ.WindowSize
+		dlq.WindowSize = &tmp
+	}
+	if dlq.WindowNackThreshold == nil {
+		tmp := pipeline.DefaultDLQ.WindowNackThreshold
+		dlq.WindowNackThreshold = &tmp
+	}
+	return dlq
```
hariso, line 52: Is there a special intention behind mutating the input value `dlq` and also returning it?

pkg/provisioning/config/enrich_test.go
```diff
@@ -111,8 +118,14 @@ func TestEnrich_DefaultValues(t *testing.T) {
 			Status:      "stopped",
 			Name:        "pipeline2",
 			Description: "empty",
-			Connectors:  []Connector{},
-			Processors:  []Processor{},
+			DLQ: DLQ{
```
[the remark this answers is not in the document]
hariso: It's due to different types. `DefaultDLQ` is `pipeline.DLQ`, whereas here we have `parser.DLQ`.

pkg/provisioning/test/pipelines1/pipelines.go
```diff
@@ -0,0 +1,160 @@
...
+import (
+	"time"
+
+	"github.com/conduitio/conduit/pkg/connector"
+	"github.com/conduitio/conduit/pkg/pipeline"
+	"github.com/conduitio/conduit/pkg/processor"
+	"github.com/conduitio/conduit/pkg/provisioning/config"
+)
+
+var P1Config = config.Pipeline{
```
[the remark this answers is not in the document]
hariso: Looks good! A question and a comment from me:
1. P1Config appears to be unused. So it's either GoLand not being able to find a usage or maybe there's another reason?
2. It would be good to have a comment in the yaml/Go files to remind that the Go files contain the expected Go values.

pkg/provisioning/config/enrich.go
```diff
@@ -23,11 +25,33 @@ func Enrich(cfg Pipeline) Pipeline {
...
+	}
+	if dlq.Settings == nil {
+		dlq.Settings = make(map[string]string, len(pipeline.DefaultDLQ.Settings))
+		for k, v := range pipeline.DefaultDLQ.Settings {
+			dlq.Settings[k] = v
+		}
+	}
+	if dlq.WindowSize == nil {
+		tmp := pipeline.DefaultDLQ.WindowSize
+		dlq.WindowSize = &tmp
```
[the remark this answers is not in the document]
hariso: It goes the other way around: if you change dlq.WindowSize, you'll change the default one as well. E.g. if you do `*dlq.WindowSize = 200`, the default one changes to 200 as well.

hariso (commented):
Great work, Lovro!

pkg/provisioning/import.go
```diff
@@ -0,0 +1,712 @@
...
+	"github.com/google/go-cmp/cmp/cmpopts"
+)
+
+// Import takes a pipeline config and imports it into Conduit.
+func (s *Service) Import(ctx context.Context, newConfig config.Pipeline) error {
+	oldConfig, err := s.Export(ctx, newConfig.ID)
+	if err != nil && !cerrors.Is(err, pipeline.ErrInstanceNotFound) {
+		s.logger.Warn(ctx).Err(err).Msgf("could not export pipeline with ID %v, trying to provision new pipeline despite the error", newConfig.ID)
+		// TODO does this behavior make sense? Should we rather return the error
+		//  and skip provisioning?
```
hariso, line 36: If there's an error, `oldConfig` should be an empty struct, so we're overwriting it, whatever it may be. `oldConfig` being empty means that lifecycle events would see an incorrect previous state, so cleanups may not happen.

pkg/provisioning/import.go
```diff
@@ -0,0 +1,712 @@
...
+
+	// first compare whole configs
+	if cmp.Equal(oldConfig, newConfig) {
+		// configs match, no need to do anything
+		return nil
+	}
+
+	// compare them again but ignore fields that can be updated, if configs
+	// are still different an update is not possible, we have to entirely
+	// recreate the processor
```
hariso, line 302: Thanks for the comment. :+1: In case we add new fields to the processor configs, how can we get notified that we need to ignore those as well?

pkg/provisioning/import.go
```diff
@@ -0,0 +1,712 @@
...
+	}
+}
+
+type action interface {
+	String() string
+	Do(context.Context) error
+	Rollback(context.Context) error
+}
+
+// --------------------
```
hariso, line 342: Nitpick: I think it would be good to have actions in a separate file.

pkg/provisioning/import.go
```diff
@@ -0,0 +1,712 @@
...
+func (a createPipelineAction) String() string {
+	return fmt.Sprintf("create pipeline with ID %v", a.cfg.ID)
+}
+func (a createPipelineAction) Do(ctx context.Context) error {
+	_, err := a.pipelineService.Create(ctx, a.cfg.ID, pipeline.Config{
+		Name:        a.cfg.Name,
+		Description: a.cfg.Description,
+	}, pipeline.ProvisionTypeConfig)
+	if err != nil {
+		return cerrors.Errorf("failed to create pipeline: %w", err)
```
hariso, line 361: It would be nice if we added the pipeline ID to the error.

pkg/provisioning/import.go
```diff
@@ -0,0 +1,712 @@
...
+		}
+	}
+
+	return nil
+}
+func (a createConnectorAction) Rollback(ctx context.Context) error {
+	err := a.connectorService.Delete(ctx, a.cfg.ID, a.pluginService)
+	if !cerrors.Is(err, connector.ErrInstanceNotFound) {
+		// ignore instance not found errors, this means the action failed to
+		// create the connector in the first place
```
hariso, line 442: +1 for the comment! At least to me, it's more useful to have them before the `if`, and not within the `if` block, because at this point we're not ignoring ErrInstanceNotFound.

pkg/provisioning/import.go
```diff
@@ -0,0 +1,712 @@
...
+	}
+
+	return nil
+}
+func (updatePipelineAction) isEqualConnectors(ids []string, connectors []config.Connector) bool {
+	if len(ids) != len(connectors) {
+		return false
+	}
+	for i := range ids {
+		if ids[i] != connectors[i].ID {
```
hariso, line 573: i.e. ordering of the IDs and connectors is important?

pkg/provisioning/import.go
```diff
@@ -0,0 +1,712 @@
...
+		Plugin:              cfg.DLQ.Plugin,
+		Settings:            cfg.DLQ.Settings,
+		WindowSize:          *cfg.DLQ.WindowSize,
+		WindowNackThreshold: *cfg.DLQ.WindowNackThreshold,
+	})
+	if err != nil {
+		return cerrors.Errorf("failed to update pipeline DLQ: %w", err)
+	}
+
+	// update connector IDs
```
hariso, line 532: Is this literally because of the connector IDs themselves, or is it actually about connectors? Also, isn't this supposed to be covered through actions?

pkg/provisioning/test/pipelines1/pipelines.go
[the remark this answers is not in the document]
lovromazgon: Oh good catch, I didn't notice it, we're actually only using the instances, I can get rid of the config variables. Sure I can add a comment to the yaml files as well 👍

pkg/provisioning/mock/provisioning_util.go
```diff
@@ -0,0 +1,40 @@
...
+import (
+	"github.com/conduitio/conduit/pkg/connector"
+	"github.com/conduitio/conduit/pkg/pipeline"
+	"github.com/conduitio/conduit/pkg/processor"
+	"github.com/golang/mock/gomock"
+)
+
+// ICreate is a utility function that lets you declare an expected call to
+// Create with arguments taken from the supplied instance.
+func (mr *PipelineServiceMockRecorder) ICreate(ctx interface{}, instance *pipeline.Instance) *gomock.Call {
```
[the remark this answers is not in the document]
lovromazgon: Not really a common name, in fact, it sounds like an interface in C# or Java 😅 I'll rename it 👍

pkg/provisioning/config/enrich.go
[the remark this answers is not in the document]
lovromazgon: Yep, that's why. I don't think we ever really change the value behind those pointers, but I wanted to be on the safe side and made sure they point to a unique value.

pkg/provisioning/mock/provisioning_util.go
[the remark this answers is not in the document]
hariso: Sorry, the weird interface names are only in C#.

pkg/provisioning/export.go
```diff
@@ -0,0 +1,129 @@
...
+		Status:      config.StatusStopped,
+		Name:        p.Config.Name,
+		Description: p.Config.Description,
+		Connectors:  nil, // extracted separately
+		Processors:  nil, // extracted separately
+		DLQ:         s.pipelineDLQToConfig(p.DLQ),
+	}
+}
+
+func (*Service) pipelineDLQToConfig(dlq pipeline.DLQ) config.DLQ {
```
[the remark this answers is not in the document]
lovromazgon: Makes sense!

pkg/provisioning/config/enrich.go
lovromazgon: The mutations are not visible to the caller, `DLQ` is passed in as a value, so we have to return it.

pkg/provisioning/import.go
```diff
@@ -0,0 +1,712 @@
...
+		s.logger.Warn(ctx).Err(err).Msgf("could not export pipeline with ID %v, trying to provision new pipeline despite the error", newConfig.ID)
+		// TODO does this behavior make sense? Should we rather return the error
+		//  and skip provisioning?
+	}
+
+	var actions []action
+	actions = append(actions, s.cascadingActionsDeleteOld(oldConfig, newConfig)...)
+	actions = append(actions, s.cascadingActionsUpsertNew(oldConfig, newConfig)...)
+
+	lastActionIndex, err := s.executeActions(ctx, actions)
```
hariso, line 43: The variable name `lastActionIndex` and what `executeActions` returns are not quite aligned. `executeActions` returns the number of actions, whereas the index would mean it's one less.

pkg/provisioning/import.go
lovromazgon: First of all, let's see when can we land in this condition - the only errors that `Export` can encounter are `*.ErrInstanceNotFound`. Now, if the pipeline is not found that's fine, in that case we just provision the new pipeline. But if a connector or processor is not found that's strange, it would mean that the pipeline references a connector or processor that doesn't exist. Essentially we would need to have a corrupted pipeline in Conduit to begin with.

You are right that we will overwrite the entities if they exist, because the pipeline/connector/processor `Service` doesn't check if the ID is already taken (this is done in the orchestrator and here in `Import` when exporting the existing pipeline). So we should be able to provision the new pipeline, _but_ there might be some old connectors or processors that are left in Conduit but not connected to the pipeline (the new pipeline won't point to them). I don't think they will interfere with the newly provisioned pipeline though, i.e. the new pipeline should be able to start and stop perfectly fine.

Connector states are not preserved in this case because connectors are _created_ and not _updated_.

To sum up: this only happens if the state is corrupted in the first place (pipeline references connectors/processors that don't exist). If the config is valid the pipeline should be provisioned correctly but there might be some corrupted state left over (i.e. dangling connectors and/or processors). Connector states are discarded as well.

pkg/provisioning/import.go
```diff
@@ -0,0 +1,712 @@
...
+		Type: processor.ParentTypePipeline,
+	}
+	for _, procNewConfig := range newConfig.Processors {
+		procOldConfig, _ := s.findProcessorByID(oldConfig.Processors, procNewConfig.ID)
+		actions = append(actions, s.prepareProcessorActions(procOldConfig, procNewConfig, parent)...)
+	}
+	return actions
+}
+
+// prepareActionsOldConfig prepares actions needed to delete existing entities
```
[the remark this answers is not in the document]
lovromazgon: Thanks for noticing, I'll correct the comment! The name sucks though, let me know if you by any chance think of a better one 🙂

pkg/provisioning/import.go
lovromazgon: That's a good question. By default, the fields will be treated as "non-updatable", they will trigger a recreation of the processor unless we add them here. I'm not sure what we could do to make this stand out 🤔 I could try to add a test that changes all fields _except_ the ones that should trigger a recreation (i.e. all updatable fields) and expects that this methods creates an update action. Then this test would start failing as soon as we add another field unless we either add that field to the test or to this code. WDYT?

pkg/provisioning/import.go
lovromazgon: Sure, how does `import_actions.go` sound?

pkg/provisioning/import.go
lovromazgon: Yes, I thought that since we now have deterministic ordering of both connectors and processors that we can make sure the IDs really are stored in the specified order. Note that this doesn't mean that the connectors will be updated if you change their order in the config file, only the `ConnectorIDs` and `ProcessorIDs` fields in the pipeline will be updated.

pkg/provisioning/import.go
lovromazgon: It's just the IDs, it has nothing to do with the connectors or processors behind these IDs. Each entity (pipeline, connector, processor) is treated completely in isolation in its own action.

pkg/provisioning/import.go
lovromazgon: Well if the error is not `nil` the index will point to the failing action. I could rename it to `failedActionIndex` if that sounds better?

pkg/provisioning/config/enrich.go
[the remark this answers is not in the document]
maha-hajja: yeah I was thinking that there is no case that we'd change the pointer value, but to be safe it makes sense!

pkg/provisioning/import.go
hariso: Sounds 100% great to me!

pkg/provisioning/import.go
hariso: Yup, it does!

pkg/provisioning/import.go
hariso: If the test doesn't take too much time to write, I think it's a good idea. Otherwise, I believe leaving a comment in the Config struct is good enough for now.

pkg/provisioning/import.go
lovromazgon: Actually now I remembered I did not include the ID on purpose because this error will be wrapped by [`executeActions`](https://github.com/ConduitIO/conduit/blob/3e44152a877971cfb123b7f56a74180d34e2731b/pkg/provisioning/import.go#L162), the wrapped error will include the name of the action which [includes the ID](https://github.com/ConduitIO/conduit/blob/3e44152a877971cfb123b7f56a74180d34e2731b/pkg/provisioning/import.go#L353) as well as what the action is trying to do (create, update, delete).

pkg/provisioning/import.go
lovromazgon: Btw, it's good you opened the discussion, that's exactly what I expected when I wrote that comment. If we agree that the behavior is what we want, I'll remove the comment and add an explanation what can happen in that case.

lovromazgon:
@hariso @maha-hajja I rebased the branch on top of main, the commits added since your review start with https://github.com/ConduitIO/conduit/pull/968/commits/3f1aaadea08f1b4c1295c3b07f50fd1f5e8e65a5.

pkg/provisioning/import.go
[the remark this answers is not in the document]
hariso: Besides maybe `prepareDeleteActions` (which isn't really amazing), I can't think of anything else tbh.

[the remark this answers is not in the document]
lovromazgon: I'd like to avoid "prepare", the functions below are already using that word (e.g. `preparePipelineActions`) and those only prepare actions for one entity. I wanted to make a distinction in the names for these two functions, as these are actually diving into all entities and call the "prepare actions" functions (that's why I used "cascading", "traverse" was another option I was thinking of).

That said, I'm ok with using "prepare" if we distinguish these from the other functions 🙂 Also I'm fine with renaming the other functions if needed.

pkg/provisioning/import.go
hariso: > First of all, let's see when can we land in this condition - the only errors that `Export` can encounter are `*.ErrInstanceNotFound` (`*` can be `pipeline`, `connector` or `processor`). Now, if the pipeline is not found that's fine, in that case we just provision the new pipeline. But if a connector or processor is not found that's strange, it would mean that the pipeline references a connector or processor that doesn't exist. Essentially we would need to have a corrupted pipeline in Conduit to begin with.

Thanks for this! My thinking is that: at this point we know and can't see any other errors besides `ABCNotFound`. That may change in future, at which point we might get unexpected results. I also can't think of any other error being returned besides `ABCNotFound`, so I think it would be good to add that to the method comment (i.e. that the method returns a nil error or `NotFound`).

pkg/provisioning/import.go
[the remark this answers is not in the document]
lovromazgon: I refactored the code and pushed it into the `actionsBuilder` which just exposes a `Build` method. These two methods were renamed to `buildForNewConfig` and `buildForOldConfig`, since it's obvious it's building actions, this makes the naming a bit better, but more importantly, the code is separated from the rest of the `Service` methods.

pkg/provisioning/import.go
hariso: We're on the same page!

PR #970, opened 2023-03-28
parsejson processor fix

maha-hajja:
### Description

for the parsejson processor, we accept the payload being nil, but if it's an empty rawData then it fails.
we should still convert rawData to structuredData if it was empty.

### Quick checks:

### DIFF
--- a/pkg/processor/procbuiltin/parsejson.go
+++ b/pkg/processor/procbuiltin/parsejson.go
@@ -35,18 +35,17 @@ func init() {
 
 // ParseJSONKey parses the record key from raw to structured data
 func ParseJSONKey(config processor.Config) (processor.Interface, error) {
-	return parseJSON(parseJSONKeyProcType, recordKeyGetSetter{}, config)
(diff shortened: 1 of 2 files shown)


pkg/processor/procbuiltin/parsejson.go
```diff
@@ -54,6 +54,11 @@ func parseJSON(
 		switch data.(type) {
 		case record.RawData:
 			var jsonData record.StructuredData
```
hariso, line 56: Nitpick: In the function's signature there's `config` which is unused. Credits go to GoLand.

hariso (approved):

pkg/processor/procbuiltin/parsejson.go
hariso: IMHO it's OK if it's left unused in those 2 methods (we don't _have_ to use it if we don't need a configuration), but where we don't need a parameter, we shouldn't put it. But again, really a nitpick.

PR #971, opened 2023-03-29
 Stream Inspector: Add metrics 

### LINKED ISSUE #742: Stream Inspector: Add metrics
### Feature description

Add metrics related to stream inspectors, for example, the number of open inspector sessions per connector/processor.

hariso:
### Description

Adds metrics for the number of inspector sessions per connector or processor.

Fixes #742.

### Quick checks:


pkg/inspector/inspector.go
```diff
@@ -114,17 +118,20 @@ func (i *Inspector) Send(ctx context.Context, r record.Record) {
 	}
 }
 
-func (i *Inspector) NewSession(ctx context.Context) *Session {
+func (i *Inspector) NewSession(ctx context.Context, componentID string) *Session {
```
hariso, line 121: As I was reviewing the PR myself, I realized it's better to have this in the inspector. After all, a single inspector manages sessions for only one component (connector or processor).

hariso: ... or no. Processors cannot be easily refactored. 

Processors themselves (i.e. `processor.Interface`) are created through builders. Builders (currently) only accept a processor configuration. Changing the builder would mean we'd change many places in the code.

Another approach is to create an inspector in `processor.Service.Create()`. A decorator would add inspection to all processors. `processor.Service` has all the important information needed for the metrics.

That would have a nice side-effect of simplifying the `procbuiltin.FuncWrapper` and `procjs.Processor` (it would reduce them to just one method). However, `processor.Store` is also building processors, so we'd need to duplicate the logic there as well.

All in all, I believe this is doable, but I think it would unnecessarily increase the scope of the PR. I'll try to refactor this in a separate one.

lovromazgon: I'd say that if we pour any amount of work into refactoring inspectors, it should be to move them to nodes and add an inspector wrapper node (or a normal inspector node that's added between other nodes). Then the logic would be contained to a single place, while connectors and processors wouldn't even know there is such a thing as an inspector. But I don't think this needs to be refactored right now, it's good enough and it works 👍

pkg/inspector/inspector.go
```diff
@@ -47,6 +50,7 @@ func (s *Session) close() {
 	s.once.Do(func() {
 		s.onClose()
 		close(s.C)
+		measure.InspectorsGauge.WithValues(s.componentID).Dec()
```
lovromazgon, line 53: If we move this call into `Session.onClose` (line 128) then we wouldn't have to store `componentID` in `Session`.

docs/metrics.md
```diff
@@ -19,19 +19,21 @@ locally, you can get metrics if you run `curl localhost:8080/metrics`.
...
+    | `conduit_dlq_bytes`                            | Histogram | Number of bytes* a DLQ connector processed per pipeline and plugin.                                            |
+    | `conduit_pipeline_execution_duration_seconds`  | Histogram | Amount of time records spent in a pipeline.                                                                    |
+    | `conduit_connector_execution_duration_seconds` | Histogram | Amount of time spent reading or writing records per pipeline, plugin and connector type (source, destination). |
+    | `conduit_processor_execution_duration_seconds` | Histogram | Amount of time spent on processing records per pipeline and processor.                                         |
+    | `conduit_dlq_execution_duration_seconds`       | Histogram | Amount of time spent writing records to DLQ connector per pipeline and plugin.                                 |
+    | `conduit_dlq_execution_duration_seconds`       | Histogram | Amount of time spent writing records to DLQ connector per pipeline and plugin.                                 |
```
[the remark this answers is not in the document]
hariso: Yup! Good catch!

PR #992, opened 2023-04-09
Update docs for built-in processors, improve filterfield processor

hariso:
### Description

1. The docs for a few processors mention how they handle raw data with or without schemas. In some of those cases, the code clearly shows that raw data isn't supported at all. I believe the comments were made to show how the processors **would** behave with schemas. This PR updates the docs to show how the processors currently behave.
2. `filterfield` processors use a function that panics if the input cannot be parsed. This PR uses a different function which returns an error.

### Quick checks:


lovromazgon (approved):
👍  We should rename the timestamp converter processor in the [docs PR](https://github.com/ConduitIO/conduit-site/pull/30).

hariso:
> +1 We should rename the timestamp converter processor in the [docs PR](https://github.com/ConduitIO/conduit-site/pull/30).

Good callout!

PR #995, opened 2023-04-11
Scripts for bumping versions

hariso:
### Description

Automates:
1.  the bumping of the SDK version in the built-in connectors
2. the bumping of the versions of built-in connectors in Conduit itself

### Quick checks:


scripts/bump-builtin-connectors.sh
```diff
@@ -0,0 +1,20 @@
+#!/bin/bash
+# Bumps the versions of built-in connectors
+
+git checkout main
+git pull origin main
+git checkout -b bump-builtin-connectors
+
+go get github.com/conduitio/conduit-connector-file@latest
```
maha-hajja, line 8: I liked the second script with the `conn` loop, I think it would make this script smaller and nicer, wdyt?

maha-hajja (commented):
these will definitely make our lives easier! thanks for taking the time on this

scripts/bump-builtin-connectors.sh
hariso: You're right, I have no idea why I went with this.:O

scripts/bump-sdk-in-connectors.sh
```diff
@@ -0,0 +1,40 @@
...
+fi
+
+SDK_V=$1
+
+for conn in 'file' 'kafka' 'generator' 's3' 'postgres' 'log'
+do
+	cd ../conduit-connector-$conn
+
+	echo
+	echo "Working on conduit-connector-$conn"
```
maha-hajja, line 25: this should work right?
```suggestion
	echo "\nWorking on conduit-connector-$conn"
```

maha-hajja (approved):
nice work!

scripts/bump-sdk-in-connectors.sh
lovromazgon: It should, but I don't think we should block because of this 😉 Given that you approved the PR I guess you won't insist on this, so I'll resolve the comment and merge, to not keep the PR open for 1 more week.

PR #996, opened 2023-04-11
Fix latest version for built-in plugins, bump built-in connectors

lovromazgon:
### Description

The issue is that `runtime/debug` build info is [not populated in tests](https://github.com/golang/go/issues/33976), so our plugins in tests use the constant defined in the connector code instead of the version taken from `go.mod`. The constant is set to `(devel)` for most built-in plugins.

This PR changes how we populate the built-in plugin registry, specifically the `latest` version for each plugin. We ensure that the `latest` tag will be set, even if the version in the plugin is not a semantic version.

This PR also bumps connectors to their latest versions.

### Quick checks:

- [X] I have written unit tests. (existing ones caught this)


hariso (approved):

go.mod
```diff
@@ -8,14 +8,14 @@ require (
 	github.com/NYTimes/gziphandler v1.1.1
 	github.com/antchfx/jsonquery v1.3.2
 	github.com/bufbuild/buf v1.17.0
-	github.com/conduitio/conduit-connector-file v0.4.0
-	github.com/conduitio/conduit-connector-generator v0.3.1
+	github.com/conduitio/conduit-connector-file v0.5.0
+	github.com/conduitio/conduit-connector-generator v0.4.0
 	github.com/conduitio/conduit-connector-kafka v0.4.1
 	github.com/conduitio/conduit-connector-log v0.1.2
 	github.com/conduitio/conduit-connector-postgres v0.3.0
```
hariso, line 15: Maybe run `go get` again?

PR #1000, opened 2023-04-13
Fix CI badge in readme

lovromazgon:
We renamed CI action "build" to "test" and forgot to update the badge.


maha-hajja (approved):
nice catch!

Issue #895, opened 2023-02-17
Version not reported in Docker images


hariso:
### Bug description

Version is not reported in Conduit's Docker images.

### Steps to reproduce

1. Run `docker run -p 8080:8080 ghcr.io/conduitio/conduit:v0.5.3`
2. Run `curl -Ss localhost:8080/info | jq .`

**Expected**
Version is reported.

**Actual**
OS and arch. reported, version isn't.
```json
{
  "version": "",
  "os": "linux",
  "arch": "amd64"
}

```

### Version

Docker image ghcr.io/conduitio/conduit:v0.5.3

lovromazgon:
Latest nightly build reports version `v0.7.0-nightly.20230517` 🎉

hariso:
> Latest nightly build reports version `v0.7.0-nightly.20230517` tada

Great news!

PR #1035, opened 2023-05-12
Don't ignore .git folder in docker build

lovromazgon:
### Description

This makes sure our docker build includes the `.git` folder, which should give us access to the git tag and let us build the version string 🤞

Fixes (hopefully) https://github.com/ConduitIO/conduit/issues/895

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,4 +1,3 @@
-.git
 .github
 .gitignore
 .golangci.goheader.template
@@ -11,3 +10,5 @@ conduit.db
 docs
 examples
 githooks
+scripts
+test


maha-hajja (approved):
let's hope it works 🤞🏼

hariso:
IIRC, it was removed to reduce Docker build time. WDYT about getting the version from git and saving it to a file, then sending it to Docker?

lovromazgon:
I just tried it out to see what difference it makes, I ran `docker build -f Dockerfile -t conduit-local .` and then removed the image two times in a row (second run caches almost everything excpet loading the context). I ran it on `main` and on this branch, here are the results:

- `main` first run 79.8s, second run 5.0s
- `lovro/dockerignore` first run 74.4s, second run 6.3s

I wouldn't say that's a noticeable difference and I don't think we should optimize this further if it solves the problem, it's just not worth it. Especially because our docker builds are triggered nightly and on releases, a few seconds don't make a difference there.

hariso:
Definitely, just one second of difference is not worth it.

PR #1036, opened 2023-05-16
Remove .dockerignore

lovromazgon:
### Description

This is a followup to https://github.com/ConduitIO/conduit/pull/1035. That PR made the situation better, but did not fix the issue entirely. The version is now reported, but has the suffix `-dirty`, most likely because some files are not added to the docker context.

This PR removes `.dockerignore` entirely. This will mean that all files in the repo are imported into the docker context and git should then report the correct version.

A fresh repo clone contains 35MB of data, `.dockerignore` currently ignores ~2.5MB of files, so the context size will increase by 7%. I think we can live with that.

Fixes (hopefully) #895.


maha-hajja (approved):
🤞🏼🤞🏼

maha-hajja (approved):
🤞🏼🤞🏼

PR #1075, opened 2023-06-13
doc: add mongodb connector

raulb:
### Description

Noticed the [MongoDB](https://github.com/conduitio-labs/conduit-connector-mongo) is not part of this list.

It was first added on 0.3.0, but it's 0.6.0 now, so unsure whether we should leave it on current version. 

### Quick checks:


hariso (approved):

hariso:
> It was first added on 0.3.0, but it's 0.6.0 now, so unsure whether we should leave it on current version.

The connector uses SDK v0.6, so we can leave 0.6 as you already have in your PR.

PR #1094, opened 2023-06-30
Enable nodejs in js processor

lovromazgon:
### Description

This adds support for [goja_nodejs](https://github.com/dop251/goja_nodejs) in the JavaScript processor. This enables us to use utilities like the buffer to manipulate raw data. For example, I used it in a pipeline to encode raw data as base64:
```
const Buffer = require("node:buffer").Buffer;

function process(record) {
    record.Payload.After.Raw = Buffer.from(record.Payload.After.Raw).toString("base64");
    return record;
}
```

### Quick checks:

- [ ] I have written unit tests. (if I get the time I'll add some later, did manually test though)


simonl2002 (approved):
lgtm

hariso (approved):

PR #1128, opened 2023-07-24
Add make lint

hariso:
### Description

Adds `make lint` which also installs the appropriate version of `golangci-lint`.

### Quick checks:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -33,7 +33,7 @@ golangci-lint-install:
 
 .PHONY: lint
 lint: golangci-lint-install
-	golangci-lint run -v --go $(shell awk '/^go/ {print $$2}' go.mod)
+	golangci-lint run -v
 
 build-server: check-go-version
 	go build -ldflags "-X 'github.com/conduitio/conduit/pkg/conduit.version=${VERSION}'" -o conduit ./cmd/conduit/main.go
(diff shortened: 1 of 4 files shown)


Makefile
```diff
@@ -25,6 +27,14 @@ test-integration:
 		docker compose -f test/docker-compose-postgres.yml -f test/docker-compose-schemaregistry.yml down; \
 		exit $$ret
 
+.PHONY: golangci-lint-install
+golangci-lint-install:
+	go install github.com/golangci/golangci-lint/cmd/golangci-lint@$(GOLANG_CI_LINT_VER)
+
+.PHONY: lint
+lint: golangci-lint-install
+	golangci-lint run -v --go $(shell awk '/^go/ {print $$2}' go.mod)
```
lovromazgon, line 36: I don't think we need to parse the go version manually, that seems to be the default behavior (see [this](https://golangci-lint.run/usage/configuration/) and [this](https://github.com/golangci/golangci-lint/blob/0b62feb8cafe98f045e2641298972113ac60e504/pkg/config/config.go#L59-L63)).

lovromazgon (approved):
Suggested one simplification, looks good otherwise.

Makefile
hariso: Thanks for the simplification!

PR #1168, opened 2023-08-07
Replace github.com/golang/mock with go.uber.org/mock

lovromazgon:
### Description

- Updates the code to use `go.uber.org/mock` instead of `github.com/golang/mock`
- Only files `pkg/plugin/standalone/v1/dispenser_test.go` and `pkg/plugin/builtin/v1/dispenser_test.go` are still using the old package because it's using a mock from connector protocol. Once we release a new version of the protocol we will be able to move to the new package.
- Add depguard rule to prevent further use of `github.com/golang/mock`


maha-hajja (approved):
LGTM

PR #1173, opened 2023-08-09
fix issue #453 add nil checkers for endpoints

### LINKED ISSUE #453: Bug: Conduit crashes when posting an empty JSON to a number of endpoints
### Bug description

When posting an empty JSON body to a number of endpoints, Conduit simply crashes. This can be seen with the connectors, processors and pipelines endpoints. I haven't checked others. 

### Steps to reproduce

Run `curl 'localhost:8080/v1/connectors' -d '{}'`

**Expected results**
Conduit replies back with a `400 Bad Request`.

**Actual results**
Conduit crashes. Following is seen in logs:
```
2022-06-14T19:12:30+00:00 INF request processed duration=0.046111 grpc_method=/api.v1.ConnectorService/CreateConnector grpc_status_code=OK http_endpoint="POST /v1/connectors" request_id=b0f5e1f7-5965-4b61-84e5-13401db1ea5f
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x28 pc=0x10d6bf3]

goroutine 27 [running]:
github.com/conduitio/conduit/pkg/web/api/fromproto.ConnectorConfig(...)
	/app/pkg/web/api/fromproto/connector.go:35
github.com/conduitio/conduit/pkg/web/api.(*ConnectorAPIv1).CreateConnector(0x3?, {0x2c04508?, 0xc0004728a0?}, 0x0?)
	/app/pkg/web/api/connector_v1.go:97 +0x73
github.com/conduitio/conduit/proto/api/v1._ConnectorService_CreateConnector_Handler.func1({0x2c04508, 0xc0004728a0}, {0x14eea40?, 0xc000672a20})
	/app/proto/api/v1/api_grpc.pb.go:566 +0x78
```

### Version

{   "version": "v0.3.0-nightly.20220614-dirty",   "os": "linux",   "arch": "amd64" }

AdamHaffar:
### Description

Adds nil checkers to catch empty configs from empty JSON post requests.

Fixes #453

### Quick checks:

- [x] I have written unit tests. (Manual testing)

### DIFF
--- a/pkg/web/api/fromproto/connector.go
+++ b/pkg/web/api/fromproto/connector.go
@@ -31,8 +31,11 @@ func ConnectorType(in apiv1.Connector_Type) connector.Type {
 }
 
 func ConnectorConfig(in *apiv1.Connector_Config) connector.Config {
-	return connector.Config{
-		Name:     in.Name,
-		Settings: in.Settings,
+	if in != nil {
+		return connector.Config{
+			Name:     in.Name,
+			Settings: in.Settings,
+		}
 	}
+	return connector.Config{}
 }
--- a/pkg/web/api/fromproto/pipeline.go
+++ b/pkg/web/api/fromproto/pipeline.go
@@ -20,10 +20,13 @@ import (
 )
 
 func PipelineConfig(in *apiv1.Pipeline_Config) pipeline.Config {
-	return pipeline.Config{
-		Name:        in.Name,
-		Description: in.Description,
+	if in != nil {
+		return pipeline.Config{
+			Name:        in.Name,
+			Description: in.Description,
+		}
 	}
+	return pipeline.Config{}
 }
 
 func PipelineDLQ(in *apiv1.Pipeline_DLQ) pipeline.DLQ {
--- a/pkg/web/api/fromproto/processor.go
+++ b/pkg/web/api/fromproto/processor.go
@@ -27,15 +27,21 @@ func _() {
 }
 
 func ProcessorConfig(in *apiv1.Processor_Config) processor.Config {
-	return processor.Config{
-		Settings: in.Settings,
-		Workers:  int(in.Workers),
+	if in != nil {
+		return processor.Config{
+			Settings: in.Settings,
+			Workers:  int(in.Workers),
+		}
 	}
+	return processor.Config{}
 }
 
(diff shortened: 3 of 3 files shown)


pkg/web/api/fromproto/pipeline.go
```diff
@@ -20,10 +20,13 @@ import (
...
+	if in != nil {
+		return pipeline.Config{
+			Name:        in.Name,
+			Description: in.Description,
+		}
 	}
+	return pipeline.Config{}
 }
 
 func PipelineDLQ(in *apiv1.Pipeline_DLQ) pipeline.DLQ {
```
lovromazgon, line 32: Seems like a `nil` check is missing for this one.

pkg/web/api/fromproto/connector.go
```diff
@@ -31,8 +31,11 @@ func ConnectorType(in apiv1.Connector_Type) connector.Type {
 }
 
 func ConnectorConfig(in *apiv1.Connector_Config) connector.Config {
-	return connector.Config{
-		Name:     in.Name,
-		Settings: in.Settings,
+	if in != nil {
```
lovromazgon, line 34: Nitpick: We generally strive to keep the "happy path" to the left (i.e. not indented) and return early in case of unexpected conditions. So I'd rather turn these statements around and check if `in == nil`.

pkg/web/api/fromproto/pipeline.go
```diff
@@ -20,13 +20,19 @@ import (
...
+	}
 	return pipeline.Config{
 		Name:        in.Name,
 		Description: in.Description,
 	}
 }
 
 func PipelineDLQ(in *apiv1.Pipeline_DLQ) pipeline.DLQ {
+	if in == nil {
+		return pipeline.DefaultDLQ
```
lovromazgon, line 34: This would change the behavior. This line assumes that a `nil` pipeline DLQ means the same as the default DLQ configuration (not really true). That should not be the case, this code should only convert from one type to the other without any logic.

TL;DR: we should return `pipeline.DLQ{}` instead.

lovromazgon (commented):
One last thing, otherwise looks good 👍

lovromazgon (approved):

PR #1177, opened 2023-08-14
Remove Assert Package 1/3

### LINKED ISSUE #260: Remove assert package
We currently have the package `pkg/foundation/assert` that provides simple assert functions for tests. We should remove it and use https://github.com/matryer/is instead.

AdamHaffar:
### Description

Refactor all `_test.go` that used `pkg/foundation/assert` to use [(https://github.com/matryer/is)](https://github.com/matryer/is) in the following packages:

### PR (1/3)
- [x] `./pipeline`
- [x] `./record`
- [x] `./conduit`
- [x] `./web`

### PR (2/3)
- [ ] `./processor` 

### PR (3/3)
- [ ] `./foundation`
- [ ] `./orchestrator` 

Partially Fixes #260

### Quick checks:

- [x] I have written unit tests. Can you even write tests for tests :o


pkg/pipeline/store_test.go
```diff
@@ -124,13 +128,13 @@ func TestConfigStore_Delete(t *testing.T) {
...
 	err = s.Delete(ctx, want.ID)
-	assert.Ok(t, err)
+	is.NoErr(err)
 
 	got, err := s.Get(ctx, want.ID)
-	assert.Error(t, err)
-	assert.True(t, cerrors.Is(err, database.ErrKeyNotExist), "expected error for non-existing key")
-	assert.Nil(t, got)
+	is.True(err != nil)
+	is.True(cerrors.Is(err, database.ErrKeyNotExist))
```
maha-hajja, line 138: you can add the reason as an inline comment, which will be logged when this statement fails
```suggestion
	is.True(cerrors.Is(err, database.ErrKeyNotExist)) // expected error for non-existing key
```

maha-hajja: also, please solve the conflicts so the workflows and tests could run

AdamHaffar: when running the test, should that sepecific test fail?

pkg/conduit/config_test.go
```diff
@@ -15,13 +15,15 @@
 package conduit
 
 import (
+	"github.com/matryer/is"
```
hariso, line 18: Imports are organized in groups, with a blank line between them. The first group is always the standard lib stuff: https://github.com/golang/go/wiki/CodeReviewComments#imports.

pkg/pipeline/stream/base_test.go
```diff
@@ -16,78 +16,84 @@ package stream
...
 
 func TestPubSubNodeBase_PubTwice(t *testing.T) {
+	is := is.New(t)
 	n := &pubSubNodeBase{}
 	n.Pub() // first one succeeds
 
 	defer func() {
-		assert.NotNil(t, recover())
+		is.True(recover() != nil)
+		is.True(recover() == nil)
```
hariso, line 60: Why do we check it twice though?

AdamHaffar: mistake on my behalf!

pkg/record/schema/proto/mutable_test.go
```diff
@@ -106,23 +114,27 @@ func TestMutableSchema_SetDescriptors_Panics(t *testing.T) {
 
 			ms := fileDescriptorSetToMutalbeSchema(t, getFileDescriptorSet(t, standaloneDescriptorSetPath))
 			ms.SetDescriptors(tc.descriptors)
-			assert.True(t, false, "expected panic")
+			is.True(false)
```
hariso, line 117: It'd be good to keep the message `expected panic`, because otherwise `is.True(false)` on its own would be confusing.

pkg/record/schema/proto/mutable_test.go
```diff
@@ -282,17 +306,19 @@ func TestMutableField_SetDescriptor_Primitive(t *testing.T) {
...
 
 			newSchema, err := ms.Build()
-			assert.Ok(t, err)
+			is.NoErr(err)
 
 			got := newSchema.Descriptors()[0].(StructDescriptor).Fields()[0].Descriptor()
 			d, ok := got.(PrimitiveDescriptor)
-			assert.True(t, ok, "expected %T, got %T", d, got)
-			assert.Equal(t, tc, d.Type())
+			is.True(ok)
```
hariso, line 313: I suggest adding a comment what was expected.

pkg/record/schema/proto/mutable_test.go
```diff
@@ -697,13 +747,15 @@ func TestMutablePrimitiveDescriptor_Type(t *testing.T) {
 	for index, wantType := range wantTypes {
 		t.Run(fmt.Sprintf("f%d", index+1), func(t *testing.T) {
 			pd, ok := allTypesDesc.Fields()[index].Descriptor().(*MutablePrimitiveDescriptor)
-			assert.True(t, ok, fmt.Sprintf("expected %T, got %T", pd, allTypesDesc.Fields()[index].Descriptor()))
-			assert.Equal(t, wantType, pd.Type())
+			is.True(ok)
```
hariso, line 750: Similarly to above, I think it's worth adding a comment what was expected.

pkg/record/schema/proto/schema_test.go
```diff
@@ -214,41 +226,41 @@ func TestStructDescriptor_Fields(t *testing.T) {
...
 				case 18:
 					_, ok := f.Descriptor().(EnumDescriptor)
-					assert.True(t, ok, "expected %T, got %T", EnumDescriptor{}, f.Descriptor())
+					is.True(ok)
 
 				default:
 					// first 15 fields should be primitive types
 					_, ok := f.Descriptor().(PrimitiveDescriptor)
-					assert.True(t, ok, "expected %T, got %T", PrimitiveDescriptor{}, f.Descriptor())
+					is.True(ok)
```
hariso, line 263: Alternatively, we could also use `is.Equal(reflect.TypeOf(PrimitiveDescriptor{}), reflect.TypeOf(f.Descriptor()))

pkg/record/schema/proto/mutable_test.go
```diff
@@ -114,7 +114,7 @@ func TestMutableSchema_SetDescriptors_Panics(t *testing.T) {
 
 			ms := fileDescriptorSetToMutalbeSchema(t, getFileDescriptorSet(t, standaloneDescriptorSetPath))
 			ms.SetDescriptors(tc.descriptors)
-			is.True(false)
+			is.True(false) //expected panic
```
lovromazgon: IMO it would be clearer with `is.Fail() // expected panic`.

hariso: Agreed. :+1:

hariso (approved):
LGMT!

PR #1185, opened 2023-08-24
Fix make target install-tools

lovromazgon:
A very simple change fixing the `install-tools` target. I noticed that the target complains about the imports pointing to non-importable packages (`main`). I added `-e` to `go tools` which ignores those errors.


maha-hajja (approved):
I upgraded too and just tested it and had the same error, so LGTM!

PR #1201, opened 2023-09-07
update http processor request

### LINKED ISSUE #1199: use the whole OpenCDC record in the http processor
### Feature description

currently, http processor sends a request with the `Payload.After` field only
we want it to send the whole openCDC record instead, in JSON format, with a header (content-type = application/json)

maha-hajja:
### Description

* Update the http processor request to send the whole record (json formatted), instead of only sending `Payload.After`
* Configurable content-type header
* Filter record out if response header is (204 No Content)

Fixes #1199 

### Quick checks:


pkg/processor/procbuiltin/httprequest.go
```diff
@@ -110,6 +125,10 @@ func httpRequest(
 			// regard status codes over 299 as errors
 			return record.Record{}, cerrors.Errorf("%s: invalid status code %v (body: %q)", processorType, resp.StatusCode, string(body))
 		}
+		// skip if body has no content
+		if resp.StatusCode == http.StatusNoContent {
+			return record.Record{}, processor.ErrSkipRecord
+		}
```
hariso, line 131: Is this change somehow related to the change in payload? The HTTP processor may simply be triggering an action/event and not actually saving/updating resources, so a 204 may mean that an action was successfully triggered. i.e. it doesn't have to necessarily mean that a record needs to be filtered out.

maha-hajja: it's not related to the payload change no. It's honestly just something that we should agree on.
So if the action was successful, do you mean that we should just send back an empty record? without the error?

lovromazgon: My 2 cents - I think we won't be able to make the processor so generic that it would work with any endpoint, we'll probably need to set some conventions and expectations, like this one for example. I don't see a nicer way of letting the endpoint filter the record out. Well apart from making this configurable, i.e. this behavior would be the default but the user could configure a custom status code (or list of codes) that cause the record to be filtered. Although that can still be added in the future, I wouldn't block the PR if this configuration is not included.

pkg/processor/procbuiltin/httprequest.go
```diff
@@ -33,6 +34,8 @@ const (
 
 	httpRequestConfigURL          = "url"
 	httpRequestConfigMethod       = "method"
+	httpRequestConfigContentType  = "contentType"
+	httpRequestContentType        = "Content-Type"
```
lovromazgon, line 38: Nitpick: if you ask me, I'd write the header string directly in code, imo it would make it easier to read. Then again, the `application/json` default content type could be a constant, since you wouldn't have to start reading the processor to figure out what the default is.

pkg/processor/procbuiltin/httprequest.go
```diff
@@ -110,6 +125,10 @@ func httpRequest(
 			// regard status codes over 299 as errors
 			return record.Record{}, cerrors.Errorf("%s: invalid status code %v (body: %q)", processorType, resp.StatusCode, string(body))
 		}
+		// skip if body has no content
+		if resp.StatusCode == http.StatusNoContent {
+			return record.Record{}, processor.ErrSkipRecord
+		}
 
 		r.Payload.After = record.RawData{Raw: body}
```
lovromazgon, line 133: I'm thinking, if it would make sense to move `r.Payload.After` to `r.Payload.Before`? 🤔 Maybe, maybe not, I'm opening the discussion to see what you think 🙂

pkg/processor/procbuiltin/httprequest.go
hariso: > hmm after thinking about it, if we send an empty record and the empty record continues to the destination, then the position won't be acked. if we skip the record in the processor using this error, then the processor will make sure that it's acked

Good point! I can't remember from the top of my head, but I think we allow processors to change any part of a record, including the key and position? That might break some thing as well, so maybe we need to limit the changes a processor can make?

hariso: > My 2 cents - I think we won't be able to make the processor so generic that it would work with any endpoint, we'll probably need to set some conventions and expectations, like this one for example. 

I agree 100%!

> I don't see a nicer way of letting the endpoint filter the record out. 

I agree, and I see the analogy between a `nil` in our Go code meaning that a record should be filtered and `204 No Content` in the HTTP world. My only concern is that the endpoint can be about triggering an action, in which case a `204 No Content` doesn't really say the record should go away. 

On the other hand, if the endpoint is OpenCDC-aware we can maybe safely assume it knows about what empty records mean in processing.

Anyway, I don't feel too strongly about this, and I don't think it's a blocker.

pkg/processor/procbuiltin/httprequest.go
hariso: It's definitely an interesting idea... It looks like, if we decide to do it, it should probably be something common to all processors, so maybe we discuss it in details in "Better Processors"?

maha-hajja: it is interesting and does make sense! 
I agree with Haris about discussing this with the "better processors' work

pkg/processor/procbuiltin/httprequest.go
hariso: Well, I'd rather say that the point of processing (and not the record itself) can be to trigger an action. How frequently that happens, and how often do you need to trigger an action and then send the record to a destination -- I don't know to be honest.

hariso (approved):

pkg/processor/procbuiltin/httprequest.go
maha-hajja: we can discuss this more during the standup today if you'd like that. 
But since it's not a blocker I would prefer merging this PR to not block the mdpx work

PR #1205, opened 2023-09-11
Remove Assert Package 2/3

### LINKED ISSUE #260: Remove assert package

AdamHaffar:
### Description

Refactor all `_test.go` that used `pkg/foundation/assert` to use [(https://github.com/matryer/is)](https://github.com/matryer/is) in the following packages:

### PR (1/3)
- [x] `./pipeline`
- [x] `./record`
- [x] `./conduit`
- [x] `./web`

### PR (2/3)
- [x] `./processor` 
- [x] `./orchestrator` 

### PR (3/3)
- [ ] `./foundation`

Partially Fixes #260

### Quick checks:

- [x] I have written unit tests. Can you even write tests for tests :o

### DIFF
--- a/pkg/processor/procbuiltin/extractfield_test.go
+++ b/pkg/processor/procbuiltin/extractfield_test.go
@@ -19,10 +19,10 @@ import (
 	"reflect"
 	"testing"
 
-	"github.com/conduitio/conduit/pkg/foundation/assert"
 	"github.com/conduitio/conduit/pkg/processor"
 	"github.com/conduitio/conduit/pkg/record"
 	"github.com/conduitio/conduit/pkg/record/schema/mock"
+	"github.com/matryer/is"
 )
 
 func TestExtractFieldKey_Build(t *testing.T) {
@@ -68,6 +68,8 @@ func TestExtractFieldKey_Build(t *testing.T) {
 }
 
 func TestExtractFieldKey_Process(t *testing.T) {
+	is := is.New(t)
+
 	type args struct {
 		r record.Record
 	}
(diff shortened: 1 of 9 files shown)


pkg/processor/procbuiltin/extractfield_test.go
```diff
@@ -135,7 +137,7 @@ func TestExtractFieldKey_Process(t *testing.T) {
 	for _, tt := range tests {
 		t.Run(tt.name, func(t *testing.T) {
 			underTest, err := ExtractFieldKey(tt.config)
-			assert.Ok(t, err)
+			is.NoErr(err)
```
hariso, line 140: The `t` in this test is a different `t` from one given here: `func TestExtractFieldKey_Process(t *testing.T)`. Hence, it would be better to initialize `is` in this subtest.

maha-hajja (commented):
since this is the last PR to remove the assert package, let's add the package name to the `depguard` linter for golangci-lint, to make sure no other file will have it in the future https://github.com/ConduitIO/conduit/blob/31e283db4884003e4427e8bedbd598f3a3e250e7/.golangci.yml#L7

lovromazgon:
Why don't we just remove the package itself? Then there's nothing to import, nothing to guard against 🙂

maha-hajja:
@lovromazgon lol totally forgot that we're the ones who created it, let's do what he said 🤣

hariso (approved):

PR #1229, opened 2023-10-04
Pipeline created from API gets overwritten by config file with duplicate ID

### LINKED ISSUE #837: Bug: Config file pipeline overwrites pipeline provisioned by API
### Bug description

If Conduit contains a pipeline provisioned through the API and it loads a pipeline from a config file with the same ID it will overwrite the pipeline.

The expected behavior is to see an error that a pipeline with the same ID was already provisioned through the API so it won't be updated.

**Note:** I would estimate the impact to be relatively low, since the API creates UUID pipeline IDs, while I expect most pipeline configuration files to have human-readable pipeline IDs.

### Steps to reproduce

1. Run Conduit and create a pipeline through the API.
2. Copy the ID (uuid) of the pipeline and create a configuration file with a pipeline with the same ID.
3. Restart Conduit and make sure it loads the configuration file (e.g. define `-pipelines.path` flag).
4. See that the pipeline was overwritten.

### Version

v0.4.0

AdamHaffar:
### Description

Adds a check in provisioning/service.go for any Pipelines in the pipelineService with the same ID as any of the pipelines in configs.

Refactors other tests in service_test.go to account for extra call of .Get() method.

Fixes #837 

### Quick checks:


pkg/provisioning/errors.go
```diff
@@ -17,5 +17,6 @@ package provisioning
 import "github.com/conduitio/conduit/pkg/foundation/cerrors"
 
 var (
-	ErrDuplicatedPipelineID = cerrors.New("duplicated pipeline ID")
+	ErrDuplicatedPipelineID    = cerrors.New("duplicated pipeline ID")
+	ErrDuplicatedAPIPipelineID = cerrors.New("duplicated pipeline ID already exists from API")
```
lovromazgon, line 21: I wouldn't name this "duplicated ID", it's more like we can't mutate the pipeline using a config file because it was created through the API. Essentially the counterpart to [this error](https://github.com/ConduitIO/conduit/blob/4be951a5374bbcc7212dd85c62706c0458679edd/pkg/orchestrator/errors.go#L24).

pkg/provisioning/service.go
```diff
@@ -102,6 +102,20 @@ func (s *Service) Init(ctx context.Context) error {
 		allPls = append(allPls, duplicateID)
 	}
 
+	// remove pipelines with duplicate IDs from API pipelines
+	var existingPipelinesFromAPI []int
+	for i, pl := range configs {
+		_, err := s.pipelineService.Get(ctx, pl.ID)
+		// if err is nil then a pipeline exists already, so we add towards duplicate ID list
```
lovromazgon, line 109: It's fine if the pipeline already exists, as long as it was provisioned using a config file. If it was provisioned by anything else, we don't want to mutate it.
```suggestion
		pl, err := s.pipelineService.Get(ctx, pl.ID)
		// check error ...
		if pl.ProvisionedBy == pipeline.ProvisionTypeConfig {
			continue // all good
		}
		// Not ok! We don't want to provision this pipeline, it was provisioned by something else.
		// ...
```

pkg/provisioning/service.go
```diff
@@ -102,6 +102,20 @@ func (s *Service) Init(ctx context.Context) error {
...
+		if err == nil {
+			multierr = cerrors.Errorf("pipelines with ID %q will be skipped: %w", pl.ID, ErrDuplicatedAPIPipelineID)
```
hariso, line 111: We should append errors together: `multierror.Append(multierr, err)`. Same thing on L114 below.

lovromazgon (approved):
LGTM 👍

PR #1242, opened 2023-10-17
Validate connector name and pipeline description through API

### LINKED ISSUE #574: Connector names validated in UI but not in the API
-  The validation of "name" key is implemented on UI, but we can create connectors without name in API test using Postman.

- Set the character limit of description to 250 characters in /v1/pipelines

AdamHaffar:
### Description

Adds validation for the following config values:

- Adds validation for Connector Names so they can't be empty strings
- Sets a character limit of 250 for Pipeline Descriptions 

Fixes #574

### Quick checks:

- [x] I have performed manual testing through API


pkg/web/api/pipeline_v1.go
```diff
@@ -124,6 +124,9 @@ func (p *PipelineAPIv1) CreatePipeline(
 	if err != nil {
 		return nil, status.PipelineError(cerrors.Errorf("failed to create pipeline: %w", err))
 	}
+	if len(created.Config.Description) > 250 {
```
lovromazgon, line 127: 250 sounds too low for the pipeline description IMO. Even our very simple [example pipeline](https://github.com/ConduitIO/conduit/blob/7cd6af09fa8f15a3cd6665a37034522b6fe33b03/examples/pipelines/file-to-file.yml#L6-L9) has a description length of 237 characters. I see the description as a field where you can describe what the pipeline is doing, for complex pipelines this could get quite long. I'd maybe go with a limit of 8192 (2^13), this should give plenty of space to the user to describe the pipeline while giving us the confidence that the description won't consume more than 8kB.

WDYT @hariso @maha-hajja @jmar910 @ahamidi? I know that longer descriptions might look bad in the UI right now, but I think we could encounter them in pipeline configuration files, the UI can always be adjusted.

pkg/web/api/connector_v1.go
```diff
@@ -128,10 +128,12 @@ func (c *ConnectorAPIv1) CreateConnector(
 		req.PipelineId,
 		fromproto.ConnectorConfig(req.Config),
 	)
-
 	if err != nil {
 		return nil, status.ConnectorError(cerrors.Errorf("failed to create connector: %w", err))
 	}
+	if created.Config.Name == "" {
```
lovromazgon, line 134: Maybe we should add a limit for the names as well? How does 64 characters sound? I would then apply this to both a pipeline name and a connector name.

And since we're adding validations, it might be good to add some for the ID as well. I propose a limit of 64 characters and maybe even a restriction of allowed characters like `[A-Za-z0-9\-\_]`. Thoughts, @hariso @maha-hajja @jmar910 @ahamidi?

lovromazgon (commented):
I would move these validations at least into the orchestrator, if not into the actual services. For example, we already check that the pipeline has a [name](https://github.com/ConduitIO/conduit/blob/7cd6af09fa8f15a3cd6665a37034522b6fe33b03/pkg/pipeline/service.go#L114-L115) here, I'd keep these validations there. This way the check is executed even if Conduit is used as a library (without endpoints).

pkg/web/api/pipeline_v1.go
maha-hajja: I think the 8192 limit sounds fair, not too long yet long enough to explain a somewhat complicated pipeline.

pkg/web/api/connector_v1.go
maha-hajja: sounds good to me

hariso: +1 from me as well.

pkg/pipeline/service.go
```diff
@@ -326,3 +329,26 @@ func (s *Service) notify(pipelineID string, err error) {
 		handler(e)
 	}
 }
+func (s *Service) validatePipeline(cfg Config, id string) error {
```
lovromazgon, line 332: I like how you extracted the validation logic into its own function. Can we do the same in the connector service please? 🙏

lovromazgon (commented):
We're getting there, I think after these comments are resolved we should be good to go 👍

pkg/connector/persister_test.go
```diff
@@ -184,7 +184,7 @@ func TestPersister_WaitsForOpenConnectorsAndFlush(t *testing.T) {
 	persister.Wait()
 
 	// we are testing a delay which is not exact, this is the acceptable margin
-	maxDelay := delay + time.Millisecond*10
+	maxDelay := delay + time.Millisecond*11
```
lovromazgon, line 187: Umm, do we need this change?

pkg/connector/service.go
```diff
@@ -275,3 +284,27 @@ func (s *Service) SetState(ctx context.Context, id string, state any) (*Instance
...
+
+	if cfg.Name == "" {
+		multierr = multierror.Append(multierr, ErrNameMissing)
+	}
+	if len(cfg.Name) > 64 {
+		multierr = multierror.Append(multierr, ErrNameOverLimit)
+	}
+	if id == "" {
+		multierr = multierror.Append(multierr, ErrIDMissing)
+	} else {
```
lovromazgon, line 299: If we change the regex to allow any number of characters (`^[A-Za-z0-9-_:]*$`) we don't need this `else`, which would make the function a bit more readable, as all validations are independently executed for all strings. Same in the pipeline validation.

pkg/pipeline/errors.go
```diff
@@ -17,13 +17,18 @@ package pipeline
...
-	ErrNameAlreadyExists   = cerrors.New("pipeline name already exists")
-	ErrConnectorIDNotFound = cerrors.New("connector ID not found")
-	ErrProcessorIDNotFound = cerrors.New("processor ID not found")
+	ErrTimeout              = cerrors.New("operation timed out")
+	ErrGracefulShutdown     = cerrors.New("graceful shutdown")
+	ErrPipelineRunning      = cerrors.New("pipeline is running")
+	ErrPipelineNotRunning   = cerrors.New("pipeline not running")
+	ErrInstanceNotFound     = cerrors.New("pipeline instance not found")
+	ErrNameMissing          = cerrors.New("must provide a pipeline name")
+	ErrIDMissing            = cerrors.New("must provide a connector ID")
```
lovromazgon, line 26:
```suggestion
	ErrIDMissing            = cerrors.New("must provide a pipeline ID")
```

lovromazgon (approved):
Good work! I'm approving in advance, even though I left a few more nitpicks. Thanks @AdamHaffar 👍

PR #1247, opened 2023-10-20
[Perf] Clone record without reflection

### LINKED ISSUE #885: Optimize record cloning
### Feature description

Record cloning has been implemented (#849) with [copier](https://github.com/jinzhu/copier/), which uses reflection under the hood. This could become a performance bottleneck with high-volume streams.

The reason for this is: the after and before fields in `Record` can be `StructuredData` which is map[string]interface{}. Implementing a deep copy option for `interface{}` would take some time, so for the mentioned PR we chose to first get the functionality fixed (especially because it's easy to reproduce) versus optimizing (for high volume streams).

lovromazgon:
### Description

This PR optimizes record cloning in Conduit so it does not use reflection.

Fixes #885

### Quick checks:

### DIFF
--- a/pkg/record/record.go
+++ b/pkg/record/record.go
@@ -17,6 +17,7 @@
 package record
 
 import (
+	"bytes"
 	"encoding/base64"
 	"encoding/json"
 	"fmt"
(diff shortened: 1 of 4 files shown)


pkg/record/record.go
```diff
@@ -220,6 +233,18 @@ func (d StructuredData) Bytes() []byte {
 	return b
 }
 
+func (d StructuredData) Clone() Data {
+	cloned := make(map[string]any, len(d))
```
nassor, line 237: What are the odds for this structured data to have the same amount of elements?

nassor (approved):

PR #1248, opened 2023-10-20
[Perf] Inspector refactoring and optimization

lovromazgon:
### Description

The inspector was allocating quite a lot of data even if no sessions were open. Now it's optimized for the hot path when there are no open sessions.

```
goos: darwin
goarch: amd64
pkg: github.com/conduitio/conduit/pkg/inspector
cpu: Intel(R) Core(TM) i9-9880H CPU @ 2.30GHz
                                │  bench.old  │              bench.new               │
                                │   sec/op    │    sec/op     vs base                │
Inspector_NoSession_Send-16       31.96n ± 2%    14.70n ± 1%  -54.01% (p=0.000 n=10)
Inspector_SingleSession_Send-16   92.88n ± 1%   122.15n ± 2%  +31.52% (p=0.000 n=10)
Inspector_10Sessions_Send-16      447.6n ± 2%    447.1n ± 2%        ~ (p=0.895 n=10)
geomean                           109.9n         92.94n       -15.46%
```

Note that the single session benchmark is slower now because of the use of `Record.Clone()`, although that usage is now correct, since we previously did not clone all fields. Note that cloning of records was sped up in #1247.

### Quick checks:

### DIFF
--- a/pkg/inspector/inspector.go
+++ b/pkg/inspector/inspector.go
@@ -27,29 +27,14 @@ import (
 
 const DefaultBufferSize = 1000
 
-// Session wraps a channel of records and provides:
-// 1. a way to send records to it asynchronously
-// 2. a way to know if it's closed or not
+// Session represents a single inspector session. Records are continuously sent
+// into channel C. If the buffer of C is full, records will be dropped. C will
+// be closed once the session is removed from the inspector.
 type Session struct {
 	C chan record.Record
 
 	id          string
 	componentID string
-	logger      log.CtxLogger
-}
-
-// send a record to the session's channel.
-// If the channel has already reached its capacity,
-// the record will be ignored.
-func (s *Session) send(ctx context.Context, r record.Record) {
-	select {
-	case s.C <- r:
-	default:
-		s.logger.
-			Warn(ctx).
-			Str(log.InspectorSessionID, s.id).
-			Msg("session buffer full, record will be dropped")
-	}
 }
 
 // Inspector is attached to an inspectable pipeline component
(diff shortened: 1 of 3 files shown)


pkg/inspector/inspector.go
```diff
@@ -87,81 +79,84 @@ func New(logger log.CtxLogger, bufferSize int) *Inspector {
...
-		<-ctx.Done()
-		s.logger.
-			Debug(context.Background()).
-			Msgf("context done: %v", ctx.Err())
-		s.close()
-	}()
+}
 
+// add a session with given ID to this Inspector.
+func (i *Inspector) add(s *Session) {
```
nassor, line 129: _[nit]_ Do we need to send the whole struct (and have a pointer of a pointer) or could be just `id string` (or `sessionID string`).

pkg/inspector/inspector_benchmark_test.go
```diff
@@ -30,3 +38,14 @@ func BenchmarkInspector_SingleSession_Send(b *testing.B) {
 		ins.Send(context.Background(), record.Record{Position: record.Position("test-pos")})
 	}
 }
+
+func BenchmarkInspector_10Sessions_Send(b *testing.B) {
+	ins := New(log.Nop(), 10)
+	for i := 0; i < 10; i++ {
+		ins.NewSession(context.Background(), "test-id")
+	}
+
```
nassor, line 47: Calling `b.ResetTimer()` zeroes the elapsed benchmark time and memory allocation counters since the beginning of the test. This way, an expensive setup can be discarded from the test results. That will make the benchmark results more precise, you should add that here and in all other benchmarks that need a setup before the benchmark.

* https://pkg.go.dev/testing#B.ResetTimer

pkg/inspector/inspector.go
lovromazgon: Well, we need the whole struct as we're storing it, and we need to know the whole thing if we want to later remove it. `Session` could actually be passed as a value instead of a pointer, but after changing it I see the same heap allocations 😕 So I'm not sure if it's worth changing it, especially since the change would cascade down into other layers all the way to the orchestrator (e.g. [here](https://github.com/ConduitIO/conduit/blob/b7ab8484e03612fa45ff41dbc0e7349a23aa8204/pkg/orchestrator/processors.go#L109)).

lovromazgon: Btw, to clarify, the session creation is not in the hot path. The actual optimization in this PR is coming from the behavior of `Inspector.Send` where we have a shortcut to skip the mutex and record cloning if there are no sessions. That is in the hot path and actually makes a difference.

nassor (approved):

PR #1249, opened 2023-10-20
[Perf] Lazy encode connectors in persister

lovromazgon:
### Description

With this change the connector persister lazily encodes connectors only when they are flushed. The connector data is still copied at the time the connector requests to be persisted. This is currently required because of the locking mechanism (see https://github.com/ConduitIO/conduit/pull/857 for more info), there is still room for improvement here as most copies will be thrown away, especially on fast pipelines.

This also changes the library used for encoding to [github.com/goccy/go-json](https://github.com/goccy/go-json), which is quite a lot faster.

### Quick checks:

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -21,6 +21,7 @@ require (
 	github.com/dop251/goja v0.0.0-20230531210528-d7324b2d74f7
 	github.com/dop251/goja_nodejs v0.0.0-20230602164024-804a84515562
 	github.com/gammazero/deque v0.2.1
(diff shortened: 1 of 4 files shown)


nassor (approved):
I also believe that it is faster but It is missing the benchmarks to support the claims.

lovromazgon:
I didn't have time at this point to create benchmarks for this, but rest assured, I tested it manually and there was about a 7% increase in throughput 😉

PR #1250, opened 2023-10-20
[Perf] Optimize stream nodes fanin and fanout

lovromazgon:
### Description

The nodes `FanIn` and `FanOut` were optimized in this PR. `FanIn` now doesn't use any reflection in case the pipeline consists of up to 6 sources. Beyond that amount, we fall back to reflection again. `FanOut` on the other hand uses a shortcut if there is only 1 destination - in that case it doesn't clone the record as there is no need.

### Quick checks:

### DIFF
--- a/pkg/pipeline/stream/fanin.go
+++ b/pkg/pipeline/stream/fanin.go
@@ -16,7 +16,6 @@ package stream
 
 import (
 	"context"
-	"reflect"
 )
 
 type FaninNode struct {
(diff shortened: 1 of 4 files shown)


pkg/pipeline/stream/fanin.go
```diff
@@ -82,6 +64,36 @@ func (n *FaninNode) Run(ctx context.Context) error {
 	}
 }
 
+func (n *FaninNode) trigger(ctx context.Context) func() (*Message, error) {
+	in := make([]<-chan *Message, len(n.in))
+	copy(in, n.in)
```
nassor, line 69: I love to see it.

nassor (approved):

Issue #997, opened 2023-04-11
"Better Processors"


neovintage:
### Feature description

Our plan is for the better processor work for 0.8.0 to include.

- [x] #1284
- [x] #1285
- [x] #1286
- [x] #1158
- [x] #1162
- [x] #1156
- [x] #1287
- [x] #1288
- [x] #1289
- [x] #1290
~- [ ] #1291~
- [x] #1292
~- [ ] Update documentation~

What we currently aren't planning on including in this first pass:
- discovery of processor types (stretch goal)
- open cdc unwrap
- permanent storage
- Splitting and combining records
- Embedded processors

simonl2002:
This likely won't be in 0.7.0 but the goal is to start the work on this.

nassor:
Extra features descriptions:
* [ ] Load Processor as plugins as we do with connectors. With that, we could create custom processors for deployments.
* [ ] Have the Processor interface method `Close` called when Conduit is turning off, so we can release external resources allocated by the Processor
* [ ] Open Telemetry Support: send & return [`SpanContext`](https://opentelemetry.io/docs/specs/otel/trace/api/#spancontext) to the processor (and connectors as well).
* [ ] Simple Key/Value storage for processors.

lovromazgon:
Relevant discussion: https://github.com/ConduitIO/conduit/discussions/1141

raulb:
Documentation will be tackled in this other issue https://github.com/ConduitIO/conduit/issues/1422.

Closing this issue since yesterday "better processors" was shipped as part of [0.9.0](https://github.com/ConduitIO/conduit/releases/tag/v0.9.0).

PR #1279, opened 2023-11-17
Better processors design document

lovromazgon:
This PR contains the design document around better processors.

Related to #997 and #1141.


docs/design-documents/20231117-better-processors.md
```diff
@@ -0,0 +1,889 @@
...
+a round hole and ended up with sub-par processors.
+
+This document tries to address these issues holistically and create a better experience
+around processors.
+
+### Goals
+
+- Keep existing processors to preserve backward compatibility.
+  - Old processors should only be deprecated but continue to work.
+- Add the ability to run a processor only on records that match a certain condition.
```
hariso, line 27: i.e. we will have a "quick" way to filter out records (through the method proposed, a template) and then it's also possible to write a standalone processor which can filter out records as well?

docs/design-documents/20231117-better-processors.md
```diff
@@ -0,0 +1,889 @@
...
+  same time, we propose to design the interface in a way that will let us add
+  that functionality in the future without breaking changes. This means that the
+  function will for now be called only with a single record.
+- Notice that the return value is a slice of type `ProcessedRecord`. The purpose
+  behind this is to be able to figure out what happened to each input record. This
+  way we can associate each input record to an output record and know what happened
+  to it (was it filtered out, split, merged, or only transformed). This distinction
+  is important because of acknowledgments (e.g. if a record is split into multiple
+  records, Conduit will acknowledge the original record only once all split records
+  are acknowledged).
```
hariso, line 122: Given that a ProcessedRecord will somehow be associated with the original record, we should somehow call that out and/or provide functions in the Processors SDK that help do that. It's not really part of the design, but calling it out since I believe it will be important (because not properly associating processed records with input records can lead to unexpected behavior).

docs/design-documents/20231117-better-processors.md
```diff
@@ -0,0 +1,889 @@
...
+
+We have to differentiate listing actual processors that are attached to actual pipelines
+and listing the specifications of all available processors. We can draw a parallel to
+listing connectors vs. listing connector plugins. We already have a mechanism to list
+actual processors using
+[`ProcessorService.ListProcessors`](https://github.com/ConduitIO/conduit/blob/df048d93e6240c2c3ec5e52af430c05bdfc7a1a3/proto/api/v1/api.proto#L713),
+we need to add a way to list available processors.
+
+Since we are adding pluggable processors we propose to use the same terminology as with
+connectors and introduce **processor plugins**. This would also give us the ability to
```
hariso, line 145: I really like how much of the connector design we can re-use here!

docs/design-documents/20231117-better-processors.md
```diff
@@ -0,0 +1,889 @@
...
+scheme as connectors (more on
+[referencing connectors](https://conduit.io/docs/connectors/referencing)). Built-in
+processors would be created in a similar manner as built-in connectors, in that they
+would implement the same interface and could essentially be extracted into standalone
+processors (more info on
+[standalone vs built-in connectors](https://conduit.io/docs/connectors/behavior#standalone-vs-built-in-connectors)).
+
+The only difference between pluggable processors and connectors is the proposed plugin
+mechanism. For pluggable processors we are leaning towards using
+[WebAssembly](https://webassembly.org). We are not decided yet which runtime to use, but
```
hariso, line 197: WebAssembly because of the support for different languages?

docs/design-documents/20231117-better-processors.md
lovromazgon: I imagined the association to be done by the index of the record in the slice. So if you get 2 records `[rec1, rec2]` and want to filter out the first one, you should return a slice with `[filtered, rec2]` (we need to still see how we implement this `filtered` entity). This allows us to support splitting and merging records in the future and associating the merged/split record with a specific input record.

docs/design-documents/20231117-better-processors.md
lovromazgon: Yes, I think it's better than introducing yet another concept.

docs/design-documents/20231117-better-processors.md
lovromazgon: Yes, it seems to fit the processor paradigm quite well. It's relatively lightweight, provides good performance, and we can manage access to resources.

docs/design-documents/20231117-better-processors.md
hariso: So in the case of splitting records, we'd get a ProcessedRecord, which actually has a slice of records in it? And for a merged record in the example you mentioned, we'd get something like `[rec1, merged]`?

hariso: (Btw, I don't think it's a must to have this discussed and nailed right away, we can decide on the details in the implementation.)

docs/design-documents/20231117-better-processors.md
```diff
@@ -0,0 +1,889 @@
...
+We already have a similar functionality in the `filterfieldkey` processor which lets you
+filter a record if some condition is met. The idea is to extract this functionality and
+create a generalized way to create conditions on processors. If the condition evaluates
+to `false`, the processor is skipped and the record is passed along to the next node
+without being processed.
+
+This way we can simplify the `filter` processor and, more importantly, give the user
+the control of defining when a processor should be executed.
+
+We propose adding a field `condition` in the processor configuration that allows you to
```
hariso, line 229: And I guess a failure to evaluate the condition is the same as any other failure in the pipeline and counts towards the DLQ quota?

docs/design-documents/20231117-better-processors.md
maha-hajja, line 27: was this a requirement from the platform team? 
does it mean that the processor won't process the record at all (according to a given condition)? or that the processor will check the record and discard if it doesn't match the condition

docs/design-documents/20231117-better-processors.md
```diff
@@ -0,0 +1,889 @@
...
+  behind this is to be able to figure out what happened to each input record. This
+  way we can associate each input record to an output record and know what happened
+  to it (was it filtered out, split, merged, or only transformed). This distinction
+  is important because of acknowledgments (e.g. if a record is split into multiple
+  records, Conduit will acknowledge the original record only once all split records
+  are acknowledged).
+- If an error happens in `Process` it should be stored in the `ProcessedRecord` that
+  was being processed when the error occurred. In that case the processor has two
+  options on how to continue processing the remaining records in the slice.
+  1. If the processor has no side effects and cheap it can process the remaining
```
maha-hajja, line 126: are we the ones deciding if a processor is cheap or not?

docs/design-documents/20231117-better-processors.md
lovromazgon: No, I came up with it 🙃 I realized that it would be useful to be able to have a condition on any processor, not only the `filter` processor. Notice that the processor implementation wouldn't even be aware of this condition, Conduit itself would evaluate it and decide if the processor should be called or skipped. For us it's not much more work, but for the user it's potentially a very useful feature.

docs/design-documents/20231117-better-processors.md
lovromazgon: I think so, yes.

docs/design-documents/20231117-better-processors.md
```diff
@@ -0,0 +1,889 @@
...
+to the processor plugin registry. Processors would be addressable using a similar naming
+scheme as connectors (more on
+[referencing connectors](https://conduit.io/docs/connectors/referencing)). Built-in
+processors would be created in a similar manner as built-in connectors, in that they
+would implement the same interface and could essentially be extracted into standalone
+processors (more info on
+[standalone vs built-in connectors](https://conduit.io/docs/connectors/behavior#standalone-vs-built-in-connectors)).
+
+The only difference between pluggable processors and connectors is the proposed plugin
+mechanism. For pluggable processors we are leaning towards using
```
lyuboxa, line 196: Does this mean all processors will be WASM?

docs/design-documents/20231117-better-processors.md
lovromazgon: Yes, the more I think about it, we should just have option 2, return early and let Conduit decide what to do.

docs/design-documents/20231117-better-processors.md
lovromazgon: I assume by "all" you mean "standalone and built-in"? Then no, IMO we should have a similar mechanism as we do with connectors, where they can be run both as built-in or pluggable connectors. Built-in connectors skip the gRPC layer, so similarly built-in processors would skip the WASM layer. Something like this:

image

docs/design-documents/20231117-better-processors.md
```diff
@@ -0,0 +1,889 @@
...
+plan to handle cases when we don't expect a string as an output. For instance,
+conditions need to evaluate to a boolean value. We propose to simply use the package
+`strconv`, specifically the function
+[`strconv.ParseBool`](https://pkg.go.dev/strconv#ParseBool) to convert the result to
+a boolean value. This means that values 1, t, T, TRUE, true, True are converted to
+`true`. In addition, any value that is not 0, f, F, FALSE, false, False will return
+an error, which we should also regard as `false`, but additionally log a warning.
+
+Note that we still need to evaluate if `text/template` is actually the implementation
+we want to use to evaluate these expressions. If the performance is too slow we might
```
lyuboxa, line 281: I do have a feeling that the template matching may add some latency, though it does depend on how well the template machine performs.

docs/design-documents/20231117-better-processors.md
```diff
@@ -0,0 +1,889 @@
...
+of `.Payload.After`. The result is a valid JSON object which will be converted back
+to structured data before stored in field `.Payload.After`.
+
+```yaml
+processors:
+  id:   example
+  type: set.field
+  settings:
+    field: ".Payload.After"
+    value: '{"hoisted":{{ .Payload.After }}}'
```
lyuboxa, line 412: At which level does `hoisted` live? Same as `.Payload.After`?

e.g.

```
{
  "hoisted": "..."
  "Payload": ...
}
```

vs 

```
{
  "Payload": {
     "After": ...
    "hoisted": ...
  }
}
```

docs/design-documents/20231117-better-processors.md
maha-hajja: oh this sounds really useful! nice idea :D

docs/design-documents/20231117-better-processors.md
```diff
@@ -0,0 +1,889 @@
...
+
+### `field.rename`
+
+Rename a group of fields. It is not allowed to rename top-level fields (`.Operation`,
+`.Position`, `.Key`, `.Metadata`, `.Payload.Before`, `.Payload.After`), only fields
+that are nested.
+
+#### Configuration
+
+- `mapping` (required) - comma separated key:value pairs, where the key is the target
```
lyuboxa, line 483: To clarify for `map: k2:k1` : `{k1:v1}` -> `{k1:v1, k2:v1}`?

docs/design-documents/20231117-better-processors.md
```diff
@@ -0,0 +1,889 @@
...
+  type: field.rename
+  settings:
+    fields: ".Key.foo:bar,.Payload.Before.id:key,.Metadata.topic:subject"
+```
+
+### `field.flatten`
+
+Flatten takes a field that is structured or a map and flattens all nested values into
+top level values. The keys leading to a nested value are concatenated with a separator,
+by default the separator is a dot (`.`).
```
lyuboxa, line 500: how do we handle this if fields already have special characters? e.g `my.key { v: y}` ->`my\.key.v => y` ?

docs/design-documents/20231117-better-processors.md
```diff
@@ -0,0 +1,889 @@
...
+
+We can also collapse only part of the structured payload and specify a custom separator:
+
+```yaml
+processors:
+  id:   example
+  type: field.flatten
+  settings:
+    field: ".Payload.After.foo"
+    separator: "-"
```
lyuboxa, line 532: `separator` 👍

lyuboxa (approved):

docs/design-documents/20231117-better-processors.md
```diff
@@ -0,0 +1,889 @@
...
+[text/template](https://pkg.go.dev/text/template). This way we are leaning on syntax
+that is thoroughly documented and has a lot of resources online outside of the
+Conduit ecosystem. Additionally, we already use the same approach in specifying the
+record [output format](https://conduit.io/docs/connectors/output-formats).
+
+Here are a few examples of addressing record fields:
+
+- `.` - references the whole record
+- `.Payload.After` - references the field that normally carries the payload
+- `{{ eq .Key.foo "123" }}` - evaluates to `true` if the key contains the value
```
maha-hajja, line 262: I think `eq` evaluates `==`, so it's true if they're equal not if it contains the value

docs/design-documents/20231117-better-processors.md
lyuboxa: @lovromazgon ok but where does the WASM layer live?;-)

hariso (approved):

docs/design-documents/20231117-better-processors.md
lovromazgon: Yes, so splitting a record would mean create a `ProcessedRecord` that itself contains multiple records so we can associate them all with the original record.

For merging I'm not exactly sure. When merging it would be nice if we could associate the merged record with all records that were merged. Once that merged record gets acked, we forward the ack for all original records that were merged into this rec. So, to get back to the example, if you get records `[rec1, rec2]` and you merge them, the result should probably be two merged records `[merged, merged]` that are somehow associated with each other, so Conduit knows it's the same record for both original records.

As you said, the details should be decided once we actually get to merging / splitting of records.

docs/design-documents/20231117-better-processors.md
lovromazgon: A part of it is in Conduit (the host), where we expose any utility functions we need and run the WASM binary, the other part of the equation is the Processor SDK which nicely exposes host functions and makes it generally simple to write a processor.

docs/design-documents/20231117-better-processors.md
lovromazgon: No, this one doesn't copy fields, it renames existing ones, so in your example `map: k2:k1` : `{k1:v1}` -> `{k2:v1}`. Your example would be doable using `field.set`.

docs/design-documents/20231117-better-processors.md
lovromazgon: Good question, since we're going with the Go template syntax you can always use `{{ index myMap "my key in a string form"}}`.

Issue #1285, opened 2023-11-28
BP: Processor plugin service


lovromazgon:
- Refactor current plugin and plugin service so it's clear it's dealing with connector plugins (rename or move into subpackage)
- Implement the processor plugin interface
- Add a processor plugin service and registry that manages and dispenses processor plugins

PR #1302, opened 2023-12-04
Refactor pkg/plugin and isolate connector plugin

lovromazgon:
### Description

This PR refactors `pkg/plugin` and extracts most of the code into `pkg/plugin/connector`. The only things left in `pkg/plugin` is the plugin service, the utility to work with plugin names and a few error variables. This will allow us to create `pkg/plugin/processor` and put in that package everything related to processor plugins, while we keep the same plugin service.

Related to #1285

### Quick checks:


pkg/plugin/service.go
```diff
@@ -32,8 +33,8 @@ import (
 //     separate process and communicates with it via gRPC. These plugins are
 //     compiled independently of Conduit and can be included at runtime.
 type registry interface {
-	NewDispenser(logger log.CtxLogger, name FullName) (Dispenser, error)
-	List() map[FullName]Specification
+	NewDispenser(logger log.CtxLogger, name FullName) (connector.Dispenser, error)
+	List() map[FullName]connector.Specification
 }
 
 type Service struct {
```
hariso, line 40: IIUC, this plugin service is supposed to return all plugins (connectors + processors)?

lovromazgon: Yes, although I'm not married to this idea, I'm open to splitting it.

hariso: I guess it depends on the usage... Do we know for sure we'll need to return all plugins somewhere, UI maybe? My guess is that it'll be always or most of the time just one or the other.

pkg/plugin/service.go
```diff
@@ -31,56 +32,62 @@ import (
 //   - The standalone registry creates a dispenser which starts the plugin in a
 //     separate process and communicates with it via gRPC. These plugins are
 //     compiled independently of Conduit and can be included at runtime.
-type registry interface {
+type registry[Dispenser, Specification any] interface {
```
hariso, line 35: :muscle:

hariso (approved):

pkg/plugin/service.go
lovromazgon: It will probably always be either one or the other, however the service will look the same for both. I might make the service itself generic (similar to how the registry interface is generic) and instantiate one for processor and one for connector plugins, that could be nicer. In a later PR, I'll merge this as is for now.

PR #1308, opened 2023-12-07
[WIP] Processor plugin

lovromazgon:
### Description

This adds a processor plugin and a builtin registry.

We should move the plugin interface into the SDK to simplify things. Then the plugin doesn't have to import Conduit and Conduit doesn't have to translate objects when sending them to the plugin (at least if it's a builtin plugin). This makes the architecture quite a bit faster and simpler than what we currently have with connector plugins.

Related to #1285

### Quick checks:


pkg/plugin/processor/plugin.go
```diff
@@ -0,0 +1,147 @@
...
+	// Parameters.
+	Description string
+	// Version string. Should be a semver prepended with `v`, e.g. `v1.54.3`.
+	Version string
+	// Author declares the entity that created or maintains this plugin.
+	Author string
+	// SourceParams and DestinationParams are maps of named Parameters that
+	// describe how to configure the plugins Destination or Source.
+	SourceParams      map[string]Parameter
+	DestinationParams map[string]Parameter
```
hariso, line 80: This comes probably from the connector SDK, but I believe we need only params for processors.

pkg/plugin/processor/plugin.go
```diff
@@ -0,0 +1,147 @@
...
+	"context"
+
+	"github.com/conduitio/conduit-commons/opencdc"
+)
+
+// ProcessorPlugin receives records, manipulates them and returns back the processed
+// records.
+//
+//nolint:revive // This interface will be moved into the SDK
+type ProcessorPlugin interface {
```
hariso, line 29: As per the conversation we had earlier: https://github.com/ConduitIO/conduit-processor-sdk/pull/1

hariso: The PR above was merged.

lovromazgon:
Closing in favor of #1322

PR #1331, opened 2024-01-05
Increase length limit for ID and Name for pipeline (128) and connector (256)

lovromazgon:
### Description

Increases the limits imposed on the length of the connector and pipeline `ID` and `Name` fields to 128 characters.

### Quick checks:


pkg/connector/service.go
```diff
@@ -29,6 +29,11 @@ import (
 
 var idRegex = regexp.MustCompile(`^[A-Za-z0-9-_:.]*$`)
 
+const (
+	idLengthLimit   = 128
```
lyuboxa, line 33: @lovromazgon can we double the connector limits on the basis that `enrich` will add the pipeline name/id to the connector name. so probably 256+1?

lyuboxa (approved):

nassor (approved):

pkg/connector/service.go
simonl2002: @lyuboxa is that a platform specific concern?

lyuboxa: @simonl2002 no, part of the pipeline files import

raulb (approved):

lovromazgon (commented):
LGTM @lyuboxa

PR #1343, opened 2024-01-23
fix: OpenCDC unwrap

samirketema:
### Description

Fixes https://github.com/ConduitIO/conduit/issues/1351

### Quick checks:

### DIFF
--- a/pkg/processor/procbuiltin/unwrap.go
+++ b/pkg/processor/procbuiltin/unwrap.go
@@ -259,6 +259,8 @@ func (o *openCDCUnwrapper) Unwrap(rec record.Record) (record.Record, error) {
 		return record.Record{}, err
 	}
 
+	// Position is the only key we preserve from the original record to maintain the reference respect other messages
+	// that will be coming from in the event of chaining pipelines (e.g.: source -> kafka, kafka -> destination)
 	return record.Record{
 		Key:       key,
 		Position:  rec.Position,
--- a/pkg/processor/procbuiltin/unwrap.go
+++ b/pkg/processor/procbuiltin/unwrap.go
@@ -131,13 +131,15 @@ func (o *openCDCUnwrapper) UnwrapOperation(structData record.StructuredData) (re
 		return operation, cerrors.Errorf("record payload after doesn't contain operation")
 	}
 
-	switch o := op.(type) {
+	switch opType := op.(type) {
 	case record.Operation:
-		operation = o
+		operation = opType
 	case string:
-		if err := operation.UnmarshalText([]byte(o)); err != nil {
+		if err := operation.UnmarshalText([]byte(opType)); err != nil {
 			return operation, cerrors.Errorf("couldn't unmarshal record operation")
 		}
+	default:
+		return operation, cerrors.Errorf("expected a record.Operation or a string, got %T", opType)
 	}
 	return operation, nil
 }
@@ -158,6 +160,8 @@ func (o *openCDCUnwrapper) UnwrapMetadata(structData record.StructuredData) (rec
 		for k, v := range m {
 			metadata[k] = fmt.Sprint(v)
 		}
+	default:
+		return metadata, cerrors.Errorf("expected a record.Metadata or a map[string]interface{}, got %T", m)
 	}
 	return metadata, nil
 }
--- a/pkg/processor/procbuiltin/unwrap.go
+++ b/pkg/processor/procbuiltin/unwrap.go
@@ -135,8 +135,7 @@ func (o *openCDCUnwrapper) UnwrapOperation(structData record.StructuredData) (re
 	case record.Operation:
 		operation = o
 	case string:
-		err := operation.UnmarshalText([]byte(o))
-		if err != nil {
+		if err := operation.UnmarshalText([]byte(o)); err != nil {
 			return operation, cerrors.Errorf("couldn't unmarshal record operation")
 		}
 	}
@@ -176,6 +175,8 @@ func (o *openCDCUnwrapper) UnwrapKey(structData record.StructuredData) (record.D
 		key = k
 	case string:
 		key = record.RawData{Raw: []byte(k)}
+	default:
+		return key, cerrors.Errorf("expected a record.Data or a string, got %T", k)
 	}
 
 	return key, nil
@@ -212,6 +213,8 @@ func (o *openCDCUnwrapper) UnwrapPayload(structData record.StructuredData) (reco
 			Before: nil,
 			After:  convertedData,
 		}
+	default:
+		return payload, cerrors.Errorf("expected a record.Change or a map[string]interface{}, got %T", p)
 	}
 	return payload, nil
 }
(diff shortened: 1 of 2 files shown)


pkg/processor/procbuiltin/unwrap.go
```diff
@@ -93,6 +96,78 @@ func (p *unwrapProcessor) Process(_ context.Context, in record.Record) (record.R
...
+			Before: nil,
+			After:  structPayload,
+		},
+		Operation: record.OperationSnapshot,
+	}, nil
+}
+
+// UnwrapKey unwraps key as a kafka connect formatted record, returns the key's payload content, or returns the
+// original key if payload doesn't exist.
+func (k *openCDCUnwrapper) UnwrapKey(key record.Data) record.Data {
```
samirketema, line 139: FYI - I didn't really implement UnwrapKey, this was mostly copied from the kafka connect unwrapper

pkg/processor/procbuiltin/unwrap.go
```diff
@@ -93,6 +96,78 @@ func (p *unwrapProcessor) Process(_ context.Context, in record.Record) (record.R
...
 
+/*
+Example of an OpenCDC record:
+{}
+*/
+
+// openCDCUnwrapper unwraps an OpenCDC record from the payload, by unmarhsalling rec.Payload.After into type Record.
+type openCDCUnwrapper struct{}
+
+func (o *openCDCUnwrapper) Unwrap(rec record.Record) (record.Record, error) {
```
samirketema, line 107: the goal in this method is to serialize the entire message in kafka to `record.Record` struct type

pkg/processor/procbuiltin/unwrap_test.go
```diff
@@ -401,6 +421,115 @@ func TestUnwrap_Process(t *testing.T) {
...
+					After: record.StructuredData{
+						"event_id":     1747353650,
+						"msg":          "string 0e8955b3-7fb5-4dda-8064-e10dc007f00d",
+						"pg_generator": false,
+						"sensor_id":    1250383582,
+						"triggered":    false,
+					},
+				},
+				Key:      record.RawData{Raw: []byte("MTc3NzQ5NDEtNTdhMi00MmZhLWI0MzAtODkxMmE5NDI0YjNh")},
+				Position: []byte("eyJHcm91cElEIjoiNGQ2ZTBhMjktNzAwZi00Yjk4LWEzY2MtZWUyNzZhZTc4MjVjIiwiVG9waWMiOiJzdHJlYW0tNzhscG5jaHg3dHpweXF6LWdlbmVyYXRvciIsIlBhcnRpdGlvbiI6MCwiT2Zmc2V0IjoyMjF9"),
```
samirketema, line 484: should position here be `NzgyNjJmODUtODNmMS00ZGQwLWEyZDAtNTRmNjA1ZjkyYTg0`? from the `record.payload.after.position` in the input?

samirketema: actually no, it shouldn't - I didn't read the method yet - apologies

pkg/processor/procbuiltin/unwrap.go
```diff
@@ -93,6 +96,172 @@ func (p *unwrapProcessor) Process(_ context.Context, in record.Record) (record.R
...
+	var key record.Data
+	ky, ok := structData["key"]
+	if !ok {
+		return key, cerrors.Errorf("record payload after doesn't contain key")
+	}
+
+	switch k := ky.(type) {
+	case record.Data:
+		key = k
+	case string:
```
samirketema, line 177: is it possible for a key to be an `int`?

pkg/processor/procbuiltin/unwrap.go
```diff
@@ -93,6 +96,172 @@ func (p *unwrapProcessor) Process(_ context.Context, in record.Record) (record.R
...
+
+	switch o := op.(type) {
+	case record.Operation:
+		operation = o
+	case string:
+		err := operation.UnmarshalText([]byte(o))
+		if err != nil {
+			return operation, cerrors.Errorf("couldn't unmarshal record operation")
+		}
+	}
```
samirketema, line 142: thoughts on adding a default case to catch unexpected types here?

pkg/processor/procbuiltin/unwrap.go
```diff
@@ -93,6 +96,172 @@ func (p *unwrapProcessor) Process(_ context.Context, in record.Record) (record.R
...
+
+	switch o := op.(type) {
+	case record.Operation:
+		operation = o
+	case string:
+		err := operation.UnmarshalText([]byte(o))
+		if err != nil {
+			return operation, cerrors.Errorf("couldn't unmarshal record operation")
+		}
```
samirketema, line 141: nit:
```suggestion
		if err := operation.UnmarshalText([]byte(o)); err != nil {
			return operation, cerrors.Errorf("couldn't unmarshal record operation")
		}
```

pkg/processor/procbuiltin/unwrap.go
```diff
@@ -93,6 +96,172 @@ func (p *unwrapProcessor) Process(_ context.Context, in record.Record) (record.R
...
+func (o *openCDCUnwrapper) Unwrap(rec record.Record) (record.Record, error) {
+	var structData record.StructuredData
+	data := rec.Payload.After
+	switch d := data.(type) {
+	case record.RawData:
+		// unmarshal raw data to structured
+		err := json.Unmarshal(data.Bytes(), &structData)
+		if err != nil {
+			return record.Record{}, cerrors.Errorf("failed to unmarshal raw data as JSON: %w", unwrapProcType, err)
+		}
```
samirketema, line 229: nit:
```suggestion
		if err := json.Unmarshal(data.Bytes(), &structData); err != nil {
			return record.Record{}, cerrors.Errorf("failed to unmarshal raw data as JSON: %w", unwrapProcType, err)
		}
```

samirketema:
@raulb I left a few nits. Also, do you mind updating the description of the PR since it's ready for review? Thank you so much for carrying this to the finish line 🙇🏾

pkg/processor/procbuiltin/unwrap.go
raulb: @samirketema My assumption is that data will be either formatted as `record.Data` or part of a JSON string. I'll add a case on default to return an error so we can catch those in case they occur.

nassor (approved):

pkg/processor/procbuiltin/unwrap.go
```diff
@@ -93,6 +96,178 @@ func (p *unwrapProcessor) Process(_ context.Context, in record.Record) (record.R
...
+		return record.Record{}, err
+	}
+
+	return record.Record{
+		Key:       key,
+		Position:  rec.Position,
+		Metadata:  metadata,
+		Payload:   payload,
+		Operation: operation,
+	}, nil
```
hariso, line 268: Nitpick: it would be good to document why we decided to unwrap everything except the position.

pkg/processor/procbuiltin/unwrap_test.go
```diff
@@ -401,6 +421,115 @@ func TestUnwrap_Process(t *testing.T) {
...
+			},
+			wantErr: false,
+		},
+		{
+			name: "opencdc with raw data",
```
hariso, line 489: I like the tests!

One tiny nitpick: it would be good to have a couple of negative tests (e.g. Payload.After is nil, Payload.After contains an invalid key/metadata/operation, etc.).

pkg/processor/procbuiltin/unwrap.go
raulb: @hariso would something make sense to you?

`Position is the only key we preserve from the original record to mantain the reference respect other messages that will be coming from`.

hariso: Sounds good... I'd also add the context in which we assume this is happening, and that is "chaining" pipelines.

lyuboxa (approved):

PR #1349, opened 2024-01-25
conditional processor execution

### LINKED ISSUE #1344: Add ability to specify condition when creating a processor

### LINKED ISSUE #1345: Add a field to the pipeline config file (bump minor version)

maha-hajja:
### Description

added a "condition" field for the processor instance, made sure to assign that field by gRPC and by pipeline config files when creating processors.
bumped the pipeline config file minor version from 2.0 to 2.1

Fixes #1344 
Fixes #1345 

### Quick checks:


pkg/orchestrator/orchestrator.go
```diff
@@ -107,7 +107,7 @@ type ConnectorService interface {
 type ProcessorService interface {
 	List(ctx context.Context) map[string]*processor.Instance
 	Get(ctx context.Context, id string) (*processor.Instance, error)
-	Create(ctx context.Context, id string, procType string, parent processor.Parent, cfg processor.Config, p processor.ProvisionType) (*processor.Instance, error)
+	Create(ctx context.Context, id string, procType string, parent processor.Parent, cfg processor.Config, p processor.ProvisionType, condition string) (*processor.Instance, error)
```
hariso, line 110: Nitpick: would be good to document what's expected to be in the string (a Go template)?

Alternatively, we could maybe have `type GoTemplate string`.

proto/api/v1/api.proto
```diff
@@ -130,6 +130,7 @@ message Processor {
 
   string type = 4 [(google.api.field_behavior) = IMMUTABLE];
   Parent parent = 6 [(google.api.field_behavior) = IMMUTABLE];
+  string condition = 9;
```
hariso, line 133: We should add some docs about the condition (i.e. that it's a Go template and whatever else is useful). I believe that adding the docs here would also put them into the Swagger docs.

hariso (commented):
Looking good to me! Left a couple of nitpicks only!

pkg/orchestrator/orchestrator.go
maha-hajja: i prefer adding docs, IMO adding more types will make it more complicated going forward with casting and type changing, I think "condition" is similar to the "type" field, which is also just a string

hariso (approved):

PR #1352, opened 2024-01-27
execute processor condition template

### LINKED ISSUE #1346: Execute and evaluate the condition in the processor

maha-hajja:
### Description

Adds a utility to parse and execute go templates on conduit records, and evaluate them into boolean.
should be used to execute the "condition" field when running processors, only needs to be called in the right place when Better Processors work is done.

Fixes #1346 

### Quick checks:


pkg/processor/template_utils.go
```diff
@@ -0,0 +1,67 @@
...
+	tmpl, err := template.New("").Funcs(sprig.FuncMap()).Parse(t.condition)
+	if err != nil {
+		return err
+	}
+	t.tmpl = tmpl
+	return nil
+}
+
+// execute executes the template for the provided record, and parses the template output into a boolean, returns an error
+// if output is not a boolean, or if template is nil (should be called after TemplateUtils.parse).
```
lovromazgon, line 52: > should be called after TemplateUtils.parse

Why don't we prevent wrong use of the type and just parse the template when creating `TemplateUtils`? In other words, if `NewTeamplateUtils` returns no error you are sure you can use the returned type to evaluate conditions.

pkg/processor/template_utils.go
```diff
@@ -0,0 +1,67 @@
...
+	"fmt"
+	"strconv"
+	"text/template"
+
+	"github.com/Masterminds/sprig/v3"
+	"github.com/conduitio/conduit/pkg/record"
+)
+
+// TemplateUtils provides methods to parse, execute go templates for provided records, and return the boolean value of the output.
+type TemplateUtils struct {
```
lovromazgon, line 28: Last but not least, some naming nitpicks. `TemplateUtils` doesn't really tell me what this type is used for. IIUC we're creating a type that knows how to evaluate a condition attached to a processor and returns `true` or `false`. Well, then how about `processorCondition`? Since we're in package `processor` we could even simply use `condition` (although that's debatable, as the type won't be used outside of this package, so it will never be referred to as `processor.condition`).

And then, what do you do with a condition? You evaluate it. So I would rename `execute` to `Evaluate`. Notice I made it public, as it's meant to be used by code outside of this type.

pkg/processor/template_utils_test.go
```diff
@@ -0,0 +1,64 @@
...
+import (
+	"testing"
+
+	"github.com/conduitio/conduit/pkg/record"
+	"github.com/matryer/is"
+)
+
+func Test_TemplateUtils_ExecuteTrue(t *testing.T) {
+	is := is.New(t)
+	condition := "{{ eq .Metadata.key \"val\" }}"
```
lovromazgon, line 26: Backticks are great in cases when you need to construct a string containing quotes.

```suggestion
	condition := `{{ eq .Metadata.key "val" }}`
```

pkg/processor/template_utils.go
```diff
@@ -0,0 +1,67 @@
...
+		return false, fmt.Errorf("template is nil, make sure to parse it first by calling TemplateUtils.parse()")
+	}
+	var b bytes.Buffer
+	err := t.tmpl.Execute(&b, t.rec)
+	if err != nil {
+		return false, err
+	}
+	output, err := strconv.ParseBool(b.String())
+	if err != nil {
+		return false, fmt.Errorf("error converting the condition go-template output to boolean, %w", err)
```
lovromazgon, line 64: We should use `cerrors.Errorf` to get a stack trace.

pkg/processor/template_utils.go
```diff
@@ -0,0 +1,67 @@
...
+)
+
+// TemplateUtils provides methods to parse, execute go templates for provided records, and return the boolean value of the output.
+type TemplateUtils struct {
+	condition string
+	rec       record.Record
+	tmpl      *template.Template
+}
+
+func NewTemplateUtils(condition string, rec record.Record) *TemplateUtils {
```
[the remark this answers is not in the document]
hariso: Anyone who even got close to Java is creating objects like mad. Please understand that.xD

pkg/processor/template_utils.go
maha-hajja: thanks for all the suggestions!
I went with `processorCondition`, thought it made more sense, especially since it won't be used outside the package (as you mentioned).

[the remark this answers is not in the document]
maha-hajja: the method `NewProcessorCondition` is public, and therefor, `TemplateUtils ` or the new name `ProcessorCondition` should be public too, cuz it's returned from `NewProcessorCondition`.
even golangci-lint doesn't like this: `exported func NewProcessorCondition returns unexported type *processor.processorCondition, which can be annoying to use`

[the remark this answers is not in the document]
maha-hajja: BUT if it's public, then it'll start complaining about the name starting with "processor" 🫠
I can rename to `Condition` or `ConditionEvaluator` or even ignore the golangci-lint for this line, WDYT?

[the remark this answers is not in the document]
lovromazgon: The question now is why we need to have a public constructor? I see this more of an internal thing only relevant for the processor. Any code outside of this package should only care about the condition as a string, while internally we know how to parse and evaluate the condition.

TL;DR: I'd make the constructor private `newProcessorCondition` and leave it at that for now. Once we tie it into the processor code we'll see if that makes sense and if we need to move this code or expose it.

lovromazgon (approved):

Issue #1291, opened 2023-11-28
BP: Backwards compatible processors


lovromazgon:
- Create processors with the same names as the old processors that forward calls to new processors and ensure the same functionality as before
- When using an old processor, Conduit should emit a log with a deprecation warning

lovromazgon:
Won't do.

PR #1354, opened 2024-01-30
Refactor processor service, processor and node

### LINKED ISSUE #1286: BP: Refactor processor service, processor and node
- The existing processor [service](https://github.com/ConduitIO/conduit/blob/main/pkg/processor/service.go#L27), processor [instance](https://github.com/ConduitIO/conduit/blob/9f1d2505d728c521bb151605597b6450f7c7052b/pkg/processor/instance.go#L62) and processor [node](https://github.com/ConduitIO/conduit/blob/9f1d2505d728c521bb151605597b6450f7c7052b/pkg/pipeline/stream/processor.go#L27) need to be refactored to use the processor plugin under the hood
- This should tie together the new functions created as part of the extended processor lifecycle and newly introduced processor plugin interface

hariso:
### Description

Fixes #1286.

Also removes all the existing built-in processors. Due to interface changes, it's a lot of (mostly mechanical) work to fix them. We're going to work on BWC in #1291.

### Quick checks:


pkg/provisioning/test/pipelines1/pipelines.go
```diff
@@ -77,8 +77,8 @@ var P1C2 = &connector.Instance{
 }
 
 var P1P1 = &processor.Instance{
-	ID:   "pipeline1:proc1",
-	Type: "js",
+	ID:     "pipeline1:proc1",
+	Plugin: "js",
```
lovromazgon, line 81: We should probably change the test pipeline yaml files to use the `plugin` field. Maybe create one with `type` that tests backwards compatibility.

pkg/provisioning/config/yaml/v2/model.go
```diff
@@ -26,15 +26,40 @@ const MajorVersion = "2"
...
+			ChangeType: internal.FieldIntroduced,
+			Message:    "field condition was introduced in version 2.1, please update the pipeline config version",
+		},
+		{
+			Field:      "pipelines.*.connectors.*.processors.*.condition",
+			ChangeType: internal.FieldIntroduced,
+			Message:    "field condition was introduced in version 2.1, please update the pipeline config version",
+		},
+	},
+	"2.2": {
```
lovromazgon, line 41: We should also update the constant `LatestVersion` in this file to `2.2`.

pkg/provisioning/config/yaml/v2/model.go
```diff
@@ -64,6 +89,7 @@ type Connector struct {
 type Processor struct {
 	ID        string            `yaml:"id" json:"id"`
 	Type      string            `yaml:"type" json:"type"`
```
lovromazgon, line 91: We should mark this field with `omitempty` so it's ignored when exporting the config.
```suggestion
	Type      string            `yaml:"type,omitempty" json:"type,omitempty"`
```

pkg/provisioning/config/parser.go
```diff
@@ -39,8 +39,10 @@ type Connector struct {
 }
 
 type Processor struct {
```
lovromazgon, line 41: Thinking out loud - should we add the same changes in `yaml/v1`? It seems that for completeness we would need to do that.

lovromazgon (commented):
Posting first half of the review. I'll check the rest later.

pkg/record/record.go
```diff
@@ -186,6 +187,33 @@ func (r Record) Clone() Record {
...
+func toOpenCDCData(data Data) opencdc.Data {
+	switch v := data.(type) {
+	case nil:
+		return nil
+	case RawData:
+		return opencdc.RawData(v.Bytes())
+	case StructuredData:
+		return opencdc.StructuredData(v)
+	default:
+		// this shouldn't happen, we hope
```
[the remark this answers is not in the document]
hariso: I was like: good to have just in case we add another Data implementation in the future. But maybe it's too much as well.

pkg/provisioning/config/parser.go
```diff
@@ -39,8 +39,10 @@ type Connector struct {
...
-	ID        string
+	ID string
+	// Type is deprecated. Plugin should be used.
```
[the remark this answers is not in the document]
hariso: Nice!

pkg/provisioning/config/yaml/v2/model.go
hariso: I was thinking about that too today, but noticed we don't that for some other fields (such as condition, workers, which are optional). Any reason not  to add `omitempty` to those too?

pkg/provisioning/config/parser.go
hariso: I was thinking about that too... I believe that we should be encouraging users to use v2 as much as possible,

pkg/pipeline/stream/processor.go
```diff
@@ -42,40 +55,59 @@ func (n *ProcessorNode) Run(ctx context.Context) error {
...
 	}
-
 	defer cleanup()
+
+	n.logger.Debug(ctx).Msg("opening processor")
+	err = n.Processor.Open(ctx)
+	if err != nil {
+		n.logger.Err(ctx, err).Msg("failed opening processor")
+		return cerrors.Errorf("couldn't open processor: %w", err)
+	}
```
lovromazgon, line 65: If we hit this error the processor will still be running (or at least the processor instance will have the boolean `running` set to `true`). We need to either call `Teardown` here or in `RunnableProcessor.Open` on an error. I think calling it here would be a bit more explicit, this way `RunnableProcessor.Open` doesn't make any assumptions on how it's used.

pkg/pipeline/stream/processor.go
```diff
@@ -42,40 +55,59 @@ func (n *ProcessorNode) Run(ctx context.Context) error {
...
-				}
-			}
-			// error was handled successfully, we recovered
-			continue
+
+		if len(recsIn) != len(recsOut) {
+			return msg.Nack(
+				cerrors.Errorf("processor was given %v record(s), but returned %v", len(recsIn), len(recsOut)),
+				n.ID(),
+			)
```
lovromazgon, line 90: If we have a DLQ configured, this will return `nil` and the node will stop without stopping the pipeline. We need to make sure we return an error so a stop of the pipeline is truggered. Something like this:
```suggestion
			err = cerrors.Errorf("processor was given %v record(s), but returned %v", len(recsIn), len(recsOut))
			if nackErr := msg.Nack(err, n.ID()); nackErr != nil {
				return nackErr
			}
			return err
```

pkg/plugin/processor/registry.go
```diff
@@ -0,0 +1,66 @@
...
+	"github.com/conduitio/conduit/pkg/foundation/cerrors"
+	"github.com/conduitio/conduit/pkg/foundation/log"
+	"github.com/conduitio/conduit/pkg/plugin"
+)
+
+type processorCreator interface {
+	NewProcessor(ctx context.Context, fullName plugin.FullName, id string) (sdk.Processor, error)
+}
+
+type Registry struct {
```
lovromazgon, line 32: There's a bit of a discrepancy now between `plugin.Service` which is the plugin service for connector plugins and this `processor.Registry` which is essentially the same thing for processors.

I'm thinking that we should move that `plugin.Service` into package `pkg/plugin/connector` so it's on the same level as this `Registry` and then either call this `Service` or call that one `Registry`, so it's clear that they are the same thing, one for connector plugins and the other for processor plugins.

pkg/plugin/processor/registry.go
```diff
@@ -0,0 +1,66 @@
...
+	// standalone processors take precedence
+	// over built-in processors with the same name
+	case plugin.PluginTypeStandalone:
+		return r.standaloneReg.NewProcessor(ctx, fullName, id)
+	case plugin.PluginTypeBuiltin:
+		return r.builtinReg.NewProcessor(ctx, fullName, id)
+	case plugin.PluginTypeAny:
+		d, err := r.standaloneReg.NewProcessor(ctx, fullName, id)
+		if err != nil {
+			r.logger.Debug(context.Background()).Err(err).Msg("could not find standalone plugin dispenser, falling back to builtin plugin")
```
lovromazgon, line 59:
```suggestion
			r.logger.Debug(ctx).Err(err).Msg("could not find standalone plugin dispenser, falling back to builtin plugin")
```

pkg/processor/errors.go
```diff
@@ -20,6 +20,8 @@ var (
 	// ErrSkipRecord is passed by a processor when it should Ack and skip a Record.
 	// It must be separate from a plain error so that we continue instead of marking
 	// the Pipeline status as degraded.
-	ErrSkipRecord       = cerrors.New("record skipped")
-	ErrInstanceNotFound = cerrors.New("processor instance not found")
+	ErrSkipRecord              = cerrors.New("record skipped")
+	ErrInstanceNotFound        = cerrors.New("processor instance not found")
+	ErrProcessorNotInitialized = cerrors.New("processor not initialized")
```
lovromazgon, line 25: I don't think we need `ErrSkipRecord` and `ErrProcessorNotInitialized`. Well, technically we still need `ErrSkipRecord` because it's used in `FuncWrapper`, but we should rather move to `ProcessorFunc` introduced in the [SDK](https://github.com/ConduitIO/conduit-processor-sdk/pull/17), then we can remove this error.

pkg/processor/instance.go
```diff
@@ -42,37 +43,50 @@ type (
...
-	// InspectIn starts an inspection session for input records for this processor.
-	InspectIn(ctx context.Context, id string) *inspector.Session
-	// InspectOut starts an inspection session for output records for this processor.
-	InspectOut(ctx context.Context, id string) *inspector.Session
-
-	// Close closes this processor and releases any resources
-	// which may have been used by it.
-	Close()
+type Registry interface {
+	Get(ctx context.Context, pluginName string, id string) (sdk.Processor, error)
```
lovromazgon, line 47: A bit of a nitpick - I think we should move this to `service.go`, since it's used in that file. Same goes for the `go:generate` statement on top.

lovromazgon: Btw, maybe we should call this `PluginRegistry`? 🤔

lovromazgon (approved):
I think we're good to merge this PR 🙂 Kudos @hariso for the good work 💪

lovromazgon:
Oh and don't worry about the buf validation, we deleted the field but reserved the name and field number, so the API is still backwards compatible.

hariso:
> I think we're good to merge this PR 🙂 Kudos @hariso for the good work 💪

Thanks for all the collaboration!:)

> Oh and don't worry about the buf validation, we deleted the field but reserved the name and field number, so the API is still backwards compatible.

Yup, I realized that. 

So FTR, I've merged the PR since this change in proto files is intended.

Issue #1359, opened 2024-02-02
Bug: can't unwrap a record on delete using opencdc


raulb:
### Bug description

When trying to delete a record using the opencdc format, unwrap throws an error:

```
2024-02-02T16:56:03+00:00 ERR pipeline stopped error="node dest-file-output-xkapn00qyn60xog:source-kafka:unwrap-source-kafka stopped with error: error executing processor: failed to write message to DLQ: DLQ nack threshold exceeded (0/1), original error: unwrap: error unwrapping record: unexpected data type string%!(EXTRA <nil>)"
```

### Steps to reproduce

1. Have a pipeline which using a the opencdc processor to unwrap records.
2. Perform a delete operation
3. Notice the error thrown.

### Version

v0.9.0-nightly.20240202

PR #1362, opened 2024-02-05
fix: unwrap openCDC records on create and update

raulb:
### Description

Fixes https://github.com/ConduitIO/conduit/issues/1359

### Quick checks:

### DIFF
--- a/pkg/processor/procbuiltin/unwrap.go
+++ b/pkg/processor/procbuiltin/unwrap.go
@@ -194,6 +194,31 @@ func (o *openCDCUnwrapper) UnwrapKey(structData record.StructuredData) (record.D
 	return key, nil
 }
 
+func (o *openCDCUnwrapper) convertPayloadData(payload map[string]interface{}, key string) (record.Data, error) {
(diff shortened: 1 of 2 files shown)


pkg/processor/procbuiltin/unwrap.go
```diff
@@ -232,6 +232,11 @@ func (o *openCDCUnwrapper) UnwrapPayload(structData record.StructuredData) (reco
 				Before: nil,
 				After:  convertedData,
 			}
+		case nil: // This can take place in operations such as `delete`
```
hariso, line 235: I believe this will solve the problem in most cases. There are a few additional cases we need to think about:
* updates can have a before field
* deletes can have a before field

In other words, if `after` is `nil`, that doesn't necessarily mean it's a delete operation.

As an alternative, maybe we can do something like:
```go
payload = record.Change{
    Before: convertData(p["before"])
    After: convertData(p["after"])
}
```
```go
func convertData(data interface{}) {
    // check if data is nil or a string or map
}
```

hariso (approved):

PR #1363, opened 2024-02-05
refactor: opencdc test examples

raulb:
### Description

Follow up on https://github.com/ConduitIO/conduit/pull/1362 so provided examples represent their operations.

### Quick checks:

### DIFF
--- a/pkg/processor/procbuiltin/unwrap_test.go
+++ b/pkg/processor/procbuiltin/unwrap_test.go
@@ -42,9 +42,36 @@ const DebeziumRecordPayload = `{
 		 "schema": {} 
 		}`
 
-const OpenCDCRecordWithAfterPayload = `{
+const OpenCDCRecordUpdateWithBeforePayload = `{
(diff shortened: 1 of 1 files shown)


pkg/processor/procbuiltin/unwrap_test.go
```diff
@@ -42,7 +42,7 @@ const DebeziumRecordPayload = `{
 		 "schema": {} 
 		}`
 
-const OpenCDCRecordCreatePayload = `{
+const OpenCDCRecordWithAfterPayload = `{
```
hariso, line 45: I like the idea behind this PR! What I'd suggest changing are the names to include the operation/test case as well. E.g. create and update always have an after, update sometimes has a before, delete has neither.

raulb: @hariso updated, let me know what you think https://github.com/ConduitIO/conduit/pull/1363/commits/84e58c05e060fb405010c45a8972805c706a83e1.

hariso (approved):

PR #1367, opened 2024-02-07
BP: refactor runnable processor

lovromazgon:
This is an intermediary PR with changes we did together with @hariso while pairing.

It includes refactorings regarding processors and introduces a `RunnableProcessor` which then allows us to use `processor.Instance` mostly just as a data object.


pkg/processor/service.go
```diff
@@ -83,20 +84,18 @@ func (s *Service) Get(_ context.Context, id string) (*Instance, error) {
 	return ins, nil
 }
 
-func (s *Service) InitInstance(ctx context.Context, i *Instance) error {
-	if i.Processor != nil {
-		return ErrProcessorRunning
+func (s *Service) MakeRunnableProcessor(ctx context.Context, i *Instance) (*RunnableProcessor, error) {
+	if i.running {
```
hariso, line 88: Thinking out loud... Can there race conditions around this? E.g. someone tries to update a processor while it's still being built.

hariso (approved):

pkg/processor/service.go
lovromazgon: Yes, but we have the same issue in all services right now, it's something I'd like to avoid adding inside of services. I'm thinking we should rather have a separate layer, maybe on the level of the orchestrator, that locks all entities that are needed in an operation. For example, creating a processor on a connector means we should also lock the pipeline and connector onto which this processor will be attached, and the processor service has no idea about those objects.

So yeah, it is an issue, but I think it should be solved holistically.

Issue #1290, opened 2023-11-28
BP: Rewrite built-in processors


lovromazgon:
- Implement new list of processors defined in the design document
- Make sure each processor is documented and uses the specification generator to extract the specification from Go docs
- Full list:
  - [x] `field.set` - set a field in the record https://github.com/ConduitIO/conduit/pull/1371
  - [x] `field.subset.exclude` - remove fields from the record if they are on the list https://github.com/ConduitIO/conduit/pull/1371
  - [ ] `field.subset.include` - remove fields from the record if they are not on the list
  - [x] `field.rename` - rename fields https://github.com/ConduitIO/conduit/pull/1371
  - [x] `field.convert` - convert one type into another in a structured payload https://github.com/ConduitIO/conduit/pull/1371
  - [x] `filter` - ack record and remove it from the pipeline https://github.com/ConduitIO/conduit/pull/1371
  - [x] `avro.encode` - encode structured data into Avro data https://github.com/ConduitIO/conduit/pull/1401
  - [x] `avro.decode` - decode Avro data into structured data https://github.com/ConduitIO/conduit/pull/1410
  - [ ] `json.encode` - encode structured data into a JSON string
  - [x] `json.decode` - decode a JSON string into structured data https://github.com/ConduitIO/conduit/pull/1405
  - [x] `unwrap.kafkaconnect` - unwrap a Kafka Connect payload into the current record #1395 
  - [x] `unwrap.debezium` - unwrap a Debezium payload into the current record #1393 
  - [x] `unwrap.opencdc` - unwrap a OpenCDC payload into the current record #1387 
  - [x] `webhook.http` - send a HTTP request for each record #1382 
  - [x] `custom.javascript` - execute JavaScript code for each record https://github.com/ConduitIO/conduit/pull/1374
- Post-merge:
  - [ ] update docs, so that parameter docs don't start with field names: https://github.com/ConduitIO/conduit/pull/1382#pullrequestreview-1914283303
  - [ ] update built-in registry

maha-hajja:
design doc https://github.com/ConduitIO/conduit/blob/main/docs/design-documents/20231117-better-processors.md

PR #1371, opened 2024-02-08
BP: add builtin processor part#1

### LINKED ISSUE #1369: `field.set` - set a field in the record

### LINKED ISSUE #1370: `field.subset.exclude` - remove fields from the record if they are on the list

maha-hajja:
### Description

three builtin processors: https://github.com/ConduitIO/conduit/issues/1290
`field.set` 
`field.subset.exclude`
`field.rename`
`field.convert`
`filter`
docs for processors: https://github.com/ConduitIO/conduit/blob/main/docs/design-documents/20231117-better-processors.md#processors

Related to #1290 

### Quick checks:


pkg/plugin/processor/builtin/values.go
```diff
@@ -0,0 +1,35 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package builtin
+
+import "github.com/conduitio/conduit/pkg/foundation/cerrors"
+
+// references
+
+const (
```
maha-hajja, line 21: any suggestion of where to keep these values? or maybe another file name? lemme know what you think

pkg/plugin/processor/builtin/filter_test.go
```diff
@@ -0,0 +1,33 @@
...
+		},
+		{
+			Metadata: map[string]string{"key2": "val2"},
+			Payload:  opencdc.Change{},
+		},
+	}
+	want := []sdk.ProcessedRecord{sdk.FilterRecord{}, sdk.FilterRecord{}}
+	output := proc.Process(context.Background(), records)
+	is.True(len(output) == 2)
+	is.Equal(output, want)
```
hariso, line 32: With this check, I think it's possible to remove the length check above.

pkg/plugin/processor/builtin/renameField.go
```diff
@@ -0,0 +1,149 @@
...
+		parts := strings.Split(pair, ":")
+		if len(parts) != 2 {
+			return cerrors.Errorf("wrong format for the %q param, should be a comma separated list of keys and values,"+
+				"ex: .Metadata.key:id,.Payload.After.foo:bar", "mapping")
+		}
+		key := strings.TrimSpace(parts[0])
+		if key == OperationReference || key == KeyReference || key == PayloadReference || key ==
+			PayloadBeforeReference || key == PayloadAfterReference || key == PositionReference || key == MetadataReference {
+			return cerrors.Errorf("cannot rename one of the top-level fields %q", key)
+		}
```
hariso, line 78: Alternatively, `slices.Contains(forbiddenField, key)` can be used.

pkg/plugin/processor/builtin/renameField.go
```diff
@@ -0,0 +1,149 @@
...
+func (p *renameField) Process(_ context.Context, records []opencdc.Record) []sdk.ProcessedRecord {
+	out := make([]sdk.ProcessedRecord, 0, len(records))
+	for _, record := range records {
+		for key, val := range p.mapping {
+			err := p.rename(record, key, val)
+			if err != nil {
+				return append(out, sdk.ErrorRecord{Error: err})
+			}
+		}
+		out = append(out, sdk.SingleRecord(record))
```
hariso, line 100: Using assignments (i.e. `out[i]`) should be faster then `append()`.

pkg/plugin/processor/builtin/renameField.go
```diff
@@ -0,0 +1,149 @@
...
+	}
+	return out
+}
+
+func (p *renameField) rename(record opencdc.Record, key, val string) error {
+	resolver1, err := sdk.NewReferenceResolver(key)
```
hariso, line 106: Similarly to another processor, it would be good to have this initialized once in Configure(), and then used here.

pkg/plugin/processor/builtin/renameField.go
```diff
@@ -0,0 +1,149 @@
...
+			if err != nil {
+				return append(out, sdk.ErrorRecord{Error: err})
+			}
+		}
+		out = append(out, sdk.SingleRecord(record))
+	}
+	return out
+}
+
+func (p *renameField) rename(record opencdc.Record, key, val string) error {
```
hariso, line 105: At least to me, it would improve the readability of the method if the parameters were called `oldName` and `newName`, or something along those lines.

pkg/plugin/processor/builtin/renameField.go
```diff
@@ -0,0 +1,149 @@
...
+	err = ref1.Delete()
+	if err != nil {
+		return err
+	}
+	return nil
+}
+
+func (p *renameField) getNameWithPrefix(newName, oldName string) string {
+	// split the oldName by dots
+	parts := strings.Split(oldName, ".")
```
hariso, line 138: What if you have something like `.Metadata["old name"]`?

pkg/plugin/processor/builtin/setField.go
```diff
@@ -0,0 +1,105 @@
...
+)
+
+type setField struct {
+	referenceResolver sdk.ReferenceResolver
+	value             string
+
+	sdk.UnimplementedProcessor
+}
+
+func (p *setField) Specification() (sdk.Specification, error) {
```
hariso, line 32: Should we also register all the rewritten plug-ins? I did so with the JS processor: https://github.com/ConduitIO/conduit/pull/1374/files#diff-6c82ac75684bca1ba4a7e1c33a0ac2058bed4b91712149bd8d1806b07e0904e6. Or maybe there's a reason we're not doing that yet?

pkg/plugin/processor/builtin/setField.go
```diff
@@ -0,0 +1,105 @@
...
+						Type: sdk.ValidationTypeRequired,
+					},
+				},
+			},
+		},
+	}, nil
+}
+
+func (p *setField) Configure(_ context.Context, cfg map[string]string) error {
+	field, ok := cfg["field"]
```
hariso, line 65: Nitpick: it should not be possible to change the value of the position. Conduit itself will complain, but it's also good to check it here.

pkg/plugin/processor/builtin/setField_test.go
```diff
@@ -0,0 +1,132 @@
...
+
+func TestSetField_Process(t *testing.T) {
+	proc := setField{}
+	var err error
+	ctx := context.Background()
+	testCases := []struct {
+		field  string
+		value  string
+		record opencdc.Record
+		want   opencdc.Record
```
hariso, line 35: If `want` is a `sdk.SingleRecord`, then you can simplify the check to be only `is.Equal(want, got)`.

pkg/plugin/processor/builtin/setField_test.go
```diff
@@ -0,0 +1,132 @@
...
+			want: opencdc.Record{
+				Payload: opencdc.Change{
+					Before: nil,
+					After: opencdc.StructuredData{
+						"foo": "new-val",
+					},
+				},
+			},
+		}}
+	for _, tc := range testCases {
```
hariso, line 76: Maybe also have a case for non-existing fields?

pkg/plugin/processor/builtin/values.go
hariso: Maybe split the file into references.go and errors.go? I'm sure we'll have more commonly used errors.

hariso (commented):
Great work! I left a few nitpicks here and there, but the PR is great!

pkg/plugin/processor/builtin/convertField.go
```diff
@@ -0,0 +1,169 @@
...
+	}, nil
+}
+
+func (p *convertField) Configure(_ context.Context, cfg map[string]string) error {
+	field, ok := cfg["field"]
+	if !ok {
+		return cerrors.Errorf("%w (%q)", ErrRequiredParamMissing, "field")
+	}
+	if !strings.HasPrefix(field, ".Payload.After") &&
+		!strings.HasPrefix(field, ".Payload.Before") &&
```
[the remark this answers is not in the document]
maha-hajja: but having the prefix `.Payload` could still result in an error, if the user used some invalid field like `.Payload.Something`, that's why I separated them.

pkg/plugin/processor/builtin/convertField.go
```diff
@@ -0,0 +1,169 @@
...
+
+	return nil
+}
+
+func (p *convertField) Open(context.Context) error {
+	return nil
+}
+
+func (p *convertField) Process(_ context.Context, records []opencdc.Record) []sdk.ProcessedRecord {
+	out := make([]sdk.ProcessedRecord, 0, len(records))
```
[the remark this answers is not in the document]
maha-hajja: from what I understand, the slice length could be `len(records)-N` if an error occurred, is that not the case?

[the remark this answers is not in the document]
hariso: I don't think so, right @lovromazgon? i.e. if we have two input records, input1 and input2, and input2 fails, then we return a slice `sdk.SingleRecord{}, sdk.ErrorRecord{}`.

pkg/plugin/processor/builtin/convertField.go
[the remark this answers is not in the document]
hariso: In that case, you would still be able to write `!strings.HasPrefix(field, ".Payload")`, because that eliminates everything except the key and payload.

And then below, when you write `resolver, err := sdk.NewReferenceResolver(field)`, you'd get an error (no Something field). It's a nitpick anyway, a few characters saved only.

pkg/plugin/processor/builtin/excludeFields.go
```diff
@@ -0,0 +1,100 @@
...
+
+func (p *excludeFields) Open(context.Context) error {
+	return nil
+}
+
+func (p *excludeFields) Process(_ context.Context, records []opencdc.Record) []sdk.ProcessedRecord {
+	out := make([]sdk.ProcessedRecord, 0, len(records))
+	for _, record := range records {
+		for _, field := range p.fields {
+			resolver, err := sdk.NewReferenceResolver(field)
```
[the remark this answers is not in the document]
maha-hajja: I do have this line in the `Configure` method for other processor, but for this one, we don't only have one field, we have a list of fields, so we would need to create a new reference resolver for each one of them.
or did you mean that I should at least create it for the first field? or maybe create a slice of reference resolvers in the `Configure` method? lemme know what you think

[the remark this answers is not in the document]
hariso: A slice of resolvers sounds like a good idea to me!

pkg/plugin/processor/builtin/convertField.go
```diff
@@ -0,0 +1,169 @@
...
+	typ               string
+
+	sdk.UnimplementedProcessor
+}
+
+func (p *convertField) Specification() (sdk.Specification, error) {
+	return sdk.Specification{
+		Name:    "field.convert",
+		Summary: "convert the type of a field",
+		Description: "convert takes the field of one type and converts it into another type (e.g. string to integer). " +
```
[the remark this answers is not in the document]
lovromazgon: Also please start the sentences with upper case and end with a period, so this will look nice in documentation (same in `Summary`) 🙏

pkg/plugin/processor/builtin/convertField.go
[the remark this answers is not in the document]
lovromazgon: Maha is right. Although your example @hariso is also right 😄 But if we turn it around, and say that input1 fails, then we return a slice `[sdk.ErrorRecors{}]`, skipping the second record.

The thing is that if the connector does some parallel processing or something, then it _may_ return records after the error record, but it doesn't _have to_. So if the last record in the slice is an error record, conduit will assume that all the later ones were skipped and not processed, so they will be sent again in the next batch (if we have a DLQ and recover from the error) or just nacked (if there's no DLQ and the pipeline stops).

That's at least the idea, for when we actually implement batching in processors.

pkg/plugin/processor/builtin/convertField.go
```diff
@@ -0,0 +1,169 @@
...
+func (p *convertField) Open(context.Context) error {
+	return nil
+}
+
+func (p *convertField) Process(_ context.Context, records []opencdc.Record) []sdk.ProcessedRecord {
+	out := make([]sdk.ProcessedRecord, 0, len(records))
+	for _, record := range records {
+		ref, err := p.referenceResolver.Resolve(&record)
+		if err != nil {
+			return append(out, sdk.ErrorRecord{Error: err})
```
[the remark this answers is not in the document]
lovromazgon: So given [my comment](https://github.com/ConduitIO/conduit/pull/1371/files#r1492849355) this should be valid, the processor is allowed to stop processing the batch if an error occurs. The so far successfully processed records can be returned normally + the error that happened.

pkg/plugin/processor/builtin/convertField.go
```diff
@@ -0,0 +1,169 @@
...
+			return p.boolToString(v)
+		}
+		return strconv.FormatBool(v)
+	default:
+		return fmt.Sprintf("%v", value)
+	}
+}
+
+func (p *convertField) boolToString(b bool) string {
+	return map[bool]string{true: "1", false: "0"}[b]
```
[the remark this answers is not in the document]
lovromazgon: Or just explicitly write out the `if`? 🙃 Why are we converting a bool to a number though? 🤔 Shouldn't this return `"true"` and `"false"`? Also, there's already functions for this like [`strconv.FormatBool()`](https://pkg.go.dev/strconv#FormatBool).

pkg/plugin/processor/builtin/filter.go
```diff
@@ -0,0 +1,57 @@
...
+}
+
+func (p *Filter) Open(context.Context) error {
+	return nil
+}
+
+func (p *Filter) Process(_ context.Context, records []opencdc.Record) []sdk.ProcessedRecord {
+	out := make([]sdk.ProcessedRecord, len(records))
+	for i := range records {
+		out[i] = sdk.FilterRecord{}
```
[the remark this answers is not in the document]
lovromazgon: `sdk.FilterRecord` is already an empty struct, which [consumes no memory](https://dave.cheney.net/2014/03/25/the-empty-struct), I don't think we can do any better here.

pkg/plugin/processor/builtin/renameField.go
```diff
@@ -0,0 +1,149 @@
...
+	sdk.UnimplementedProcessor
+}
+
+func (p *renameField) Specification() (sdk.Specification, error) {
+	return sdk.Specification{
+		Name:    "field.rename",
+		Summary: "rename a group of fields",
+		Description: "rename a group of fields. It is not allowed to rename top-level fields (.Operation, " +
+			".Position, .Key, .Metadata, .Payload.Before, .Payload.After), processor only runs on structured data.",
+		Version: "v1.0",
```
lovromazgon, line 38: For the version we should probably use the full semantic version `vX.X.X`. As for the actual version to use, I'd maybe go with `v0.1.0`? Conduit itself is currently `v0.X`, it feels appropriate to keep processors also on a pre-release version. WDYT @maha-hajja @hariso?

pkg/plugin/processor/builtin/renameField.go
maha-hajja: this one also could have multiple fields to rename, so I'll go with a slice of resolvers in the `Configure` method

pkg/plugin/processor/builtin/renameField.go
maha-hajja: I had that in mind but totally forgot about it, yep agreed!

pkg/plugin/processor/builtin/renameField.go
maha-hajja: oof good catch! didn't even occur to me, will handle that too

lovromazgon: Good point! We might need to expand the resolver with a method that returns a resolver with a renamed field, it would be trivial to do there.

pkg/plugin/processor/builtin/setField.go
lovromazgon: We can prepare that default processor map in the JS processor PR and then add the others in a separate PR, as to not get merge conflicts.

maha-hajja: hmm I was gonna add them at the end, but no reason not to.
however, since we're gonna need `processor.New` for each processor, does that mean I need to separate each processor to their own directory? cuz we can't redeclare `New` multiple times in the same package.

pkg/plugin/processor/builtin/setField.go
lovromazgon: This could be alternatively done by creating a dummy record, resolving the field and trying to set the field to a test string. Then the fact that the position can't be set would only be coded in one place (the reference).

maha-hajja: sorry I don't think I understand, wdym by "value of the position"?

pkg/plugin/processor/builtin/setField.go
```diff
@@ -0,0 +1,105 @@
...
+}
+
+func (p *setField) Process(_ context.Context, records []opencdc.Record) []sdk.ProcessedRecord {
+	out := make([]sdk.ProcessedRecord, 0, len(records))
+	for _, record := range records {
+		ref, err := p.referenceResolver.Resolve(&record)
+		if err != nil {
+			return append(out, sdk.ErrorRecord{Error: err})
+		}
+		err = ref.Set(p.value)
```
lovromazgon, line 94: From the [design doc](https://github.com/ConduitIO/conduit/blob/main/docs/design-documents/20231117-better-processors.md#fieldset):
> `value` (required) - is a Go template expression which will be evaluated and stored in field.

So we should evaluate it and give it the record as the data it operates on. Then this processor becomes very powerful 💪

pkg/plugin/processor/builtin/renameField.go
hariso: Agreed!

pkg/plugin/processor/builtin/renameField.go
hariso: Thanks for the input Lovro!

pkg/plugin/processor/builtin/convertField.go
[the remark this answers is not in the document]
maha-hajja: if you have a string with the value `true` it won't be converted to a number, unless the param `type` is `int` or `float`.
the whole idea is converting whatever is the oldValue to a string, and then the string to the goal type, just to have fewer cases to handle.
the problem with that was when you want to convert a `boolean` to a number (`float` or `int`), then you're first converting the boolean to a string then to a number, but the string output from `strconv.FormatBool()` is `true` or `false`, not `1` and `0`, that's why I added this method just for the case where we're converting a bool to an `int` or `float`.
I can go with explicitly writing out the `if` statement, maybe that'll be easier to understand, and change the method name to `boolToStringNumber`, lemme know what you think about that

pkg/plugin/processor/builtin/setField.go
maha-hajja: ooh didn't notice that part! yep definitely too powerful :D

pkg/plugin/processor/builtin/convertField.go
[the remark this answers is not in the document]
lovromazgon: > if you have a string with the value `true` it won't be converted to a number

Oh right, my bad 🤦

Anyway, as long as the functionality is correct I'm fine either way 😉

pkg/plugin/processor/builtin/setField.go
hariso: I can't comment much on how Go idiomatic is it to have a package for each of them. Personally, I feel that might be a bit too much since some processors will be really simple. Maybe we can include the processor name in the constructor name? We'd then have `builtin.NewRenameField` for example.

pkg/plugin/processor/builtin/setField.go
maha-hajja: or I can also just add an exclusion validation for `.Position` using paramgen!

pkg/plugin/processor/builtin/excludeFields_test.go
```diff
@@ -0,0 +1,97 @@
...
+
+	"github.com/conduitio/conduit-commons/opencdc"
+	sdk "github.com/conduitio/conduit-processor-sdk"
+	"github.com/matryer/is"
+)
+
+func TestExcludeFields_Process(t *testing.T) {
+	is := is.New(t)
+	proc := excludeFields{
+		fields: []string{".Metadata", ".Payload.After.foo"},
```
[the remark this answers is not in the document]
maha-hajja: what should we expect when excluding `.Payload.After.something`? the current behaviour would be that nothing changes, since the resolver would reference and then delete it

pkg/plugin/processor/builtin/setField.go
maha-hajja: the specific processor name constructor sounds like a good idea, but let's still revisit this after JS processor is merged and we have the processor map ready

pkg/plugin/processor/builtin/setField_test.go
maha-hajja: setting non-existence fields would create the field and set its value, will add a case for that

pkg/plugin/processor/builtin/setField.go
hariso: that's a good idea!

pkg/plugin/processor/builtin/excludeFields_test.go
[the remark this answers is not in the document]
hariso: Well, I'd say there are 2 options there:
1. do nothing, because the goal is not to have the field
2. fail, because someone might have expected the field to be there.

I'm leaning towards #1, because #2 is too strict, and it's likely that a user will simply want to normalize some data. If someone _really_ wants to validate the input data, they can write a processor for that.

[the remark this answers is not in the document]
maha-hajja: I like 1 too, cuz the user might want to remove a field IF it exists, it might exist in some records and not others

pkg/plugin/processor/builtin/convertField_test.go
```diff
@@ -0,0 +1,252 @@
...
+	ctx := context.Background()
+	var err error
+	testCases := []struct {
+		name   string
+		field  string
+		typ    string
+		record opencdc.Record
+		want   sdk.SingleRecord
+	}{
+		{
```
hariso, line 37: Nitpick: maybe have a few negative test cases, e.g. string to int, where strings is not a valid number, or converting a very big float into an int and so on.

pkg/plugin/processor/builtin/renameField.go
```diff
@@ -0,0 +1,125 @@
...
+
+	"github.com/conduitio/conduit-commons/config"
+	"github.com/conduitio/conduit-commons/opencdc"
+	sdk "github.com/conduitio/conduit-processor-sdk"
+	"github.com/conduitio/conduit/pkg/foundation/cerrors"
+	"golang.org/x/exp/slices"
+)
+
+type renameField struct {
+	mapping            map[string]string
```
hariso, line 31: Keys (old field reference) in `mapping` are unused. Maybe we can use a slice here, and match references and new field names with indexes? E.g. referenceResolvers[i] is renamed to `newNames[i]`.

pkg/plugin/processor/builtin/convertField.go
```diff
@@ -0,0 +1,162 @@
...
+		ApplyDefaults(cfg.Parameters())
+
+	err := inputCfg.Validate(cfg.Parameters())
+	if err != nil {
+		return cerrors.Errorf("invalid configuration: %w", err)
+	}
+	err = inputCfg.DecodeInto(&cfg)
+	if err != nil {
+		return cerrors.Errorf("failed decoding configuration: %w", err)
+	}
```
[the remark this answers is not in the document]
lovromazgon: `sdk.ParseConfig` - https://github.com/ConduitIO/conduit-processor-sdk/pull/22

pkg/plugin/processor/builtin/convertField.go
```diff
@@ -0,0 +1,162 @@
...
+	if !strings.HasPrefix(cfg.Field, ".Payload") && !strings.HasPrefix(cfg.Field, ".Key") {
```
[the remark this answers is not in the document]
lovromazgon: You both make good points 😄

pkg/plugin/processor/builtin/convertField.go
```diff
@@ -0,0 +1,156 @@
...
+}
+
+func newConvertField() *convertField {
+	return &convertField{}
+}
+
+type convertFieldConfig struct {
+	// Field is the target field, as it would be addressed in a Go template (e.g. `.Payload.After.foo`).
+	// you can only convert fields that are under .Key and .Payload, and said fields should contain structured data.
+	Field string `json:"field" validate:"required,regex=^.(Payload|Key).*"`
```
lovromazgon, line 43: We need to escape `.`, otherwise we are matching any character at the start (e.g. `aPayload` is currently valid).
```suggestion
	Field string `json:"field" validate:"required,regex=^\.(Payload|Key).*"`
```

pkg/plugin/processor/builtin/convertField.go
```diff
@@ -0,0 +1,156 @@
...
+
+func (p *convertField) Specification() (sdk.Specification, error) {
+	return sdk.Specification{
+		Name:    "field.convert",
+		Summary: "Convert the type of a field.",
+		Description: `Convert takes the field of one type and converts it into another type (e.g. string to integer). 
+The applicable types are string, int, float and bool. Converting can be done between any combination of types. Note that
+booleans will be converted to numeric values 1 (true) and 0 (false). Processor is only applicable to .Key, .Payload.Before
+and .Payload.After prefixes, and only applicable if said fields contain structured data.
+if the record contains raw JSON data, then use the processor "decode.json" to parse it into structured data first.`,
```
lovromazgon, line 56:
```suggestion
If the record contains raw JSON data, then use the processor "decode.json" to parse it into structured data first.`,
```

pkg/plugin/processor/builtin/convertField.go
```diff
@@ -0,0 +1,156 @@
...
+		Version:    "v0.1.0",
+		Author:     "Meroxa, Inc.",
+		Parameters: convertFieldConfig{}.Parameters(),
+	}, nil
+}
+
+func (p *convertField) Configure(ctx context.Context, m map[string]string) error {
+	err := sdk.ParseConfig(ctx, m, &p.config, convertFieldConfig{}.Parameters())
+	if err != nil {
+		return cerrors.Errorf("failed to parse configurations: %w", err)
```
lovromazgon, line 66: Nitpick: I'd maybe rather call it "configuration" (singular) instead of "configurations", I see it as one configuration that contains multiple values / parameters.

lovromazgon (commented):
We're really close with this PR, I'm just missing some examples for the sake of documentation, then we can merge! 👍

pkg/plugin/processor/builtin/convertField.go
maha-hajja: hmm looks like we can't have that in a tag, golangci-lint is complaining, and I tried a test with `sPayload` and `regex=^\.(Payload|Key).*`, the validation still passed, so it's not even considering the existence of the `escape`.
the validation passed but ofc the `Configure()` method failed, because of the resolver.
should we just keep it as is, and we know it will fail when trying to get the reference? or move the validation back to `Configure()` manually?
this looks like it would be a bigger problem for paramgen itself, cuz regexes usually have `escape` characters

hariso: Hmm, it's interesting that the linter complains. What does it say? Dumb question, but did you run paramgen again after updating the regex? The regex looks correct, I'm surprised it didn't work.:S

lovromazgon: Maha's right, because this is a string we need to escape the `\` itself, so this should work:
```suggestion
	Field string `json:"field" validate:"required,regex=^\\.(Payload|Key).*"`
```

pkg/plugin/processor/builtin/convertField.go
```diff
@@ -0,0 +1,156 @@
...
+}
+
+func newConvertField() *convertField {
+	return &convertField{}
+}
+
+type convertFieldConfig struct {
+	// Field is the target field, as it would be addressed in a Go template (e.g. `.Payload.After.foo`).
+	// you can only convert fields that are under .Key and .Payload, and said fields should contain structured data.
+	Field string `json:"field" validate:"required,regex=^\\.(Payload|Key)\\.*"`
```
lovromazgon, line 43: Umm we shouldn't escape the last `.` though, that actually represents "any character" 🙈

lovromazgon (approved):
💪

PR #1374, opened 2024-02-09
BP: Add custom.javascript

hariso:
### Description

TBD

### Quick checks:


.golangci.yml
```diff
@@ -1,6 +1,8 @@
 run:
   timeout: 3m
+  skip-dirs-use-default: false
```
hariso, line 3: The `builtin` directory is in the default list of skipped directories: https://golangci-lint.run/usage/configuration/#run-configuration.

pkg/plugin/processor/builtin/registry.go
```diff
@@ -41,6 +46,10 @@ type blueprint struct {
 
 type ProcessorPluginConstructor func(log.CtxLogger) sdk.Processor
 
+func NewDefaultRegistry(logger log.CtxLogger) *Registry {
+	return NewRegistry(logger, defaultBuiltinProcessors)
+}
```
lovromazgon, line 51: tbh I don't see a real need for this function, on the contrary, I'd rather see a call to `NewRegistry` and know what constructors were provided explicitly. I would rather ...

pkg/plugin/processor/builtin/registry.go
```diff
@@ -23,8 +23,13 @@ import (
 	"github.com/conduitio/conduit/pkg/foundation/cerrors"
 	"github.com/conduitio/conduit/pkg/foundation/log"
 	"github.com/conduitio/conduit/pkg/plugin"
+	"github.com/conduitio/conduit/pkg/plugin/processor/builtin/js"
 )
 
+var defaultBuiltinProcessors = map[string]ProcessorPluginConstructor{
```
lovromazgon, line 29: ... expose this variable!
```suggestion
var DefaultBuiltinProcessors = map[string]ProcessorPluginConstructor{
```

pkg/plugin/connector/builtin/v1/internal/fromplugin/specifier.go
```diff
@@ -74,7 +74,8 @@ func SpecifierParameter(in cpluginv1.SpecifierParameter) connectorPlugin.Paramet
 	// in.Required is converted to a validation of type ValidationTypeRequired
 	// making sure not to duplicate the required validation
 	if in.Required && !requiredExists {
-		//nolint:makezero // we don't know upfront if we need add this validation
+		//nolint:makezero // regarding ' append to slice `validations` with non-zero initialized length'
```
maha-hajja, line 77: super nitpicky, but it just looked weird 
```suggestion
		//nolint:makezero // regarding 'append to slice `validations` with non-zero initialized length'
```

pkg/plugin/processor/builtin/js/processor.go
```diff
@@ -0,0 +1,370 @@
...
+}
+
+func (p *processor) Specification() (sdk.Specification, error) {
+	return sdk.Specification{
+		Name:        "custom.javascript",
+		Summary:     "JavaScript processor",
+		Description: "A JavaScript processor for Conduit",
+		Version:     "v0.1.0",
+		Author:      "Meroxa, Inc.",
+		Parameters: map[string]config.Parameter{
```
maha-hajja, line 81: can you please use paramgen for the params

pkg/plugin/processor/builtin/js/processor.go
```diff
@@ -0,0 +1,370 @@
...
+	var jsRecs []interface{}
+	err := runtime.ExportTo(v, &jsRecs)
+	if err != nil {
+		return p.makeProcessedRecords(
+			sdk.ErrorRecord{Error: cerrors.Errorf("failed exporting JavaScript records to Go values: %w", err)},
+			recordsCount,
+		)
+	}
+
+	out := make([]sdk.ProcessedRecord, len(jsRecs))
```
maha-hajja, line 311: lets make sure this slice is fixed for all the processors, so that the length is not len(records), and that we append to it wach time

pkg/plugin/processor/builtin/js/processor.go
hariso: I was thinking about that... It didn't look much better with it, given that it's only 2 parameters, and that there's custom validation involved. I can give it another try though.

pkg/plugin/processor/builtin/js/processor.go
hariso: Done!

pkg/plugin/processor/builtin/js/processor.go
lovromazgon: I'm with Maha on this one, I think using the same approach for all processors will make it easier to maintain them. Let's say we change something in the parameter specs or add a feature to paramgen down the road, then it's trivial to regenerate the code for all processors.

hariso: For the record, we had a chat about this on Slack, and we will probably provide a method that validates and parses a configuration, to remove some of the boilerplate code. That should make using paramgen even more compelling, even in cases where there are just 1 or 2 configuration parameters.

pkg/plugin/processor/builtin/js/processor.go
```diff
@@ -0,0 +1,389 @@
...
+
+func (p *processor) newRuntime(logger log.CtxLogger) (*goja.Runtime, error) {
+	rt := goja.New()
+	require.NewRegistry().Enable(rt)
+
+	runtimeHelpers := map[string]interface{}{
+		"logger":         &logger,
+		"SingleRecord":   p.newSingleRecord(rt),
+		"FilterRecord":   p.newFilterRecord(rt),
+		"ErrorRecord":    p.newErrorRecord(rt),
```
lovromazgon, line 194: I'm not a fan of exposing this to the javascript processor, it needlessly complicates writing a JS processor. I'd rather keep the functionality as it was before, meaning that the processor gets 1 record at a time (we can loop through the batch and supply the records one by one), and if it returns `nil` it's considered a filtered record, if it returns a record we use it, anything else is considered an error.

lovromazgon: I'm aware that this means we'll call the processor multiple times when processing a batch and lose some performance, but if the user wants performance they wouldn't use the JS processor in the first place, it's already slow. And since we now have WASM as an alternative, I think we don't have to worry about the performance here and rather keep it simple.

hariso: I was considering that too, but I voted for this one because it keeps the API very similar across languages.

hariso: > Do you mean the API compared to WASM processors?

Yup!



> And once we have, that approach will require 4 functions to be implemented instead of 1

Can you elaborate a bit, why 4 functions?






> In any case, these processors will not be interchangeable

That's true, the code won't be interchangeable, but the idea is that the concepts should be. In other words, if you wrote a processor in WASM, ideally, you'll just skim through the JS docs and write a processor in JS.

lovromazgon: Well you have `Open`, `Configure`, `Read` and `Teardown`. Also you'll need to provide specifications, so there's probably 5. And that's where I'd offer the full functionality. Maybe there will be a simple way to write the WASM processor, similar to how the Go SDK provides `NewProcessorFunc`, and that could be the API that is similar to the `js` processor. But I see no reason to provide a complexer API here, for the reasons stated in my first answer.

hariso: Gotcha. Well, I wouldn't say it's that much more complex It's just making the `Process()` function more similar to how it looks like across other languages. I don't mind changing it though, especially given the timeline.:)

hariso: Done!

pkg/plugin/processor/builtin/js_examples_test.go
```diff
@@ -0,0 +1,77 @@
...
+
+	RunExample(p, example{
+		Description: "",
+		Config: map[string]string{
+			"script": `function process(rec) {
+					rec.Metadata["processed"] = "true";
+					let existing = String.fromCharCode.apply(String, rec.Payload.After);
+					rec.Payload.After = RawData("hello, " + existing);
+					return rec;
+				}`,
```
lovromazgon, line 36:
```suggestion
  rec.Metadata["processed"] = "true";
  let existing = String.fromCharCode.apply(String, rec.Payload.After);
  rec.Payload.After = RawData("hello, " + existing);
  return rec;
}`,
```

lovromazgon (approved):
LGTM 👍

PR #1375, opened 2024-02-09
Rename connector and processor plugin service

lovromazgon:
This is an intermediary PR, doing this:

- renames `pkg.plugin.connector.Service` to `PluginService`
- renames `pkg.plugin.processor.Registry` to `PluginService` (now these two have parity)
- renames a bunch of variables where we used `pluginService` to `connPluginService` (mostly in tests)
- renames `pkg.orchestrator.PluginService` to `ConnectorPluginService`
- renames `pkg.provisioning.PluginService` to `ConnectorPluginService`
- probably some more renames that I forgot to mention, anywhere we had `PluginService` it should now be `ConnectorPluginService`, except in the connector packages, where it's obvious that we are referring to the connector plugin service

Besides renaming, there's two more changes:
- introduces `pkg.orchestrator.ProcessorPluginService` (will be needed once we add API endpoints to list processor plugins)
- adds `pkg.plugin.processor.PluginService` to health check


hariso (approved):
:rocket:

PR #1380, opened 2024-02-13
Improve builtin plugin sandbox

lovromazgon:
### Description

In this PR we change the behavior of the builtin plugin sandbox. Plugin functions are now run in a goroutine so we can detach from the call in case it blocks forever (e.g. a plugin bug). We add tests to confirm that this behavior matches the behavior in standalone plugins.

### Quick checks:

### DIFF
--- a/pkg/plugin/builtin/v1/sandbox.go
+++ b/pkg/plugin/builtin/v1/sandbox.go
@@ -16,9 +16,17 @@ package builtinv1
 
 import (
 	"context"
+	"sync"
 
 	"github.com/conduitio/conduit/pkg/foundation/cerrors"
+	"github.com/conduitio/conduit/pkg/foundation/log"
 )
+
+var sandboxChanPool = sync.Pool{
+	New: func() any {
+		return make(chan any)
+	},
+}
 
 // runSandbox takes a function and runs it in a sandboxed mode that catches
 // panics and converts them into an error instead.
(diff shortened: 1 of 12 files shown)


pkg/plugin/builtin/v1/sandbox.go
```diff
@@ -28,27 +36,71 @@ func runSandbox[REQ any, RES any](
...
+	logger log.CtxLogger,
 ) (res RES, err error) {
-	defer func() {
-		if r := recover(); r != nil {
-			panicErr, ok := r.(error)
-			if !ok {
-				panicErr = cerrors.Errorf("panic: %v", r)
-			}
-			err = panicErr
+	c := sandboxChanPool.Get().(chan any)
```
lovromazgon, line 41: The changes in this function are the most important in this PR.

pkg/plugin/standalone/v1/source.go
```diff
@@ -112,10 +112,6 @@ func (s *sourcePluginClient) Ack(_ context.Context, p record.Position) error {
 }
 
 func (s *sourcePluginClient) Stop(ctx context.Context) (record.Position, error) {
-	if s.stream == nil {
-		return nil, plugin.ErrStreamNotOpen
-	}
```
hariso, line 117: Quick question: how does this help with the blocking call?

pkg/plugin/builtin/v1/sandbox.go
```diff
@@ -28,27 +36,71 @@ func runSandbox[REQ any, RES any](
...
+			c <- err
 		}
+	}
+
+	go func() {
+		defer func() {
+			if r := recover(); r != nil {
+				err, ok := r.(error)
+				if !ok {
+					err = cerrors.Errorf("panic: %v", r)
```
hariso, line 64: Out of curiosity, when is this possible? If a function panicked, then recover() should always return the error, or no?

pkg/plugin/standalone/v1/source.go
lovromazgon: It doesn't, it's an unrelated thing I noticed. I'm not sure why I added it in the first place, it makes the plugin adapter just a bit too smart, so I removed it. Essentially this `if` prevents `Stop` from being called if the stream is not open. But there's actually nothing stopping us from forwarding the call, because technically we don't need the stream to be open for `Stop` to be called (it's not special, it's a normal function), so the plugin shouldn't prevent us from doing that.

I just noticed it because the test cases were failing for `Stop` because the stream is not opened in those tests.

pkg/plugin/builtin/v1/sandbox.go
lovromazgon: So `recover()` will return whatever value was passed to `panic()`, which can be anything. Here's an example: https://go.dev/play/p/XC48MpRhbFY

hariso: Thanks for the example!

hariso (approved):

Issue #1289, opened 2023-11-28
BP: Specification generator


lovromazgon:
- Implement a tool like [paramgen](https://github.com/ConduitIO/conduit-connector-sdk/tree/main/cmd/paramgen) that reads the Go documentation of a processor struct (or config? or both?) and generates a specification struct that can be used in the processor without manually writing it
- Part of this issue is figuring out how the Go documentation should be structured
- Warn the user if a part of the specification is missing, but still generate the code (a "strict" flag would be nice to enforce all required parts of the specification)

PR #1382, opened 2024-02-14
BP: Add webhook.http

hariso:
### Description

Part of #1290.

Relevant documentation: https://github.com/ConduitIO/conduit/blob/main/docs/design-documents/20231117-better-processors.md#webhookhttp

### Quick checks:


pkg/plugin/processor/builtin/webhook_http.go
```diff
@@ -0,0 +1,309 @@
...
+		Description: `
+A processor that sends an HTTP request to the specified URL with the specified HTTP method
+(default is POST) with a content-type header as the specified value (default is application/json).
+
+The whole record as json will be used as the request body and the raw response body will be set under 
+Record.Payload.After. If the response code is (204 No Content) then the record will be filtered out.
+`,
+		Version:    "v0.1.0",
+		Author:     "Meroxa, Inc.",
+		Parameters: nil,
```
hariso, line 100: Waiting for #1289 to finish this.

pkg/plugin/processor/builtin/webhook_http_test.go
```diff
@@ -0,0 +1,483 @@
...
+		// {
+		//	name: "invalid backoff retry config is ignored",
+		//	config: map[string]string{
+		//		"request.url":                 "http://example.com",
+		//		"backoffRetry.min":    "not-a-duration",
+		//		"backoffRetry.max":    "not-a-duration",
+		//		"backoffRetry.factor": "not-a-number",
+		//	},
+		//	wantErr: "something",
+		// },
```
hariso, line 139: I copied this over from the legacy httprequest processor, so that we can discuss it here. I wasn't sure what's the reason for ignoring errors on optional parameters. IMO, if a value is provided, then it must be valid, because there was a clear intention to use it. If we silently set it to a default value, it may lead to unexpected behavior.

pkg/plugin/processor/builtin/webhook_http.go
lovromazgon: You got it, the code is in the feature branch. Also you can and should now use [paramgen](https://github.com/ConduitIO/conduit-commons/tree/main/paramgen) to make your life simpler 🙂

pkg/plugin/processor/builtin/webhook_http.go
```diff
@@ -0,0 +1,365 @@
...
+
+type webhookHTTP struct {
+	sdk.UnimplementedProcessor
+
+	logger log.CtxLogger
+	config webhookHTTPConfig
+}
+
+func NewWebhookHTTP(l log.CtxLogger) sdk.Processor {
+	return &webhookHTTP{logger: l.WithComponent("builtin.webhookHTTP")}
```
lovromazgon, line 94: Hmm, not sure exactly why we provided the logger here, since the logger will also be provided through the context. Generally the idea is to use use `sdk.Logger(ctx)` to fetch the logger.

No need to change this now, it's a small thing that can be fixed later on, but we need to make sure we decide on a way to handle logging here.

pkg/plugin/processor/builtin/webhook_http_test.go
hariso: We discussed this on a standup, all provided options (required or non-required) need to be validated.

pkg/plugin/processor/builtin/webhook_http.go
hariso: Agreed! Yeah, I wanted to make it compatible with the ProcessorPluginConstructor.

pkg/plugin/processor/builtin/webhook/http.go
```diff
@@ -0,0 +1,306 @@
...
+	p.responseBodyRef = &responseBodyRef
+
+	// This field is optional and, if not set, response status won't be saved.
+	if cfg.ResponseStatusRef != "" {
+		responseStatusRef, err := sdk.NewReferenceResolver(cfg.ResponseStatusRef)
+		if err != nil {
+			return cerrors.Errorf("failed parsing response.status %v: %w", p.config.ResponseStatusRef, err)
+		}
+		p.responseStatusRef = &responseStatusRef
+	}
```
hariso, line 139: This is a bit verbose. Unfortunately, adding a decoding hook for the reference resolver didn't help. The reason is that sdk.ReferenceResolver is not acceptable as a field type by `paramgen`.

pkg/plugin/processor/builtin/webhook/http.go
```diff
@@ -0,0 +1,306 @@
...
+	"github.com/conduitio/conduit/pkg/foundation/log"
+	"github.com/goccy/go-json"
+	"github.com/jpillora/backoff"
+)
+
+type httpConfig struct {
+	// URL used in the HTTP request.
+	URL string `json:"request.url" validate:"required"`
+	// HTTP request method to be used.
+	Method string `json:"request.method" default:"POST"`
```
maha-hajja, line 40: we can add an "inclusion" validation for this param, since it had limited options

pkg/plugin/processor/builtin/webhook/http.go
```diff
@@ -0,0 +1,306 @@
...
+	// Value of the Content-Type header.
+	ContentType string `json:"request.contentType" default:"application/json"`
+
+	// Maximum number of retries for an individual record when backing off following an error.
+	BackoffRetryCount float64 `json:"backoffRetry.count" default:"0"`
```
maha-hajja, line 45: we can add a validation (`gt` -1), for the `BackoffRetryFactor` too

pkg/plugin/processor/builtin/webhook/http.go
```diff
@@ -0,0 +1,306 @@
...
+	"github.com/goccy/go-json"
+	"github.com/jpillora/backoff"
+)
+
+type httpConfig struct {
+	// URL used in the HTTP request.
+	URL string `json:"request.url" validate:"required"`
+	// HTTP request method to be used.
+	Method string `json:"request.method" default:"POST"`
+	// Value of the Content-Type header.
```
maha-hajja, line 41: I believe go docs require us to start the comment with the variable name itself.
```suggestion
	// ContentType Value of the Content-Type header.
```
plz make sure all the others are updated too

pkg/plugin/processor/builtin/webhook/http.go
```diff
@@ -0,0 +1,306 @@
...
+	if cfg.ResponseStatusRef != "" {
+		responseStatusRef, err := sdk.NewReferenceResolver(cfg.ResponseStatusRef)
+		if err != nil {
+			return cerrors.Errorf("failed parsing response.status %v: %w", p.config.ResponseStatusRef, err)
+		}
+		p.responseStatusRef = &responseStatusRef
+	}
+
+	// preflight check
+	_, err = http.NewRequest(cfg.Method, cfg.URL, bytes.NewReader([]byte{}))
```
maha-hajja, line 142: maybe this check is better to have under `Open()`? IIUC, `Configure()` should only parse and validate configurations, it shouldn't attempt to connect to anything.
so basically the URL can be inactive when the processor is created, but it should be available when the processor starts running, so after it calls `Open()`

pkg/plugin/processor/builtin/webhook/http.go
```diff
@@ -0,0 +1,306 @@
...
+}
+
+func (p *httpProcessor) Open(context.Context) error {
+	return nil
+}
+
+func (p *httpProcessor) Process(ctx context.Context, records []opencdc.Record) []sdk.ProcessedRecord {
+	out := make([]sdk.ProcessedRecord, len(records))
+	for i, rec := range records {
+		out[i] = p.processRecordWithBackOff(ctx, rec)
```
maha-hajja, line 158: same comment as the PRs before 👀 let's go with appending records instead

hariso: Done!

pkg/plugin/processor/builtin/webhook/http.go
hariso: Done!

pkg/plugin/processor/builtin/webhook/http.go
hariso: Done!

pkg/plugin/processor/builtin/webhook/http.go
```diff
@@ -0,0 +1,307 @@
...
+type httpConfig struct {
+	// URL used in the HTTP request.
+	URL string `json:"request.url" validate:"required"`
+	// Method is the HTTP request method to be used.
+	Method string `json:"request.method" default:"POST"`
+	// ContentType is the value of the Content-Type header.
+	ContentType string `json:"request.contentType" default:"application/json"`
+
+	// BackoffRetryCount is the maximum number of retries for an individual record
+	// when backing off following an error.
```
lovromazgon, line 44: Given that these comments will now appear in the documentation 1:1 I'm wondering if we should ignore the Go convention of starting with the field name and rather just write the documentation, like:
```suggestion
	// The maximum number of retries for an individual record
	// when backing off following an error.
```

lovromazgon (approved):
LGTM, one comment about docs. We also need to merge the feature branch and regenerate `processors.json`.

pkg/plugin/processor/builtin/webhook/http.go
hariso: That makes sense (even though I think it doesn't look too bad even with the field name). I've added a todo for us in https://github.com/ConduitIO/conduit/issues/1290, so we can go through other processors as well.

PR #1384, opened 2024-02-15
Run processor examples, gather specifications

### LINKED ISSUE #1289: BP: Specification generator

lovromazgon:
### Description

This PR adds the utility test function `RunExample` for builtin processors. The idea is that we can write runnable examples for builtin processors, which will be tested for their correctness and gathered together with the processors' specifications in a JSON file called `processors.json`. We can use that JSON file in our documentation to display documentation about our processors.

The function will compare the record before and after the processing and output the diff in the [unified format](https://www.gnu.org/software/diffutils/manual/html_node/Unified-Format.html).

Adding an example looks like this:

```go
// Simulates a processor that will be tested.
func newTestProcessor() sdk.Processor {
	return sdk.NewProcessorFunc(sdk.Specification{
		Name:        "test",
		Version:     "v0.1.0",
		Description: "Test processor",
	}, func(ctx context.Context, rec opencdc.Record) (opencdc.Record, error) {
		rec.Operation = opencdc.OperationUpdate
		return rec, nil
	})
}

func ExampleTestProcessor_1() {
	p := newTestProcessor()

	RunExample(p, example{
		Description: `When the processor receives a record with operation "create", ` +
			`it will transform the operation into "update".`,
		Config: map[string]string{},
		Have:   opencdc.Record{Operation: opencdc.OperationCreate},
		Want:   sdk.SingleRecord{Operation: opencdc.OperationUpdate},
	})

	// Output:
	// processor transformed record:
	// --- before
	// +++ after
	// @@ -1,10 +1,10 @@
	//  {
	//    "position": null,
	// -  "operation": "create",
	// +  "operation": "update",
	//    "metadata": null,
	//    "key": null,
	//    "payload": {
	//      "before": null,
	//      "after": null
	//    }
	//  }
}
```

Closes #1289 

### Quick checks:


pkg/plugin/processor/builtin/internal/diff/README.md
```diff
@@ -0,0 +1,16 @@
+# Diff
```
lovromazgon, line 1: This package doesn't really need to be reviewed, see readme.

hariso (approved):

PR #1387, opened 2024-02-16
BP: Add unwrap.opencdc

hariso:
### Description

Part of #1290.

### Quick checks:


pkg/plugin/processor/builtin/unwrap_opencdc.go
```diff
@@ -0,0 +1,290 @@
...
+func newUnwrapOpenCDC(logger log.CtxLogger) sdk.Processor {
+	return &unwrapOpenCDC{logger: logger}
+}
+
+func (u *unwrapOpenCDC) Specification() (sdk.Specification, error) {
+	return sdk.Specification{
+		Name:    "unwrap.opencdc",
+		Summary: "A processor that unwraps the OpenCDC record saved in one of record's fields.",
+		Description: `
+The unwrap.cdc processors unwraps the OpenCDC record saved in one of record's fields.
```
maha-hajja, line 47:
```suggestion
The unwrap.opencdc processors unwraps the OpenCDC record saved in one of the record's fields.
```

pkg/plugin/processor/builtin/unwrap_opencdc.go
```diff
@@ -0,0 +1,290 @@
...
+	}, nil
+}
+
+func (u *unwrapOpenCDC) Configure(_ context.Context, m map[string]string) error {
+	field, ok := m["field"]
+	if !ok {
+		return cerrors.New("missing required parameter 'field'")
+	}
+
+	if !strings.HasPrefix(field, ".Payload") {
```
maha-hajja, line 74: maybe also the .Key can be an openCDC record?

pkg/plugin/processor/builtin/unwrap_opencdc.go
```diff
@@ -0,0 +1,290 @@
...
+}
+
+func (u *unwrapOpenCDC) Open(context.Context) error {
+	return nil
+}
+
+func (u *unwrapOpenCDC) Process(_ context.Context, records []opencdc.Record) []sdk.ProcessedRecord {
+	out := make([]sdk.ProcessedRecord, len(records))
+	for i, rec := range records {
+		out[i] = u.processRecord(rec)
```
maha-hajja, line 94: I think we talked about this slice in my processor's PR, should be better to initialize it and append records.
we should also make sure to return the slice if an error is returned for any record by `u.processRecord`

pkg/plugin/processor/builtin/unwrap_opencdc.go
```diff
@@ -0,0 +1,290 @@
...
+	}
+	return operation, nil
+}
+
+// unmarshalMetadata extracts metadata from a structuredData record.
+func (u *unwrapOpenCDC) unmarshalMetadata(structData opencdc.StructuredData) (opencdc.Metadata, error) {
+	var metadata opencdc.Metadata
+	meta, ok := structData["metadata"]
+	if !ok {
+		return metadata, cerrors.Errorf("record payload after doesn't contain metadata")
```
maha-hajja, line 191: I guess this error msg is a bit confusing, and it only mentions payload.after, but it can be something else

pkg/plugin/processor/builtin/unwrap_opencdc.go
```diff
@@ -0,0 +1,290 @@
...
+	}
+	return metadata, nil
+}
+
+// unmarshalKey extracts key from a structuredData record.
+func (u *unwrapOpenCDC) unmarshalKey(structData opencdc.StructuredData) (opencdc.Data, error) {
+	var key opencdc.Data
+	ky, ok := structData["key"]
+	if !ok {
+		return key, cerrors.Errorf("record payload after doesn't contain key")
```
maha-hajja, line 213: same as before, with payload.after

pkg/plugin/processor/builtin/unwrap_opencdc.go
```diff
@@ -0,0 +1,290 @@
...
+		return nil, nil
+	}
+}
+
+// unmarshalPayload extracts payload from a structuredData record.
+func (u *unwrapOpenCDC) unmarshalPayload(structData opencdc.StructuredData) (opencdc.Change, error) {
+	var payload opencdc.Change
+	pl, ok := structData["payload"]
+	if !ok {
+		return payload, cerrors.Errorf("record payload doesn't contain payload")
```
maha-hajja, line 265: this error msg might make sense for us, but the user will definitely be confused 👀

pkg/plugin/processor/builtin/unwrap_opencdc.go
hariso: It's like in the docs, but I actually wasn't sure how it can come to that at all.

Note to myself: apply your suggestion from another PR and do the prefix validation through paramgen.

pkg/plugin/processor/builtin/unwrap_opencdc.go
hariso: Agreed!

pkg/plugin/processor/builtin/unwrap_opencdc.go
hariso: True!

pkg/plugin/processor/builtin/unwrap_opencdc.go
hariso: I've changed this to just "no metadata", since the calling method already says "failed unmarshalling metadata". Let me know how it looks like now!

pkg/plugin/processor/builtin/unwrap_opencdc.go
```diff
@@ -0,0 +1,294 @@
...
+		Description: `
+The unwrap.opencdc processors unwraps the OpenCDC record saved in one of the record's fields.
+This is useful in situations where a record goes through intermediate systems before 
+being written to a final destination. In these cases, the original OpenCDC record is
+part of the payload read from the intermediate system and needs to be unwrapped before
+being written to the destination.
+'`,
+		Version: "v0.1.0",
+		Author:  "Meroxa, Inc.",
+		Parameters: config.Parameters{
```
maha-hajja, line 55: paramgen plzzz 🙌

pkg/plugin/processor/builtin/unwrap_opencdc.go
```diff
@@ -0,0 +1,294 @@
...
+	switch v := ref.Get().(type) {
+	case opencdc.RawData:
+		// unmarshal raw data to structured
+		if err := json.Unmarshal(v.Bytes(), &data); err != nil {
+			return sdk.ErrorRecord{Error: cerrors.Errorf("failed to unmarshal raw data as JSON: %w", err)}
+		}
+	case opencdc.StructuredData:
+		data = v
+	case nil:
+		return sdk.ErrorRecord{Error: cerrors.New("field to unmarshal is nil")}
```
maha-hajja, line 120:
```suggestion
		return sdk.ErrorRecord{Error: cerrors.New("field to unmarshal, nil value")}
```

pkg/plugin/processor/builtin/unwrap_opencdc_test.go
```diff
@@ -0,0 +1,577 @@
+// Copyright © 2023 Meroxa, Inc.
```
maha-hajja, line 1: 2024 is here 💁🏽

pkg/plugin/processor/builtin/unwrap_opencdc.go
hariso: I could be missing something, but how exactly does this improve the messaging?

lovromazgon: I don't get it either 🤔

maha-hajja: lol nvmnd in my mind it looked better

pkg/plugin/processor/builtin/unwrap/unwrap_opencdc.go
```diff
@@ -0,0 +1,293 @@
...
+func NewOpenCDCProcessor(logger log.CtxLogger) sdk.Processor {
+	return &openCDCProcessor{logger: logger}
+}
+
+func (u *openCDCProcessor) Specification() (sdk.Specification, error) {
+	return sdk.Specification{
+		Name:    "unwrap.opencdc",
+		Summary: "A processor that unwraps the OpenCDC record saved in one of record's fields.",
+		Description: `
+The unwrap.opencdc processors unwraps the OpenCDC record saved in one of the record's fields.
```
lovromazgon, line 54: We shouldn't repeat the summary, it will look strange in docs.

pkg/plugin/processor/builtin/unwrap/unwrap_opencdc.go
```diff
@@ -0,0 +1,293 @@
...
+}
+
+func (u *openCDCProcessor) Open(context.Context) error {
+	return nil
+}
+
+func (u *openCDCProcessor) Process(_ context.Context, records []opencdc.Record) []sdk.ProcessedRecord {
+	out := make([]sdk.ProcessedRecord, 0, len(records))
+	for _, rec := range records {
+		proc := u.processRecord(rec)
```
lovromazgon, line 90: You know what I'll suggest here, right? 😄 
```suggestion
		proc, err := u.processRecord(rec)
```

pkg/plugin/processor/builtin/unwrap/unwrap_opencdc.go
```diff
@@ -0,0 +1,293 @@
...
+	}
+
+	opencdcRec, err := u.unmarshalRecord(data)
+	if err != nil {
+		return sdk.ErrorRecord{Error: cerrors.Errorf("failed unmarshalling record: %w", err)}
+	}
+	opencdcRec.Position = rec.Position
+
+	// Position is the only key we preserve from the original record to maintain the reference respect other messages
+	// that will be coming from in the event of chaining pipelines (e.g.: source -> kafka, kafka -> destination)
```
lovromazgon, line 128: Nitpick - I'd expect these comments above the line where we update the position.

pkg/plugin/processor/builtin/unwrap/unwrap_opencdc.go
```diff
@@ -0,0 +1,293 @@
...
+		return metadata, cerrors.New("no metadata")
+	}
+
+	switch m := meta.(type) {
+	case opencdc.Metadata:
+		metadata = m
+	case map[string]interface{}:
+		metadata = make(opencdc.Metadata, len(m))
+		for k, v := range m {
+			metadata[k] = fmt.Sprint(v)
```
lovromazgon, line 200: Nitpick - iirc `fmt.Sprintf` isn't that performant, and it's likely that `v` is already a string. So I'd maybe first try to cast `v` to `string` and fall back to `fmt.Sprintf` if that fails.

pkg/plugin/processor/builtin/unwrap/unwrap_opencdc.go
```diff
@@ -0,0 +1,293 @@
...
+	}
+	switch k := ky.(type) {
+	case opencdc.StructuredData:
+		key = k
+	case map[string]interface{}:
+		convertedData := make(opencdc.StructuredData, len(k))
+		for kk, v := range k {
+			convertedData[kk] = v
+		}
+		key = convertedData
```
lovromazgon, line 224: Since `opencdc.StructuredData` is also a map, we could save an allocation and just do `key = (opencdc.StructuredData)(k)`. Same down in line 246.

pkg/plugin/processor/builtin/unwrap/unwrap_opencdc.go
```diff
@@ -0,0 +1,293 @@
...
+		}
+	default:
+		return nil, cerrors.Errorf("expected a opencdc.Metadata or a map[string]interface{}, got %T", m)
+	}
+
+	return metadata, nil
+}
+
+// unmarshalKey extracts key from a structuredData record.
+func (u *openCDCProcessor) unmarshalKey(structData opencdc.StructuredData) (opencdc.Data, error) {
```
lovromazgon, line 210: Actually, why don't we just reuse `convertPayloadData`? I see the only difference is that the key is apparently required. Why though? It's not really a required field.

pkg/plugin/processor/builtin/unwrap_opencdc.go
lovromazgon: Who knows what kinds of records we'll get, for all we know the key can contain anything, raw data, structured data, a wrapped record, base64 encoded photos from my vacation 🙃

pkg/plugin/processor/builtin/unwrap/unwrap_opencdc.go
hariso: Ah, I missed changing this processor.

pkg/plugin/processor/builtin/unwrap_opencdc.go
hariso: Metadata too...

pkg/plugin/processor/builtin/unwrap/opencdc_test.go
```diff
@@ -596,16 +590,7 @@ func TestUnwrapOpenCDC_Process(t *testing.T) {
...
-			if gotErr {
-				wantErr, ok := tt.want.(sdk.ErrorRecord)
-				is.True(ok)
-				is.Equal(wantErr.Error.Error(), errRec.Error.Error())
-
-				return
-			}
-
-			diff := cmp.Diff(tt.want, got[0], cmpopts.IgnoreUnexported(sdk.SingleRecord{}))
+			diff := cmp.Diff(tt.want, got[0], cmpProcessedRecordOpts...)
```
lovromazgon, line 593: This could be simplified a bit:
```suggestion
			is.Equal("", cmp.Diff(tt.want, got[0], cmpProcessedRecordOpts...))
```

lovromazgon (approved):
💪

pkg/plugin/processor/builtin/unwrap/opencdc_test.go
hariso: Nice one!

PR #1393, opened 2024-02-23
BP: Add unwrap.debezium

hariso:
### Description

Part of #1290.

Reference from the design doc: https://github.com/ConduitIO/conduit/blob/main/docs/design-documents/20231117-better-processors.md#unwrapdebezium

Notes:
1. ~~Breaking change: Compared to the legacy `unwrap` processor for Debezium records, this one has a breaking change, since it does NOT run on raw data. `decode.json` should be used if needed.~~
2. The implementation doesn't run on `.Key` as it looks like we don't have a use case for that.
3. `field` is not required. The implementation assigns the default value `.Payload.After`, which should be a good default in most cases.

### Quick checks:


pkg/plugin/processor/builtin/util_test.go
```diff
@@ -0,0 +1,53 @@
...
+
+func AreEqual(t *testing.T, want, got sdk.ProcessedRecord) {
+	is := is.New(t)
+
+	if wantErr, ok := want.(sdk.ErrorRecord); ok {
+		gotErr, gotOk := got.(sdk.ErrorRecord)
+		if gotOk {
+			is.Equal(wantErr.Error.Error(), gotErr.Error.Error())
+		}
+	}
```
lovromazgon, line 34: Do we need this part? seems like the comparer below is taking care of comparing the errors. And if that's so, we might just get away with having a global variable with the options, like:

```go
var cmpProcessedRecordOpts = []cmp.Option{ ... }
```

And we would use it in tests directly:

```go
is.Equal("", cmp.Diff(want, got, cmpProcessedRecordOpts...))
```

My issue with these custom assert functions is that I don't know what they are comparing until I read them. If I see `is.Equal` I know we're asserting the equality, and then I see `cmp.Diff` I know we're diffing two records and we expect no diff. My 2 cents 🙃

pkg/plugin/processor/builtin/unwrap_debezium.go
```diff
@@ -0,0 +1,347 @@
...
+		return opencdc.OperationDelete, nil
+	case debeziumOpRead:
+		return opencdc.OperationSnapshot, nil
+	case debeziumOpUnset:
+		return opencdc.OperationUpdate, nil
+	}
+	return 0, cerrors.Errorf("%q is an invalid operation", op)
+}
+
+func (u *unwrapDebezium) unwrapKey(key opencdc.Data) opencdc.Data {
```
lovromazgon, line 319: I wonder if these `unwrap` processors can reuse some of the same utilities, iirc this is the same as in the kafka connect processor. Maybe a refactoring down the road, once we have all processors merged in the same branch.

pkg/plugin/processor/builtin/unwrap_debezium.go
```diff
@@ -0,0 +1,347 @@
...
+}
+
+func (u *unwrapDebezium) Specification() (sdk.Specification, error) {
+	return sdk.Specification{
+		Name:    "unwrap.debezium",
+		Summary: "Unwraps a Debezium record from the input OpenCDC record.",
+		Description: `This processor unwraps a Debezium record from the input OpenCDC record.
+
+This is useful in cases where Conduit acts as an intermediary between a Debezium source and a Debezium destination. 
+In such cases, the Debezium record is set as the OpenCDC record's payload, and needs to be unwrapped for further usage.`,
```
lovromazgon, line 72: We should probably mention that the resulted unwrapped record will completely replace the current record. Only the position will be copied over to the new record. We should also explain how metadata is handled (the debezium metadata and record metadata is merged, on conflicts debezium metadata takes precedence).

pkg/plugin/processor/builtin/unwrap_debezium.go
```diff
@@ -0,0 +1,347 @@
...
+	debeziumFieldAfter     = "after"
+	debeziumFieldSource    = "source"
+	debeziumFieldOp        = "op"
+	debeziumFieldTimestamp = "ts_ms"
+)
+
+type unwrapDebeziumConfig struct {
+	// Field is a reference to the field which contains the Debezium record.
+	//
+	// For more information about record references, see: https://github.com/ConduitIO/conduit-processor-sdk/blob/cbdc5dcb5d3109f8f13b88624c9e360076b0bcdb/util.go#L66.
```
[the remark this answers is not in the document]
hariso: Yup, yup, definitely. I can leave a todo here.

pkg/plugin/processor/builtin/unwrap_debezium.go
hariso: Yup, yup, definitely, we can do a bit of refactoring down the road. I've been intentionally creating individual PRs to keep them small.

pkg/plugin/processor/builtin/util_test.go
hariso: Agreed, it's not that pretty. If the name is the issue, I'll try suggesting a better one (AreRecordsEqual or smth along those lines).

A few times I needed a simple way to compare processed records. Additionally, I wanted `cmp.Diff` because it shows the difference very nicely.

The main issue in tests with `sdk.ErrorRecord` is that the errors themselves cannot be easily compared (well, unless you use `cerrors.Is/As`, but that's not always nice). That's why I need a separate check for ErrorRecords.

pkg/plugin/processor/builtin/unwrap_debezium.go
```diff
@@ -0,0 +1,347 @@
...
+}
+
+func (u *unwrapDebezium) Process(_ context.Context, records []opencdc.Record) []sdk.ProcessedRecord {
+	out := make([]sdk.ProcessedRecord, 0, len(records))
+	for _, rec := range records {
+		proc := u.processRecord(rec)
+		out = append(out, proc)
+		if _, ok := proc.(sdk.ErrorRecord); ok {
+			return out
+		}
```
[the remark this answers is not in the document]
hariso: By idiomatic you mean that there should be a result and an error? If so, don't we consider an ErrorRecord (and also the FilterRecord) results too?

pkg/plugin/processor/builtin/util_test.go
lovromazgon: The name is not the issue, also `cmp.Diff` is fine (I like it especially for comparing complex objects like records). I'm just suggesting keeping the assertion in the test instead of having a separate assertion function. The functionality would still be the same, it's just the readability that changes.

Regarding errors, I'm trying to say that the code I'm commenting on (the loop containing `is.Equal(wantErr.Error.Error(), gotErr.Error.Error())`) seems redundant, as the comparer passed to `cmp.Diff` will produce the same output, for example:

```go
	unwrap_debezium_test.go:235:  !=   sdk.ErrorRecord{
		- 	Error: e"expected error message",
		+ 	Error: e"data to be unwrapped doesn't contain a payload field",
		  }
```

Here's essentially what I'm suggesting: https://github.com/ConduitIO/conduit/commit/51acade531e149503f35ab52dfa2496e70814fd2.

hariso: Ah, I understand what you mean now. Looks good!

pkg/plugin/processor/builtin/unwrap/debezium_test.go
```diff
@@ -0,0 +1,262 @@
...
+	"github.com/conduitio/conduit-commons/opencdc"
+	sdk "github.com/conduitio/conduit-processor-sdk"
+	"github.com/conduitio/conduit/pkg/foundation/cerrors"
+	"github.com/conduitio/conduit/pkg/foundation/log"
+	"github.com/google/go-cmp/cmp"
+	"github.com/google/go-cmp/cmp/cmpopts"
+	"github.com/matryer/is"
+)
+
+var cmpProcessedRecordOpts = []cmp.Option{
```
hariso, line 30: An identical one is in examples_test, but it cannot be used (test file, different package). We plan to move things around a little bit anyway (after we merge the rewritten processors), which is when I think we can find a nice place for these options.

pkg/plugin/processor/builtin/unwrap_debezium.go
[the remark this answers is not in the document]
hariso: I'm not strongly on either side.:) I've refactored the code according to the suggestion you made.

pkg/plugin/processor/builtin/unwrap/debezium.go
```diff
@@ -0,0 +1,335 @@
...
+
+func NewDebezium(logger log.CtxLogger) sdk.Processor {
+	return &debeziumProcessor{logger: logger}
+}
+
+func (d *debeziumProcessor) Specification() (sdk.Specification, error) {
+	return sdk.Specification{
+		Name:    "unwrap.debezium",
+		Summary: "Unwraps a Debezium record from the input OpenCDC record.",
+		Description: `This processor unwraps a Debezium record from the input OpenCDC record.
```
lovromazgon, line 71: This is already in the summary, a bit redundant.

lovromazgon (approved):
LGTM!

PR #1395, opened 2024-02-27
BP: Add unwrap.kafkaconnect

hariso:
### Description

Part of #1290.

Reference from the design doc: https://github.com/ConduitIO/conduit/blob/main/docs/design-documents/20231117-better-processors.md#unwrapkafkaconnect

Notes:
1. The implementation doesn't run on `.Key` as it looks like we don't have a use case for that.
2. `field` is not required. The implementation assigns the default value `.Payload.After`, which should be a good default in most cases.

### Quick checks:


pkg/plugin/processor/builtin/unwrap/unwrap_kafkaconnect.go
```diff
@@ -0,0 +1,176 @@
...
+	}
+
+	return sdk.SingleRecord{
+		Key:      u.unwrapKey(rec.Key),
+		Position: rec.Position,
+		Metadata: rec.Metadata,
+		Payload: opencdc.Change{
+			After: opencdc.StructuredData(structPayload),
+		},
+		Operation: opencdc.OperationSnapshot,
```
hariso, line 136: @lovromazgon @maha-hajja I wanted to clarify this. IIUC, given that these records come from Kafka, where we don't have any updates or deletes, the operation can only be `Create` or `Snapshot`. Is there a special reason why we chose `Snapshot` here? Taken from: https://github.com/ConduitIO/conduit/blob/main/pkg/processor/procbuiltin/unwrap.go#L336

pkg/plugin/processor/builtin/unwrap/unwrap_kafkaconnect.go
```diff
@@ -0,0 +1,176 @@
...
+	// get payload
+	structPayload, ok := kc["payload"].(map[string]any)
+	if !ok {
+		return nil, cerrors.New("referenced record field doesn't contain a payload field")
+	}
+
+	return sdk.SingleRecord{
+		Key:      u.unwrapKey(rec.Key),
+		Position: rec.Position,
+		Metadata: rec.Metadata,
```
hariso, line 132: @maha-hajja In the existing processor (in the legacy processor [here](https://github.com/ConduitIO/conduit/blob/main/pkg/processor/procbuiltin/unwrap.go#L331)), the metadata is always `nil`. It looks like it's an oversight (according to Lovro as well). Are we missing something, or should we correct this?

pkg/plugin/processor/builtin/unwrap/unwrap_kafkaconnect.go
lovromazgon: Hmm, that should probably be `OperationCreate` if you ask me.

maha-hajja: I don't think there was a specific reason no, I agree with changing it to `OperationCreate`

pkg/plugin/processor/builtin/unwrap/unwrap_kafkaconnect.go
maha-hajja: yep definitely an oversight! good catch :D

pkg/plugin/processor/builtin/unwrap/unwrap_kafkaconnect.go
hariso: Thanks!

pkg/plugin/processor/builtin/unwrap/unwrap_kafkaconnect.go
hariso: Thanks!

lovromazgon (approved):
LGTM! Just the description again, maybe repeating the summary will sound repetitive in the docs.

pkg/plugin/processor/builtin/unwrap_examples_test.go
```diff
@@ -0,0 +1,100 @@
...
+				},
+				"schema": map[string]interface{}{},
+			},
+			Payload: opencdc.Change{
+				After: opencdc.RawData(`{
+						"payload": {
+							"description": "test2"
+						},
+						"schema": {}
+					}`),
```
lovromazgon, line 55: btw, this formatting looks _very_ strange in the generated docs, I'd lose the tabs and use spaces if you ask me, also move everything to the left.
```suggestion
  "payload": {
    "description": "test2"
  },
  "schema": {}
}`),
```

lovromazgon: Or maybe even shove everything in a single line?

hariso: I've moved everything to the left, since it's nicer to read in the test, and might produce a nice diff. If it's ugly in the docs, we can always change it.

PR #1401, opened 2024-03-01
BP: Add encode.avro

hariso:
### Description

The `schemaregistry` and `internal` packages have been taken from the legacy processor. The only change made to those is the replacement of the `record` package with the `opencdc` package (from conduit-commons). 

The PR also updates some of the generated code for other processors' configurations (due to an update in paramgen).

Part of #1290 

Reference from the design doc: https://github.com/ConduitIO/conduit/blob/main/docs/design-documents/20231117-better-processors.md#encodeavro

As per a conversation in a review of this PR, the processor will be renamed to `avro.encode`.

### Quick checks:


pkg/plugin/processor/builtin/avro/encode.go
```diff
@@ -0,0 +1,165 @@
...
+	encoder encoder
+}
+
+func NewEncodeProcessor(logger log.CtxLogger) sdk.Processor {
+	return &encodeProcessor{logger: logger}
+}
+
+func (p *encodeProcessor) Specification() (sdk.Specification, error) {
+	return sdk.Specification{
+		Name:    "encode.avro",
```
lovromazgon, line 50: In the design doc we said we should have `encode.avro` and `decode.avro`, this would mean that the code would need to live in the package `encode`. Obviously, it makes more sense to have both processors in the same package as they are sharing functionality, so I'm thinking if we should rather rename the processor to `avro.encode` and `avro.decode`? Same for the JSON processor. CC @maha-hajja

hariso: Makes sense to me. :+1:

pkg/plugin/processor/builtin/avro_examples_test.go
```diff
@@ -28,7 +28,7 @@ import (
 
 //nolint:govet // a more descriptive example description
 func ExampleEncodeProcessor_autoRegister() {
-	url, cleanup := schemaregistry.ExampleSchemaRegistryURL("ExampleEncodeProcessor_autoRegister")
+	url, cleanup := schemaregistry.ExampleSchemaRegistryURL("ExampleEncodeProcessor_autoRegister", 54321)
```
hariso, line 31: Are example tests ever run in parallel (asking because the same port is used for the HTTP example)?

lovromazgon: I don't think they are, even if we used the `-parallel` flag it would only run tests that call `t.Parallel()`, but since there's no such function in examples I'd expect that they are always run sequentially.

lovromazgon (approved):
LGTM!

PR #1402, opened 2024-03-04
Builtin processors refactoring, housekeeping

lovromazgon:
### Description

The changes:
- Processor implementations are moved into folder `impl`
- Processors are structured in folders according to their name (e.g. `field.subset.exclude` is in `impl/field/subset/exclude.go`)
- Example utilities are moved into `internal/exampleutil`
- The `Example` struct now also has a field `Summary` which is displayed in docs as the header, while the description can be longer and is displayed as normal text
- Specification JSON files are now collected and exported into separate files in the folder `internal/exampleutil/specs`
- I have done some housekeeping around naming to keep things consistent between processors

I may have also changed a couple of descriptions here and there.

### Quick checks:

- [X] I have written unit tests. (existing ones still pass)


raulb (approved):
Changes look good to me. I think it makes sense to start separating these processors in different packages.

PR #1405, opened 2024-03-05
add json.decode processor

maha-hajja:
### Description

Related to  #1290 

### Quick checks:


pkg/plugin/processor/builtin/jsonDecode.go
```diff
@@ -0,0 +1,120 @@
...
+type jsonDecodeConfig struct {
+	// Field is the target field, as it would be addressed in a Go template (e.g. `.Payload.After`).
+	// Applicable values are `.Key`, `.Payload.Before` and `.Payload.After`, as they accept structured data format.
+	Field string `json:"field" validate:"required,inclusion=.Key|.Payload.Before|.Payload.After"`
+}
+
+func (p *jsonDecode) Specification() (sdk.Specification, error) {
+	return sdk.Specification{
+		Name:    "json.decode",
+		Summary: "Decodes a specific field from raw data (string) to JSON structured data.",
```
lovromazgon, line 47:
```suggestion
		Summary: "Decodes a specific field from JSON raw data (string) to structured data.",
```

pkg/plugin/processor/builtin/jsonDecode.go
```diff
@@ -0,0 +1,120 @@
...
+}
+
+func newJSONDecode() *jsonDecode {
+	return &jsonDecode{}
+}
+
+type jsonDecodeConfig struct {
+	// Field is the target field, as it would be addressed in a Go template (e.g. `.Payload.After`).
+	// Applicable values are `.Key`, `.Payload.Before` and `.Payload.After`, as they accept structured data format.
+	Field string `json:"field" validate:"required,inclusion=.Key|.Payload.Before|.Payload.After"`
```
lovromazgon, line 41: I wouldn't be that strict, I'd also support anything under `.Key.*`, `.Payload.Before.*` and `.Payload.After.*`. Imagine pulling records from Postgres and one column contains JSON data that we want to parse, so `.Payload.After["my_json_column"]` is a valid input and a valid use case.

pkg/plugin/processor/builtin/jsonDecode.go
```diff
@@ -0,0 +1,120 @@
...
+	for _, record := range records {
+		rec := record
+		ref, err := p.referenceResolver.Resolve(&rec)
+		if err != nil {
+			return append(out, sdk.ErrorRecord{Error: err})
+		}
+
+		data := ref.Get()
+		switch d := data.(type) {
+		case opencdc.RawData:
```
lovromazgon, line 87: If we add support for fields nested in the key or payload, we'll have to check for `[]byte` and `string` types as well.

pkg/plugin/processor/builtin/jsonDecode.go
```diff
@@ -0,0 +1,131 @@
...
+	}
+	return out
+}
+
+func (p *jsonDecode) Teardown(context.Context) error {
+	return nil
+}
+
+func (p *jsonDecode) setJSONData(bytes []byte, ref sdk.Reference) error {
+	var jsonData opencdc.StructuredData
```
lovromazgon, line 121: This assumes that the JSON data is always an object (map), what if I want to decode an array? 😢

lovromazgon: I added the code myself, to keep this moving.

hariso (approved):

pkg/plugin/processor/builtin/jsonDecode.go
```diff
@@ -0,0 +1,132 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+//go:generate paramgen -output=jsonDecode_paramgen.go jsonDecodeConfig
+
+package builtin
```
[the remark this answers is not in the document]
lovromazgon: I agree, though I'm not sure how much that would interfere with the ongoing work from Maha, and given that we have that refactoring PR I'd say we can do it there.

hariso (approved):
Great team work here!

PR #1410, opened 2024-03-06
BP: add avro.decode

hariso:
### Description

Please include a summary of the change and what type of change it is (new feature, bug fix, refactoring, documentation).
Please also include relevant motivation and context.
List any dependencies that are required for this change.

Fixes # (issue)

### Quick checks:


pkg/plugin/processor/builtin/avro/decode.go
```diff
@@ -0,0 +1,206 @@
...
+		return v, nil
+	case []byte:
+		return v, nil
+	case string:
+		decoded := make([]byte, base64.StdEncoding.DecodedLen(len(v)))
+		n, err := base64.StdEncoding.Decode(decoded, []byte(v))
+		if err != nil {
+			return nil, cerrors.Errorf("couldn't base64-decode data: %w", err)
+		}
+		return opencdc.RawData(decoded[:n]), nil
```
lovromazgon, line 198: Hmm, so the assumption is that the string is coming from a JSON encoded record? 🤔 Is there a use case when this string wouldn't be base64 encoded?

pkg/plugin/processor/builtin/avro/decode.go
```diff
@@ -0,0 +1,206 @@
...
+func (p *decodeProcessor) Specification() (sdk.Specification, error) {
+	return sdk.Specification{
+		Name:    "avro.encode",
+		Summary: "Decodes a field's raw data in the Avro format",
+		Description: `The processor takes raw data (bytes or a Base64-encoded string) in the specified field and decodes 
+it from the [Avro format](https://avro.apache.org/) into structured data. It extracts the schema ID from the data, 
+downloads the associated schema from the [schema registry](https://docs.confluent.io/platform/current/schema-registry/index.html)
+and decodes the payload. The schema is cached locally after it's first downloaded. 
+
+Currently, the processor only supports the Avro format. If the processor encounters structured data or the data
```
lovromazgon, line 118: Well yeah, it's the `avro.decode` processor 😄 This is probably taken from the old processor, which was a "generic" processor.
```suggestion
If the processor encounters structured data or the data 
```

lovromazgon (commented):
We need an example and this should be good to go!

pkg/plugin/processor/builtin/avro/decode.go
hariso: Ah, exactly!

pkg/plugin/processor/builtin/avro/decode.go
lovromazgon: We decided to put this functionality in a separate processor [base64 encode/decode](https://github.com/ConduitIO/conduit/pull/1412).

lovromazgon (approved):
LGTM!

pkg/plugin/processor/builtin/avro/decode.go
hariso: For the record, Lovro and I chatted about this. While it's likely that a string that represents Avro bytes is a Base64 encoding, it doesn't have to be. However, that's not always the case, meaning the processor will not always be usable out of the box. We will provide a Base64 encoder and decoder that can be used in these case.

PR #1411, opened 2024-03-06
 BP: call the conditional execution methods in the processors

### LINKED ISSUE #1389: BP: call the conditional execution methods in the processors

hariso:
### Description

Fixes #1389 

### Quick checks:

### DIFF
--- a/pkg/processor/runnable_processor.go
+++ b/pkg/processor/runnable_processor.go
@@ -94,6 +94,11 @@ func (p *RunnableProcessor) Process(ctx context.Context, records []opencdc.Recor
 
 		if len(keptRecords) > 0 {
 			outRecs = p.proc.Process(ctx, keptRecords)
+			if len(outRecs) > len(keptRecords) {
+				return []sdk.ProcessedRecord{
+					sdk.ErrorRecord{Error: cerrors.New("processor returned more records than input")},
+				}
+			}
 		}
 		if err != nil {
 			outRecs = append(outRecs, sdk.ErrorRecord{Error: err})
@@ -111,6 +116,8 @@ func (p *RunnableProcessor) Process(ctx context.Context, records []opencdc.Recor
 			tmp := make([]sdk.ProcessedRecord, len(outRecs)+len(passthroughRecordIndexes))
 			prevIndex := -1
 			for i, index := range passthroughRecordIndexes {
+				// TODO index-i can be out of bounds if the processor returns
+				//  fewer records than the input.
 				copy(tmp[prevIndex+1:index], outRecs[prevIndex-i+1:index-i])
 				tmp[index] = sdk.SingleRecord(records[index])
 				prevIndex = index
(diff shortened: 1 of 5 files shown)


pkg/processor/runnable_processor.go
```diff
@@ -59,19 +62,39 @@ func (p *RunnableProcessor) Process(ctx context.Context, records []opencdc.Recor
...
+		keep, err := p.evaluateCondition(rec)
+		if err != nil {
+			return append(out, sdk.ErrorRecord{Error: cerrors.Errorf("failed evaluating condition: %w", err)})
+		}
+		if !keep {
+			out = append(out, sdk.FilterRecord{})
+			continue
+		}
+
+		proc := p.proc.Process(ctx, records)
```
hariso, line 76: Just noticed this is wrong, only `rec` should be sent.

pkg/processor/runnable_processor.go
```diff
@@ -59,19 +62,41 @@ func (p *RunnableProcessor) Process(ctx context.Context, records []opencdc.Recor
...
-	for _, outRec := range outRecs {
-		singleRec, ok := outRec.(sdk.SingleRecord)
+	out := make([]sdk.ProcessedRecord, 0, len(records))
+	for _, rec := range records {
+		keep, err := p.evaluateCondition(rec)
+		if err != nil {
+			return append(out, sdk.ErrorRecord{Error: cerrors.Errorf("failed evaluating condition: %w", err)})
+		}
+		if !keep {
+			out = append(out, sdk.FilterRecord{})
```
maha-hajja, line 72: wouldn't this mean that we're filtering out the record when the condition is `false`? from what I understand we should append the original record here

pkg/processor/runnable_processor.go
```diff
@@ -59,19 +62,41 @@ func (p *RunnableProcessor) Process(ctx context.Context, records []opencdc.Recor
...
 }
 
 func (p *RunnableProcessor) Teardown(ctx context.Context) error {
 	err := p.proc.Teardown(ctx)
 	p.running = false
 	return err
 }
+
+func (p *RunnableProcessor) evaluateCondition(rec opencdc.Record) (bool, error) {
+	if p.cond == nil {
```
maha-hajja, line 97: WDYT about setting the default condition to `true` when enriching the pipeline config file too, in `pkg/provisioning/config/enrich.go`

hariso: That's a good idea I believe. The only reason that makes me think we should keep it this way is maybe slightly greater performance, since we won't be executing a template at all (even if it's simple).

pkg/processor/runnable_processor.go
```diff
@@ -59,19 +62,41 @@ func (p *RunnableProcessor) Process(ctx context.Context, records []opencdc.Recor
...
+		keep, err := p.evaluateCondition(rec)
+		if err != nil {
+			return append(out, sdk.ErrorRecord{Error: cerrors.Errorf("failed evaluating condition: %w", err)})
+		}
+		if !keep {
+			out = append(out, sdk.SingleRecord(rec))
+			continue
+		}
+
+		proc := p.proc.Process(ctx, records)
```
lovromazgon, line 76: Ugh, so yeah, this is a workaround that works now only because it's a single record in the slice. Why not just make it work as it's supposed to?

lovromazgon (approved):

PR #1413, opened 2024-03-06
add encodeJson processor

maha-hajja:
### Description

Related to #1290 

### Quick checks:


pkg/plugin/processor/builtin/jsonEncode.go
```diff
@@ -0,0 +1,119 @@
...
+	ref, err := p.referenceResolver.Resolve(&rec)
+	if err != nil {
+		return nil, cerrors.Errorf("failed to resolve the field: %w", err)
+	}
+
+	var raw []byte
+	switch val := ref.Get().(type) {
+	case []byte:
+		raw = val
+		// do we want to fail if the []byte or string is not a valid json string?
```
maha-hajja, line 102: ..

pkg/plugin/processor/builtin/jsonEndoe_test.go
hariso, line 1: If your rename the files as we have them [here,](https://github.com/ConduitIO/conduit/pull/1402/files#diff-b68a74b1729e5ee25c44f45ad7e5c745fbf22e592021296875efed7872487989) in the refactoring PR, it will make the merge easier.:)

pkg/plugin/processor/builtin/jsonEncode.go
hariso: I'm not quite sure this is what we want to do. Shouldn't we do `json.Marshal(val)` regardless of what `val` is? Well, we'll handle a couple of special cases, like when `val` is nil.

pkg/plugin/processor/builtin/jsonEncode.go
```diff
@@ -0,0 +1,119 @@
...
+	"github.com/conduitio/conduit/pkg/foundation/cerrors"
+)
+
+type jsonEncode struct {
+	referenceResolver sdk.ReferenceResolver
+
+	sdk.UnimplementedProcessor
+}
+
+func newJSONEncode() *jsonEncode {
```
hariso, line 33: It would be good for this constructor to accept a log.CtxLogger, so we can add it to the default built-in registry. And a nitpick for the return value, it would be good if it's `sdk.Processor`.

pkg/plugin/processor/builtin/jsonEncode.go
maha-hajja: yes that was my initial thought! then I got lost thinking about other cases and couldn't decide :/ 
if we do that, then any string that isn't a json will fail, which does make sense

pkg/plugin/processor/builtin/jsonEncode.go
```diff
@@ -0,0 +1,119 @@
...
+	// Field is the target field, as it would be addressed in a Go template (e.g. `.Payload.After.foo`).
+	// you can only encode fields that are under .Key, .Payload.Before and .Payload.After.
+	Field string `json:"field" validate:"required,regex=^\\.(Payload|Key).*,exclusion=.Payload"`
+}
+
+func (p *jsonEncode) Specification() (sdk.Specification, error) {
+	return sdk.Specification{
+		Name:    "json.encode",
+		Summary: "Encodes a specific field from structured data to JSON raw data (string).",
+		Description: `The processor takes structured data from the target field, encodes it into JSON raw data (string)
```
hariso, line 47: > The processor takes structured data

Related to another comment: shouldn't we be able to take any type of data and JSON-ify it?

pkg/plugin/processor/builtin/jsonEncode.go
```diff
@@ -0,0 +1,119 @@
...
+
+	var raw []byte
+	switch val := ref.Get().(type) {
+	case []byte:
+		raw = val
+		// do we want to fail if the []byte or string is not a valid json string?
+	case opencdc.RawData:
+		raw = val
+	case opencdc.StructuredData:
+		raw = val.Bytes()
```
lovromazgon, line 106: I'd rather manually call `json.Marshal` so we don't rely on the implementation details of `Bytes()`.

pkg/plugin/processor/builtin/jsonEncode.go
lovromazgon: I agree, we should call `json.Marshal` on whatever the value is, I think we shouldn't even care about the type at all. Like, if it's an `int` that's fine, if it's a slice, map, whatever. Just throw it in the JSON meatgrinder and collect the pieces on the other side 🥩 🤷

pkg/plugin/processor/builtin/jsonEncode.go
```diff
@@ -0,0 +1,119 @@
...
+	case nil:
+		return sdk.SingleRecord(rec), nil
+	default:
+		return nil, cerrors.Errorf("unexpected data type %T", val)
+	}
+
+	err = ref.Set(opencdc.RawData(raw))
```
lovromazgon, line 113: No need to cast the value to `opencdc.RawData`, the reference resolver can handle it. If it's going to set fields like `.Key`, `.Payload.After` or `.Payload.Before` it will set the bytes as raw data, otherwise it will leave them as `[]byte`.

pkg/plugin/processor/builtin/jsonEncode.go
hariso: Exactly, the output doesn't have to be a JSON object, but a JSON value.

pkg/plugin/processor/builtin/jsonEndoe_test.go
maha-hajja: done, only need to update the test to have access to `internal.cmpProcessedRecordOpts`, as mentioned in the todos

pkg/plugin/processor/builtin/jsonEncode.go
maha-hajja: done! 
the design document does mention that it converts from structured data, so maybe we should update it there too?
or maybe not, cuz it's only a design doc and implementation can change, I don't mind either way 🤷‍♀️

hariso: IIRC, sometimes earlier we said we won't be changing design docs, as they show the design at a point in time. The user-facing docs will explain how it works anyway. :+1:

pkg/plugin/processor/builtin/json/encode_test.go
```diff
@@ -0,0 +1,179 @@
...
+			name:   "nil value",
+			config: map[string]string{"field": ".Payload.After"},
+			record: opencdc.Record{
+				Payload: opencdc.Change{
+					After: nil,
+				},
+			},
+			want: sdk.SingleRecord{
+				Payload: opencdc.Change{
+					After: opencdc.RawData("null"),
```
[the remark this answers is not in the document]
lovromazgon: Hmm, yeah, maybe that would be better. I think I implemented it the same way when implementing `base64.encode`.

pkg/plugin/processor/builtin/jsonEncode.go
lovromazgon: If anything, I'd add a note to the design doc instead of changing it. But I don't see that as a high priority right now.

pkg/plugin/processor/builtin/json/encode_test.go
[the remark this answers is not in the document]
lovromazgon: Done. Also changed the output of the processor to be a `string`.

hariso (approved):

PR #1416, opened 2024-03-07
Remove obsolete page about processors

hariso:
### Description

The page is obsolete, all the docs are on conduit.io

### Quick checks:


maha-hajja (approved):

README.md
```diff
@@ -283,7 +283,7 @@ documentation:
   a Conduit pipeline works.
 - [Pipeline Configuration Files](docs/pipeline_configuration_files.md)
   explains how you can define pipelines using YAML files.
-- [Processors](docs/processors.md) contains examples and more information about
+- [Processors](https://conduit.io/docs/processors/getting-started) contains examples and more information about
```
maha-hajja, line 286: oh we might change this, one of the doc updates items was (delete (getting started with processors) page)
we can keep the name but we should double-check when it's updated that it matches here

hariso: Oh, right! Let's then hold off from this, until that is done.

raulb: I don't think we should delete it, but rather update it, right? Ultimately, we do want to guide users on how to get started with processors, so I think that page would be valuable. This is something @lovromazgon and I were just discussing. Made part of https://github.com/ConduitIO/conduit/issues/1422

i.e.: I think it could be fine just leaving the link on `processors/getting-started` for now. WDYT?

raulb (approved):
LGTM!

README.md
hariso: If that's the case, I'll merge this one!

PR #1439, opened 2024-03-14
Make sure plugins with non-semantic version are registered as `latest`

lovromazgon:
### Description

Given this change https://github.com/ConduitIO/conduit-processor-sdk/pull/31 it would be nice if Conduit allowed you to use `my-processor@latest` (or no version) when using a processor without a semantic version. This fixes the behavior both for connectors and processors.

### Quick checks:

### DIFF
--- a/pkg/plugin/name.go
+++ b/pkg/plugin/name.go
@@ -72,17 +72,27 @@ func (fn FullName) PluginVersion() string {
 	return PluginVersionLatest // default
 }
 
-func (fn FullName) PluginVersionGreaterThan(other FullName) bool {
-	leftVersion := fn.PluginVersion()
-	rightVersion := other.PluginVersion()
-
-	leftSemver, err := semver.NewVersion(leftVersion)
-	if err != nil {
-		return false // left is an invalid semver, right is greater either way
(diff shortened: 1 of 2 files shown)


pkg/plugin/name.go
```diff
@@ -72,18 +72,29 @@ func (fn FullName) PluginVersion() string {
...
-func (fn FullName) PluginVersionGreaterThan(other FullName) bool {
-	leftVersion := fn.PluginVersion()
-	rightVersion := other.PluginVersion()
+func (fn FullName) PluginVersionGreaterThan(right FullName) bool {
+	if right == "" {
+		return true // right is empty, left is greater either way
+	}
+	if fn == "" {
+		return true // left is empty, right is greater either way
+	}
```
raulb, line 81: I understand this could be called either from a processor or a connector, maybe we could simplify it this way, what do you think?

```suggestion
	if right == "" || fn == "" {
                return true // if either right or left is empty, the other is greater either way
         }
```

raulb (approved):

pkg/plugin/name.go
lovromazgon: yeah, we could do that, if that wouldn't be a typo :D the second branch should return `false` 🙈

PR #1447, opened 2024-03-19
Add more example pipelines

### LINKED ISSUE #1118: Add more example pipelines
### Feature description

Originally suggested by @ahmeroxa:

Getting started with Conduit pipelines would be much easier if we had more examples of pipelines of different types (multiple sources, destinations, with and without processors, with and without DLQ, etc.)

Examples:

- [x] multiple sources
- [x] multiple destinations
- [x] source processor
- [ ] destination processors
- [x] js processor
- [x] usage of standalone connector
- [x] pipeline with a processor
- [x] pipeline with a dlq

hariso:
### Description

Closes #1118.

Note that I didn't add an example of a destination processor, since there's already an example of a source processor.

### Quick checks:


raulb (approved):
@hariso Looks good. I was thinking that maybe we could incorporate these to our public docs. The only one we've got now there is the one [from file to file](https://conduit.io/docs/pipeline-configuration-files/getting-started). At the very least, we could add a callout from docs to these. What do you think?

hariso:
@raulb Good that you remember that one... The file-to-file example on conduit.io mentions the general steps for running a pipeline file, so maybe we can add a link to this directory?

raulb:
@hariso yup. I will also include the ones you added here. Thank you for adding them in the first place. 🥂

PR #1470, opened 2024-04-04
chore: upgrade ember-cli-mirage to 3.0.3

raulb:
Will replace https://github.com/ConduitIO/conduit/pull/1399


lovromazgon (approved):
LGTM, I clicked through the UI, seems like it all works!

PR #1498, opened 2024-04-17
add dynamic url to HTTP processor + fix RawData bug

### LINKED ISSUE #1495: HTTP Processor: dynamic `request.url`
### Feature description

The parameter `request.url` is currently a static string, which means that building the URL using data taken from the record is not possible. We should evaluate the parameter as a Go template, to enable this functionality. It is particularly useful when query parameters need to be set dynamically.

Example:
```
request.url: "http://example.com/{{.Payload.After.something}}"
```

### LINKED ISSUE #1497: HTTP Processor: change RawData into []byte
### Bug description

the processor response is saved as `RawData` even if it's under a `StructuredData` field.
investigate the problem, and change `RawData` into `[]byte` (should work for `.Payload.After` and other fields)

### Steps to reproduce

example pipeline config file:

```
version: 2.2
pipelines:
  - id: sfdc-to-webhook
    status: running
    description: >
      TBD.
    connectors:
      - id: employees-source
        type: source
        plugin: builtin:generator
        settings:
          format.type: "structured"
          format.options: "id:int,name:string,company:string,trial:bool"
          recordCount: "15"
          readTime: "1s"
      - id: webhook-sink
        type: destination
        plugin: standalone:http
        settings:
          url: "https://webhook.site/4fd2add3-db9f-4fe2-bf4d-c1cc88e60dff"
    processors:
      - id: fetch-pokemon-ditto
        plugin: "webhook.http"
        settings:
          backoffRetry.count: "0"
          backoffRetry.factor: "2"
          backoffRetry.max: "5s"
          backoffRetry.min: "100ms"
          request.body: ""
          request.contentType: "application/json"
          request.method: "GET"
          request.url: "https://pokeapi.co/api/v2/pokemon/ditto"
          response.body: ".Payload.After.response"
          response.status: ".Metadata[\"http_status\"]"
#      - id: decode-json
#        plugin: json.decode
#        settings:
#          field: ".Payload.After.response"

```
doesn't work without the json decoder processor.

### Version

latest

maha-hajja:
### Description

Fixes #1495 
Fixes #1497 

### Quick checks:


hariso (approved):

pkg/plugin/processor/builtin/impl/webhook/http_examples_test.go
```diff
@@ -85,15 +85,76 @@ value of the HTTP response's code in the metadata field ` + "`http_status`" + `.
...
+	})
+
+	// Output:
+	// processor transformed record:
+	// --- before
+	// +++ after
+	// @@ -1,12 +1,13 @@
+	//  {
+	//    "position": null,
+	//    "operation": "Operation(0)",
```
lovromazgon, line 131: We should at least have a valid operation here, maybe also a position, otherwise it looks a bit strange.

pkg/plugin/processor/builtin/impl/webhook/http.go
```diff
@@ -139,6 +153,18 @@ func (p *httpProcessor) Configure(ctx context.Context, m map[string]string) erro
...
 }
 
+func (p *httpProcessor) EvaluateURL(rec opencdc.Record) (string, error) {
+	if p.urlTmpl == nil {
+		return p.config.URL, nil
+	}
+	var b bytes.Buffer
+	err := p.urlTmpl.Execute(&b, rec)
+	if err != nil {
+		return p.config.URL, cerrors.Errorf("error while evaluating URL template: %w", err)
```
lovromazgon, line 163:
```suggestion
		return "", cerrors.Errorf("error while evaluating URL template: %w", err)
```

hariso:
@maha-hajja LGTM! Maybe we could also document which types are expected in `StructuredData` values? Or maybe even have methods on `StructuredData` that will check a value's type before storing it?

pkg/plugin/processor/builtin/impl/webhook/http_test.go
```diff
@@ -342,6 +356,99 @@ func TestHTTPProcessor_Success(t *testing.T) {
...
+	}
+
+	for _, tc := range tests {
+		t.Run(tc.name, func(t *testing.T) {
+			is := is.New(t)
+
+			wantMethod := tc.config["request.method"]
+			if wantMethod == "" {
+				wantMethod = "POST" // default
+			}
```
lovromazgon, line 417: Nitpick: if we're focusing on the URL template (as the name of the test suggests), we shouldn't be concerned with the request method, request body or response body. That's for other tests to assert, so this one can be much more focused.

lovromazgon (approved):
Looks good, I left a few suggestions for improvements 😉

PR #1499, opened 2024-04-18
http processor: replace space with +

maha-hajja:
### Description

if the URL template values contain spaces, the spaces will be replaced by `+`

### Quick checks:

### DIFF
--- a/pkg/plugin/processor/builtin/impl/webhook/http.go
+++ b/pkg/plugin/processor/builtin/impl/webhook/http.go
@@ -164,7 +164,8 @@ func (p *httpProcessor) EvaluateURL(rec opencdc.Record) (string, error) {
 	if err != nil {
 		return "", cerrors.Errorf("error while evaluating URL template: %w", err)
 	}
-	return b.String(), nil
+	url := strings.ReplaceAll(b.String(), " ", "+")
+	return url, nil
 }
 
 func (p *httpProcessor) Process(ctx context.Context, records []opencdc.Record) []sdk.ProcessedRecord {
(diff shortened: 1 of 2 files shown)


pkg/plugin/processor/builtin/impl/webhook/http.go
```diff
@@ -164,7 +164,8 @@ func (p *httpProcessor) EvaluateURL(rec opencdc.Record) (string, error) {
 	if err != nil {
 		return "", cerrors.Errorf("error while evaluating URL template: %w", err)
 	}
-	return b.String(), nil
+	url := strings.ReplaceAll(b.String(), " ", "+")
```
hariso, line 167: Alternatively, you can also use the `urlquery` function which is available in Go templates.

lovromazgon: I was first thinking we shouldn't do this and let the user use `urlquery` as Haris proposed. Then I dug deeper and found out that `url.Parse` will already replace spaces in the path with `%20`, while it will complain if the protocol or hostname contain a space. So I wouldn't want to change spaces elsewhere except in the query string.

Another reason for this can be found in this [SO question](https://stackoverflow.com/questions/1634271/url-encoding-the-space-character-or-20), which explains that `+` is only applicable in the query string.

So I'm on board with making this improvement, but I'd make it smarter and only target the query string. Luckily Go provides utilities for us to safely do this:

```suggestion
	u, err := url.Parse(b.String())
	// handle error ...

	// parse query and encode it
	q, err := url.ParseQuery(u.RawQuery)
	// handle error ...
	u.RawQuery = q.Encode()	// encode query URL
```

Here's a snippet of what this will do: https://go.dev/play/p/COtRw6aHhYI

hariso: That makes sense... I also see we're already doing that in the HTTP connector.

lovromazgon (approved):

PR #1515, opened 2024-04-25
Upgrade builtin connectors, SDK and commons

lovromazgon:
### Description

Upgrade builtin connectors, connector SDK and conduit commons to latest version.

Processor specs needed to be regenerated, as we don't show parameters in examples if the parameter value is empty. Also I had to add `maps.Clone` in 3 places, so when we pass a config to the processor or connector, it can freely modify it without changing the map that's stored in Conduit.


pkg/plugin/connector/builtin/v1/internal/toplugin/destination.go
```diff
@@ -23,7 +25,7 @@ func DestinationConfigureRequest(in map[string]string) cpluginv1.DestinationConf
 	out := cpluginv1.DestinationConfigureRequest{}
 	if len(in) > 0 {
 		// gRPC sends `nil` if the map is empty, match behavior
-		out.Config = in
+		out.Config = maps.Clone(in)
```
hariso, line 28: It's interesting that we're cloning it. Why so?

hariso (approved):

pkg/plugin/connector/builtin/v1/internal/toplugin/destination.go
lovromazgon: In builtin plugins we were passing the very same map (pointer) that was used to store the config in the `Processor` and `Connector` entities. So if the plugin changes that map, we will see the change. Now we clone it so the plugin can do whatever it wants with that map without side effects.

lovromazgon: Btw, why the change now? It showed up as a race condition in tests. Apparently a change in the SDK now causes the map to be manipulated.

PR #1532, opened 2024-05-02
Schema support (design doc)

hariso:
### Description

This PR documents the design for Conduit's support for schemas. A few questions are still open, and, from that point, the design is not completely done.

### Quick checks:


docs/design-documents/20240430-schema-support.md
```diff
@@ -0,0 +1,161 @@
...
+
+This design document focuses on the schema support in Conduit in the context of
+the above.
+
+## Requirements
+
+1. Records **should not** carry the full schema. 
+
+   Reason: If a record would carry the whole schema, that would increase the
+   record size a lot.
```
lyuboxa, line 23: I think "a lot" is relative to the size of the schema and records. 👍

docs/design-documents/20240430-schema-support.md
```diff
@@ -0,0 +1,161 @@
...
+  * default value
+
+The following types are supported:
+* Primitive:
+  * boolean
+  * integers: 8, 16, 32, 64-bit 
+  * float: single precision (32-bit) and double precision (64-bit) IEEE 754 floating-point number
+  * bytes
+  * string
+* Complex:
```
lyuboxa, line 65: Do want to add "time" as a support type? Or are these just native golang/protobuf types?

hariso: No special reason why I left it out, I'll add it.:)

lyuboxa: There are probably going to be more, but Time seems to be the one persistent across most payloads

docs/design-documents/20240430-schema-support.md
lyuboxa: Thanks, my thinking is that there can be very payload heavy records with simple schema and then there can be big schema of ints.. so the runes to describe will probably take more space than the numbers.

docs/design-documents/20240430-schema-support.md
```diff
@@ -0,0 +1,281 @@
...
+remote Conduit instances.
+
+## Questions
+
+1. **Q**: Should a schema be accessible across Conduit instances?
+
+   Schemas need to be accessible across pipelines (requirement #3). If the
+   schema registry is part of Conduit, should we make it accessible via the API?
+   
+   **A**:
```
hariso, line 265: Running one pipeline per Conduit instance could be a thing, so it looks like the answer to this is "yes".

docs/design-documents/20240430-schema-support.md
```diff
@@ -0,0 +1,281 @@
...
+2. **Q**: Should Conduit allow schemas from other streaming tools?
+
+   Conduit is sometimes used together with other streaming tools, such as Kafka
+   Connect. For example, Kafka Connect is used to stream data to a topic, and
+   then Conduit is used to read the data from the topic and write it into a
+   destination.
+
+   **A**:
```
hariso, line 273: We also have the Kafka Connect wrapper... So it looks like the answer is "yes"?

docs/design-documents/20240430-schema-support.md
```diff
@@ -0,0 +1,394 @@
...
+   cost of repeatedly fetching or creating the same schema many times (
+   especially over gRPC), schemas should be cached by the SDK.
+
+## Non-goals
+
+- Schema auto-generation: Given that a source connector usually has direct
+  access to a source collection, and also knows most about the source types,
+  schema auto-generation is not a requirement.
+
+## Schema format
```
hariso, line 88: This section describes the types we need. I'll mention our options for the actual schema format (our own, Avro, Protobuf schema, etc.)

docs/design-documents/20240430-schema-support.md
```diff
@@ -0,0 +1,411 @@
...
+**Disadvantages**:
+
+1. Protobuf libs don't offer a way to create a `.proto` file, i.e. that needs to
+   be done manually.
+
+#### Chosen option
+
+
+
+### Schema service interface
```
hariso, line 245: Explore possibility of "reversing" a connection (Conduit initiates connection to connector, then connector establishes a connection to Conduit)

docs/design-documents/20240430-schema-support.md
```diff
@@ -0,0 +1,411 @@
...
+message FetchSchemaResponse {}
+```
+
+**Advantages**: 
+
+1. Easy to understand: the gRPC methods, together with requests and responses,
+   can easily be understood from a proto file.
+2. An HTTP API for the schema registry can easily be exposed (if needed).
+
+**Disadvantages**:
```
hariso, line 337: Might be a security concern.

docs/design-documents/20240430-schema-support.md
```diff
@@ -0,0 +1,411 @@
...
+  string id = 1;
+}
+message FetchSchemaResponse {}
+```
+
+**Advantages**: 
+
+1. Easy to understand: the gRPC methods, together with requests and responses,
+   can easily be understood from a proto file.
+2. An HTTP API for the schema registry can easily be exposed (if needed).
```
hariso, line 335: Could be become an "internal connector utilities service" (storage, etc. can be added)

docs/design-documents/20240430-schema-support.md
```diff
@@ -0,0 +1,411 @@
...
+
+**Disadvantages**:
+
+1. Changes needed to communicate Conduit's gRPC port to the connector.
```
hariso, line 339: Streams are faster (not much of a problem, since schema changes happen less often)

docs/design-documents/20240430-schema-support.md
hariso: Note: running a service on a random port might not work on security hardened VMs.

docs/design-documents/20240430-schema-support.md
```diff
@@ -0,0 +1,411 @@
...
+
+#### Chosen option
+
+**Option 2** is the chosen method since it offers more clarity and the support
+for remote Conduit instances.
```
hariso, line 344: Use random port, add some security through a token used only by a connector.

docs/design-documents/schema_service.proto
```diff
@@ -0,0 +1,20 @@
+syntax = "proto3";
```
lovromazgon, line 1: Do we need this file? It's already embedded in the design document itself.

docs/design-documents/20240430-schema-support.md
```diff
@@ -0,0 +1,413 @@
...
+
+A skeleton of the gRPC definition of the service would be:
+
+```protobuf
+syntax = "proto3";
+
+service SchemaService {
+  rpc Create(CreateSchemaRequest) returns (CreateSchemaResponse);
+  rpc Fetch(FetchSchemaRequest) returns (FetchSchemaResponse);
+}
```
lovromazgon, line 325: I'm thinking if we should name this `ConnectorUtilService` and in the future extend it with more calls for other functionalities (e.g. storage).

```suggestion
service ConnectorUtilService {
  rpc CreateSchema(CreateSchemaRequest) returns (CreateSchemaResponse);
  rpc FetchSchema(FetchSchemaRequest) returns (FetchSchemaResponse);
  
  // Down the road:
  rpc Set(SetRequest) returns (SetResponse);
  rpc Get(GetRequest) returns (GetResponse);
  rpc Delete(DeleteRequest) returns (DeleteResponse);
}
```

We could as well leave this as `SchemaService` and then introduce `StorageService`. Maybe that's even better and we should leave it as it is. I'm just thinking out loud here 🙃

docs/design-documents/20240430-schema-support.md
```diff
@@ -0,0 +1,413 @@
...
+   
+   Reason: Multiple collections support.
+3. A schema should be accessible across pipelines and Conduit instances.
+
+   Reason: Pipelines might work together to stream data from A to B, through an
+   intermediary C. For example, PostgreSQL to Kafka, and then Kafka to
+   Snowflake. Also, in the future, pipelines might run in isolation on different
+   instances. Additionally, Conduit may be used so that there's one pipeline per
+   instance. In such cases, we'd need this feature.
+4. It should be possible for a schema to evolve.
```
lovromazgon, line 56: Are we addressing this requirement? I can't find anything about schema versioning, which could probably address this part.

docs/design-documents/20240430-schema-support.md
```diff
@@ -0,0 +1,413 @@
...
+10. Schema auto-generation
+
+    Reason: Schema auto-generation is a useful feature when working with data
+    sources that have structured data encoded into a binary format (e.g. a JSON
+    file). In such cases, it's useful to have the schema auto-generated as a
+    starting point for further processing and/or writing into a destination.
+
+## Schema structure
+
+A destination connector should work with one schema format only, regardless of
```
lovromazgon, line 75: Hmm, didn't we decide in the end that we will support multiple schema types, but will focus on Avro for the time being? This makes the contents of this chapter depend on the format that's being used.

docs/design-documents/20240430-schema-support.md
```diff
@@ -0,0 +1,413 @@
...
+the underlying or intermediary schema formats used. This makes it easy for a
+connector developer to write code, since it doesn't require handling of
+potentially multiple schema types.
```
lovromazgon, line 78: IIRC we said the schema handling will be transparent to the destination connector, in the sense that if the source connector provides a schema, the destination connector will get structured data with the correct types. If they need a schema to update the destination data structure, they can get the schema, but if they don't, they can just work on the structured data.

In the end, I think we should avoid as much manual schema handling in destination connectors as possible.

docs/design-documents/20240430-schema-support.md
```diff
@@ -0,0 +1,420 @@
...
+
+1. A function that registers a schema.
+2. A function that fetches a schema.
+
+### Processor SDK
+
+The Processor SDK needs to provide the following functions:
+
+1. A function that registers a schema.
+2. A function that fetches a schema.
```
lovromazgon, line 394: This should go into conduit commons.

docs/design-documents/schema_service.proto
hariso: Definitely not!

docs/design-documents/20240430-schema-support.md
```diff
@@ -0,0 +1,420 @@
...
+### Conduit Commons
+
+Conduit Commons needs to provide the following functions that will be used by
+multiple libraries (Connector SDK, Processor SDK):
+
+1. A function that creates an Avro schema
+2. A function that encodes values using an Avro schema
+3. A function decodes a slice of bytes into a value, using an Avro schema
+
+### Connector SDK
```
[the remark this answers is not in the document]
hariso: Functions for registering a schema and encoding are already mentioned, so do you mean a single function for all that?

docs/design-documents/20240430-schema-support.md
hariso: The reason I put it here is because we might need different mechanisms for connectors and processors. For standalone connectors it makes sense to gRPC calls to Conduit. For built-in ones, maybe less so. But standalone processors won't be able to do anything on the network, so we'll probably need to do it through the usual commands-and-response thingy we have. I believe we can figure out the details once we come to that.

docs/design-documents/20240430-schema-support.md
hariso: Yup, this is outdated (coming from the initial though to use a single format in the SDK, and then a different internal format).

docs/design-documents/20240430-schema-support.md
```diff
@@ -0,0 +1,420 @@
...
+#### Chosen option
+
+**Option 2** is the chosen method since it offers more clarity and the support
+for remote Conduit instances.
+
+## Required changes
+
+### Conduit
+
+Conduit needs to expose a gRPC schema service as explained above. The gRPC
```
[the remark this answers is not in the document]
hariso: Isn't the connector protocol meant for services a connector exposes? It's also not really connector specific (can be accessed by processors too).

docs/design-documents/20240430-schema-support.md
lovromazgon: Oh, you're right, these can't go into commons, my bad!

docs/design-documents/20240430-schema-support.md
hariso: From our discussions, it didn't seem that we had a use case for working with schema versions directly. The purpose of this requirement is just to highlight that a schema might evolve, and that a pipeline should be able to handle it (e.g. schema is not overwritten, older schemas can still be fetched with their IDs, and so on). I can rephrase this requirement to clarify it.

docs/design-documents/20240430-schema-support.md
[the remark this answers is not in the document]
lovromazgon: Yeah, it's connector specific, WASM processors should get their own interface, and that one probably won't be through gRPC, since they can directly call into Conduit.

docs/design-documents/20240430-schema-support.md
hariso: That's fine, it's always good to double check.:)

docs/design-documents/20240430-schema-support.md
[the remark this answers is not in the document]
lovromazgon: I mean a connector middleware ([source](https://github.com/ConduitIO/conduit-connector-sdk/blob/dee368320087f1804ecffbb0813ce834a3019ab6/source_middleware.go#L18-L20), [destination](https://github.com/ConduitIO/conduit-connector-sdk/blob/dee368320087f1804ecffbb0813ce834a3019ab6/destination_middleware.go#L28-L30)), so that schema handling can be automated for all connectors. It's related to what I mentioned in that [other comment](https://github.com/ConduitIO/conduit/pull/1532#discussion_r1603331495).

docs/design-documents/20240430-schema-support.md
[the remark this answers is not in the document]
hariso: Right, processors are out of the equation here. Still, isn't the connector protocol meant for services a connector exposes?

docs/design-documents/20240430-schema-support.md
hariso: Yup, I'll update this part!

docs/design-documents/20240430-schema-support.md
hariso: I was thinking about something similar too... Naming it `ConnectorUtilService` would make it clear what clients should connect to it. On the other hand, having very different methods in a single service might make the service confusing. 

To me, the second option is slightly better. We can make it clear in the docs that connectors are supposed to be using the service (and it'll be running on a separate port too, so it'll make it even more clear). I don't think we'll have problems changing the name (since client code won't care much about it).

docs/design-documents/20240430-schema-support.md
```diff
@@ -0,0 +1,420 @@
...
+
+However, **our goal is to eventually implement option 1**.
+
+### Schema format
+
+This section discusses the schema format to be used.
+
+There are two aspects of this:
+
+- the schema format used internally (when registering a schema in the schema
```
raulb, line 188:
```suggestion
1. The schema format used internally (when registering a schema in the schema
```

docs/design-documents/20240430-schema-support.md
```diff
@@ -0,0 +1,420 @@
...
+  service, updating it, fetching, etc.)
+- the schema format exposed to connector developers (through the Connector SDK)
```
raulb, line 190:
```suggestion
2. The schema format exposed to connector developers (through the Connector SDK)
```

docs/design-documents/20240430-schema-support.md
```diff
@@ -0,0 +1,420 @@
...
+However, **our goal is to eventually implement option 1**.
+
+### Schema format
+
+This section discusses the schema format to be used.
+
+There are two aspects of this:
+
+- the schema format used internally (when registering a schema in the schema
+  service, updating it, fetching, etc.)
```
raulb, line 189:
```suggestion
  service, updating, fetching, etc.)
```

docs/design-documents/20240430-schema-support.md
```diff
@@ -0,0 +1,420 @@
...
+Having them different makes only sense if we expose our own format in the
+Connector SDK. The internal format is dictated by the schema registry that we
+will use, which supports only widely known formats, i.e. we won't be able to use
+our own schema format for that purpose.
+
+**Advantages**:
+
+1. We can decouple the Connector SDK and Conduit release cycle from the schema
+   internal format release cycle
+2. We want to limit or add feature on top of internal the schema format
```
raulb, line 201:
```suggestion
2. We want to limit or add features on top of the internal schema format
```

docs/design-documents/20240430-schema-support.md
```diff
@@ -0,0 +1,420 @@
...
+#### Internal schema format
+
+##### Option 1: Avro
+
+We use Avro as the schema format used by the Connector SDK and internally.
+
+**Advantages**:
+
+1. Schema is a first-class citizen
+2. A widely used schema format.
```
raulb, line 220: I see inconsistencies with the use of `.` when enumerating items. Removing this one for consistency purposes.

```suggestion
2. A widely used schema format
```

docs/design-documents/20240430-schema-support.md
```diff
@@ -0,0 +1,420 @@
...
+3. A popular option with Kafka Connect (makes it easier for users to migrate)
+
+**Disadvantages**:
+
```
raulb, line 224: I would include something like: 

`No disadvantages were identified`.

docs/design-documents/20240430-schema-support.md
```diff
@@ -0,0 +1,420 @@
...
+Every field in a schema can be marked as optional (nullable). Alternatively,
+nullable fields can also be represented as a union of the `null` type and the
+field type. However, simply checking a boolean flag that a field is
+optional/nullable is better developer experience.
+
+## Schema operations
+
+The required schema operations are:
+
+1. create (using a name and list of fields)
```
raulb, line 113:
```suggestion
1. register (using a name and list of fields)
```

docs/design-documents/20240430-schema-support.md
raulb: Slightly unrelated, but I see that sometimes we refer as `create` and other times as `register` when it comes to the action of `creating` a schema. We should be consistent with this.

docs/design-documents/20240430-schema-support.md
```diff
@@ -0,0 +1,420 @@
...
+registry (Apicurio Registry), that is accessed by multiple Conduit
+instances.
+
+## Other considerations
+
+1. **Permissions**: creating a collection generally requires a broader set of
+   permissions then just writing data to a collection. For some users, the
+   benefit of having restricted permissions might outweigh the benefit of
+   auto-creating collections.
+2. **OpenCDC structured data**: it's becoming less useful, because it's types
```
raulb, line 418:
```suggestion
2. **OpenCDC structured data**: it's becoming less useful, because its types
```

docs/design-documents/20240430-schema-support.md
```diff
@@ -0,0 +1,420 @@
...
+
+The goal of schema support is to make the information about data types available
+in a pipeline. This is needed to cover a few use cases, such as automatically
+creating a destination collection or propagating schema changes.
+
+## Requirements
+
+1. Records **should not** carry the whole schema.
+
+   Reason: If a record would carry the whole schema, that might increase a
```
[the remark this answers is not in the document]
hariso: Actually, I started with `would`, and then Lyubo mentioned that it's not always the case, which I agree with.

docs/design-documents/20240430-schema-support.md
hariso: Hmm, let me try thinking of something, it can't be perfect.:)

docs/design-documents/20240430-schema-support.md
```diff
@@ -0,0 +1,416 @@
+<!-- markdownlint-disable MD004 MD007 MD033 -->
```
lovromazgon, line 1: Lol that's one way to deal with the linter 😄 I'd prefer if we follow the linter suggestions and change the content to ensure it follows the same style as other documents.

hariso: Yeah, I don't like it either, but the auto-generated TOC breaks these rules (no HTML, using asterisk, indentation) I should have added a comment explaining  the reason for the above.

lovromazgon: Can you update the TOC manually when you're done and don't anticipate any more changes? It's not like we'll touch this document again and will need to re-generate it.

hariso: Sounds like a good idea!

docs/design-documents/20240430-schema-support.md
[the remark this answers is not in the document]
lovromazgon: I don't see it as only the place for the services the connector exposes. It can contain anything related to the "protocol" that Conduit and the connector need to follow to communicate. This service sounds like it's part of that protocol.

docs/design-documents/20240430-schema-support.md
lovromazgon: I'm thinking of how will the destination connector figure out the schema evolved? If the schema could be identified using a name + version it could store in memory the last version it has seen for a schema name and then trigger some action if it sees a new version. Without versions, the connector would have no idea if the new schema in a record is actually an older one or a new one.

hariso: It looks like the schema version's ID should be enough for that, or maybe no?

docs/design-documents/20240430-schema-support.md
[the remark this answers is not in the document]
hariso: Gotcha. I was understanding as a "protocol" that needs to be followed when talking to a connector, and not the other way around too (to Conduit). In that case, then your suggestion definitely makes sense. :+1:

docs/design-documents/20240430-schema-support.md
hariso: The one that came after.:) A record that came after and has a different schema ID signals that its schema is a newer one.

lovromazgon: That's _probably_ the case, but might not always be the case, depending on how the source connector reads records. It could very well be that it gets some late-arriving records with the old schema. Or imagine two source connectors reading from independent sources, which have the same record schema. The schema gets updated in one source, but not the other, then records with different schema IDs might be interleaved for a while.

In any case, I don't want us making assumptions, rather give the destination the info it needs to make the right decision.

hariso: I believe it would be interesting to discuss such cases a bit deeper, but given that it's just one field (the version field) that's readily available, it's probably easiest just to expose it and, as you said, give the destination more info. :+1:

docs/design-documents/20240430-schema-support.md
```diff
@@ -0,0 +1,421 @@
...
+A schema consists of following:
+
+- reference: a string that uniquely identifies a schema in Conduit
+- list of fields, where each field is described with following:
+  - name
+  - type
+  - optional (boolean value)
+  - default value
+
+The following types are supported:
```
lovromazgon, line 80: The supported types really depend on the schema format, I'd remove this part to not mislead anyone. Maybe instead mention that multiple schema types will be supported, but we will start with Avro, as mentioned in the `Schema format` chapter.

lovromazgon (approved):
One last nitpick, LGTM 👍

docs/design-documents/20240430-schema-support.md
hariso: You're right about that. The point of this section was to mention what we consider are must-haves, i.e. requirements for a schema whatever the schema might be.

raulb: @hariso it's possible that all data serialization systems have those types,  but to make it super clear, I would mention that the types will depend on the system used. Maybe adding a link to avro's for reference since we're starting with that one https://avro.apache.org/docs/1.11.1/specification/#schema-declaration.

In a second thought, maybe we could mention those would be the types we'll come up with once we bring our own schema structure. In other words, the whole Schema structure will be **how it'll look like**, but https://avro.apache.org/docs/1.11.1/specification/#schema-declaration is what we're starting with.

docs/design-documents/20240430-schema-support.md
raulb: @hariso is this required for this release? Wondering if that can be tackled independently or should be part of this design document.

raulb (approved):
@hariso I only left one comment regarding Schema structure. Everything else looks great! Thanks for working on this.

raulb:
@hariso github is not showing accordingly what [this comment](https://github.com/ConduitIO/conduit/pull/1532#discussion_r1611469940) is about. For reference, it's in response to [your comment here](https://github.com/ConduitIO/conduit/pull/1532/files#r1598676155).

docs/design-documents/20240430-schema-support.md
hariso: I think I can do that as part of the SDK work, where I'll need that connection from a connector to Conduit.

docs/design-documents/20240430-schema-support.md
hariso: @raulb You're right about that, the types will depend on the actual schema format being used. I can add that to the doc.

Here's a bit of context: I wrote the requirements prior to us choosing the schema format, so we know what are the minimum requirements for a schema. For example, Lyubo asked for a timestamp type (for SQL connectors).

Issue #1548, opened 2024-05-09
Feature: Support adding headers for the webhook.http processor


maha-hajja:
### Feature description

the user should have the ability to add headers to the HTTP request, like Authorization

PR #1552, opened 2024-05-10
add headers to http processor

maha-hajja:
### Description

Fixes https://github.com/ConduitIO/conduit/issues/1548.

### Quick checks:

### DIFF
--- a/pkg/plugin/processor/builtin/impl/webhook/http.go
+++ b/pkg/plugin/processor/builtin/impl/webhook/http.go
@@ -47,7 +47,7 @@ type httpConfig struct {
 	// Deprecated: use `headers.Content-Type` instead.
 	ContentType string `json:"request.contentType" default:"application/json"`
 	// Headers to add to the request, use `headers.*` to specify the header and its value (e.g. `headers.Authorization: "Bearer key"`).
-	Headers map[string]string
+	Headers map[string]string `json:"headers"`
 
 	// Maximum number of retries for an individual record when backing off following an error.
 	BackoffRetryCount float64 `json:"backoffRetry.count" default:"0" validate:"gt=-1"`
--- a/pkg/plugin/processor/builtin/impl/webhook/http.go
+++ b/pkg/plugin/processor/builtin/impl/webhook/http.go
@@ -74,6 +74,27 @@ type httpConfig struct {
 	ResponseStatusRef string `json:"response.status"`
 }
 
+func (c *httpConfig) validateHeaders() error {
+	if c.Headers == nil {
+		c.Headers = make(map[string]string)
+	}
+
+	var headerValue string
+	for name, value := range c.Headers {
+		if strings.ToLower(name) == "content-type" {
+			headerValue = value
+		}
+	}
+
+	if headerValue != "" && c.ContentType != "" {
+		return cerrors.Errorf("Configuration error, cannot provide both \"request.contentType\" and \"headers.Content-Type\", use \"headers.Content-Type\" only.")
+	}
+
+	c.Headers["Content-Type"] = headerValue
+
+	return nil
+}
+
 type httpProcessor struct {
 	sdk.UnimplementedProcessor
 
@@ -110,7 +131,7 @@ func (p *httpProcessor) Configure(ctx context.Context, m map[string]string) erro
 		return cerrors.Errorf("failed parsing configuration: %w", err)
 	}
 
-	err = p.validateHeaders()
+	err = p.config.validateHeaders()
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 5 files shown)


pkg/plugin/processor/builtin/impl/webhook/http.go
```diff
@@ -156,6 +163,23 @@ func (p *httpProcessor) Configure(ctx context.Context, m map[string]string) erro
 	return nil
 }
 
+func (p *httpProcessor) validateHeaders() error {
+	if p.config.Headers != nil {
+		_, ok := p.config.Headers["Content-Type"]
```
hariso, line 168: Since this is user input, it might be a different case (lower, mixed, etc.)

pkg/plugin/processor/builtin/impl/webhook/http.go
```diff
@@ -44,8 +44,10 @@ type httpConfig struct {
 	URL string `json:"request.url" validate:"required"`
 	// Method is the HTTP request method to be used.
 	Method string `json:"request.method" default:"GET"`
-	// The value of the `Content-Type` header.
+	// Deprecated: use `headers.Content-Type` instead.
 	ContentType string `json:"request.contentType" default:"application/json"`
+	// Headers to add to the request, use headers.* to specify the key and set its value, ex: "headers.Authorization"="Bearer key".
+	Headers map[string]string
```
lovromazgon, line 50: I'd specify the `json` tag explicitly, just to be sure we don't rely on the implicit behavior.

```suggestion
	Headers map[string]string `json:"headers"`
```

pkg/plugin/processor/builtin/impl/webhook/http.go
lovromazgon: We should add tests to make sure we recognize different cases.

pkg/plugin/processor/builtin/impl/webhook/http.go
```diff
@@ -44,8 +44,10 @@ type httpConfig struct {
 	URL string `json:"request.url" validate:"required"`
 	// Method is the HTTP request method to be used.
 	Method string `json:"request.method" default:"GET"`
-	// The value of the `Content-Type` header.
+	// Deprecated: use `headers.Content-Type` instead.
 	ContentType string `json:"request.contentType" default:"application/json"`
+	// Headers to add to the request, use headers.* to specify the key and set its value, ex: "headers.Authorization"="Bearer key".
```
lovromazgon, line 49:
```suggestion
	// Headers to add to the request, use `headers.*` to specify the header and its value (e.g. `headers.Authorization: "Bearer key"`).
```

pkg/plugin/processor/builtin/impl/webhook/http.go
```diff
@@ -72,6 +74,27 @@ type httpConfig struct {
...
+		if strings.ToLower(name) == "content-type" {
+			headerValue = value
+		}
+	}
+
+	if headerValue != "" && c.ContentType != "" {
+		return cerrors.Errorf("Configuration error, cannot provide both \"request.contentType\" and \"headers.Content-Type\", use \"headers.Content-Type\" only.")
+	}
+
+	c.Headers["Content-Type"] = headerValue
```
lovromazgon, line 93: We still have 2 issues - we might be duplicating the `Content-Type` header if it uses a different case (e.g. `content-type`), and we are not using the deprecated `ContentType` option.

hariso: Yup, I'm aware of that, but I wasn't sure why the tests don't show that. Dumb thing, I was checking errors only, not the configs too.:)

pkg/plugin/processor/builtin/impl/webhook/http.go
```diff
@@ -72,6 +75,32 @@ type httpConfig struct {
...
+			hName = name
+		}
+	}
+
+	if hValue != "" && c.ContentType != "" {
+		return cerrors.Errorf("Configuration error, cannot provide both \"request.contentType\" and \"headers.Content-Type\", use \"headers.Content-Type\" only.")
+	}
+
+	delete(c.Headers, hName)
+	c.Headers["Content-Type"] = cmp.Or(hValue, c.ContentType)
```
lovromazgon, line 96: I'll say, there's an edge case, a bit exotic, but still. If I provide two headers, `content-type` and `Content-Type`, it's random if one or both headers will be sent by the processor, depending on how Go decides to traverse the `c.Headers` map.

My proposal, to make this completely safe:
```suggestion
	if c.ContentType == "" {
		return nil // Nothing to replace in headers
	}

	for name, _ := range c.Headers {
		if strings.ToLower(name) == "content-type" {
			return cerrors.Errorf("Configuration error, cannot provide both \"request.contentType\" and \"headers.Content-Type\", use \"headers.Content-Type\" only.")
		}
	}

	c.Headers["Content-Type"] = c.ContentType
```

lovromazgon (approved):
:shipit:

PR #1597, opened 2024-05-16
Webhook processor fixes

lovromazgon:
### Description

Followup to https://github.com/ConduitIO/conduit/pull/1552. The default from the deprecated field now conflicts with any `headers.content-type` option. This PR removes the default from the field and manually sets the default. It also updates an example to show how headers can be configured.

Additionally, it changes the behavior so a response status <500 is not considered a failing request.

### Quick checks:

- [X] I have followed the [Code Guidelines](https://github.com/ConduitIO/conduit/blob/main/docs/code_guidelines.md).
- [X] There is no other [pull request](https://github.com/ConduitIO/conduit/pulls) for the same update/change.
- [X] I have written unit tests.
- [X] I have made sure that the PR is of reasonable size and can be easily reviewed.

PR #1598, opened 2024-05-16
Error processor

lovromazgon:
### Description

Implements an `error` processor, that acts similar to the `filter` processor, only that it negatively acknowledges all records passed to it. This allows the user to stop the pipeline (or send records to the DLQ) if a record matches a certain condition.

I realized we need this when we updated the [webhook processor](https://github.com/ConduitIO/conduit/pull/1597) to not fail on HTTP statuses less than 500. If a user wants to still configure the pipeline to fail if the status code is for example 404, they can configure an error processor now, e.g.:

```yaml
version: 2.2
pipelines:
  - id: example
    status: running
    connectors:
      ...
    processors:
      - id: request
        plugin: webhook.http
        settings:
          response.status: '.Metadata["http_status"]'
      - id: fail-on-404
        plugin: error
        condition: `{{ eq .Metadata.http_status "404" }}`
```

> [!IMPORTANT]
> ~Depends on https://github.com/ConduitIO/conduit-processor-sdk/pull/46, validate action will fail until we merge that PR and update the processor-sdk dependency.~ Updated.

### Quick checks:


pkg/plugin/processor/builtin/impl/error.go
```diff
@@ -0,0 +1,101 @@
...
+
+func (p *errorProcessor) Specification() (sdk.Specification, error) {
+	return sdk.Specification{
+		Name:    "error",
+		Summary: "Returns an error for all records that get passed to the processor.",
+		Description: `Any time a record is passed to this processor it returns an error,
+which results in the record being sent to the DLQ if it's configured, or the pipeline stopping.
+
+**Important:** Make sure to add a [condition](https://conduit.io/docs/processors/conditions)
+to this processor, otherwise all records will provoke an error.`,
```
raulb, line 57:
```suggestion
to this processor, otherwise all records will trigger an error.`,
```

pkg/plugin/processor/builtin/internal/exampleutil/specs/error.json
```diff
@@ -0,0 +1,45 @@
+{
+  "specification": {
+    "name": "error",
+    "summary": "Returns an error for all records that get passed to the processor.",
+    "description": "Any time a record is passed to this processor it returns an error,\nwhich results in the record being sent to the DLQ if it's configured, or the pipeline stopping.\n\n**Important:** Make sure to add a [condition](https://conduit.io/docs/processors/conditions)\nto this processor, otherwise all records will provoke an error.",
```
raulb, line 5:
```suggestion
    "description": "Any time a record is passed to this processor it returns an error,\nwhich results in the record being sent to the DLQ if it's configured, or the pipeline stopping.\n\n**Important:** Make sure to add a [condition](https://conduit.io/docs/processors/conditions)\nto this processor, otherwise all records will trigger an error.",
```

raulb (approved):
LGTM! I like the idea of having an error processor, and it makes perfect sense to me. I left only a wording change suggestion.

@lovromazgon I think this could make also a good small blog post. I can help you with that if you want.

lovromazgon:
@raulb a blog post would be nice, feel free to take the initiative on that one, I'm kinda swamped right now.

Issue #1563, opened 2024-05-14
[Schemas] Limit schema service usage to connectors


hariso:
Part of #1560.

The schema service is currently intended to be used by connectors only. To limit the usage of the service to connectors only we should generate a magic cookie that a connector will use to identify itself (can be saved in an environment variable).

We should also run the schema service on a "random" port.

Pull requests:
- https://github.com/ConduitIO/conduit-connector-protocol/pull/112
- https://github.com/ConduitIO/conduit-connector-sdk/pull/150
- https://github.com/ConduitIO/conduit/pull/1701

raulb:
@hariso @lovromazgon apologies I couldn't bring this one home before I took some time off.

What I was planning on doing was adding to `pkg/plugin/schema_cookie.go` on conduit the following:

```go
package plugin

import (
	"fmt"
	"math/rand"
	"os"
	"strings"
	"time"

	"github.com/conduitio/conduit/pkg/plugin/connector"
)

const charset = "abcdefghijklmnopqrstuvwxyz" +
	"ABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789"

func randomString(length int) string {
	rand.New(rand.NewSource(time.Now().UnixNano()))
	sb := strings.Builder{}
	sb.Grow(length)
	for i := 0; i < length; i++ {
		sb.WriteByte(charset[rand.Intn(len(charset))])
	}
	return sb.String()
}

// GetSchemaCookieValue receives the name of the plugin and the version, and registers into the provided context a cookie that will be used later on to validate if this plugin is registered.
// This cookie will be stored in an environment variable
func GetSchemaCookieValue(pluginName, pluginVersion string) (connector.SchemaCookie, error) {
	name := strings.ToUpper(fmt.Sprintf("%s_%s", pluginName, pluginVersion))
	value := randomString(32)

	if err := registerCookie; err != nil {
		return connector.SchemaCookie{}, err
	}

	return connector.SchemaCookie{
		Name:  name,
		Value: value,
	}, nil

}

func registerCookie(cookie connector.SchemaCookie) error {
	return os.Setenv(cookie.Name, cookie.Value)
}
```

hariso:
@raulb No problems at all, you already did a lot and didn't leave us much work at all!

PR #1605, opened 2024-05-20
 [Schemas] In-memory schema service

### LINKED ISSUE #1561: [Schemas] In-memory schema service
Part of https://github.com/ConduitIO/conduit/issues/1560.

We need to implement a service that exposes methods to fetch and register schemas. The service will be used for other internal, connector-related, methods.  "Connector utility service" is just a working name, we can change to whatever we agree might be better.

It also needs to run on a randomly generated port (in the future this port might be made configurable). The port should be exposed through an environment variable so that connectors can use it.

In this issue, we can have an in-memory implementation, so that the Connector SDK work can be unblocked.

**Note**: We should keep in mind that connectors and processors communicate with Conduit differently.

hariso:
### Description

Closes #1561.  Depends on https://github.com/ConduitIO/conduit-connector-protocol/pull/88. 

### Quick checks:


pkg/conduit/runtime.go
```diff
@@ -155,13 +160,23 @@ func NewRuntime(cfg Config) (*Runtime, error) {
...
 	}
 	return r, nil
 }
 
+func newGRPCStatsHandler() *promgrpc.StatsHandler {
+	h := promgrpc.ServerStatsHandler()
+	promclient.MustRegister(h)
+
+	return h
+}
```
hariso, line 178: Previously, the stats handler was manually created because message stats were not needed (since we didn't use streams). We started using streams (between Conduit and connectors), so we can now use this stats handler.

pkg/connservices/doc.go
```diff
@@ -0,0 +1,18 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+/*
+Package connservices contains APIs meant to be used by connectors only.
+*/
+package connservices
```
lovromazgon, line 18: I would pull this into `plugin/connectors`, it's only relevant for connector plugins.

pkg/plugin/processor/builtin/impl/avro/schemaregistry/fake.go
```diff
@@ -94,10 +94,10 @@ const (
 	errorCodeSchemaNotFound  = 40403
 )
 
-// fakeRegistry is a simple fake registry meant to be used in tests. It stores
+// FakeRegistry is a simple fake registry meant to be used in tests. It stores
```
lovromazgon, line 97: I think this might also be more usfeul in conduit-commons, I expect we'll need it in tests in the connector and processor SDK.

pkg/conduit/runtime.go
lovromazgon: We don't really collect metrics from the plugin using the gRPC stats handler. Regardless, I'm not sure why we had a custom stats handler in the first place tbh, even if we don't have streams, it doesn't really hurt if we register those stream stats collectors 👍

pkg/connservices/proto_converter.go
```diff
@@ -0,0 +1,69 @@
...
+import (
+	"fmt"
+
+	schemav1 "github.com/conduitio/conduit-connector-protocol/proto/schema/v1"
+	"github.com/conduitio/conduit/pkg/foundation/cerrors"
+	"github.com/conduitio/conduit/pkg/schema"
+)
+
+type protoConverter struct {
+}
```
lovromazgon, line 26: I assume you took the `protoConverter` idea from the processors package. I propose to rather move these converters into the connector-protocol and use the same [`fromproto`](https://github.com/ConduitIO/conduit-connector-protocol/tree/main/cplugin/v2/fromproto) and [`toproto`](https://github.com/ConduitIO/conduit-connector-protocol/tree/main/cplugin/v2/toproto) pattern. A nice side-effect of this is that we won't be able to use any types from Conduit, we'll have to use types that are defined in the protocol or in conduit-commons, which will in turn make it possible to reuse them in connectors and processors.

pkg/schema/instance.go
```diff
@@ -0,0 +1,29 @@
...
+	TypeAvro Type = iota
+)
+
+type Instance struct {
+	ID      string
+	Name    string
+	Version int32
+	Type    Type
+	Bytes   []byte
+}
```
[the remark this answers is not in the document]
hariso: A standalone connector will get a Protobuf schema message. From it, we can take the bytes, get the Avro schema of it, and make it available. Similar is true for standalone processors.

pkg/plugin/processor/builtin/impl/avro/schemaregistry/fake.go
hariso: Hmm, but the Connector SDK for example should use a schema registry only through Conduit, i.e. it's Conduit's schema API which we can mock and use in the tests.

pkg/schema/instance.go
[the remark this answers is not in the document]
hariso: Is it going to be used across all of them though? A standalone connector will get a Protobuf schema message. From it, we can take the bytes, get the Avro schema of it, and make it available. Similar is true for standalone processors.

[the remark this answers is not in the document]
lovromazgon: The protobuf message will have to be transformed into a Go type like the one we have here, so the connector developer doesn't have to work with proto types. Sure, we _could_ have separate objects in all repos where we need it, but that's why we introduced conduit commons, so we don't duplicate the types and utilities around them.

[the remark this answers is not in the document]
hariso: You're right, we'll need a Go type for the schema. But that will be an Go type representing an Avro schema, like [this](https://pkg.go.dev/github.com/hamba/avro/v2#Schema), no?

pkg/plugin/processor/builtin/impl/avro/schemaregistry/fake.go
lovromazgon: Mocking the registry is a valid approach, but the registry is already a pretty dumb thing, when I was working on that `schemaregistry` package I found it much simpler to work with the fake registry (that's why I added it in the first place). Similarly how we use the in-memory DB in Conduit tests, I think it will be easier to write tests using a proper in-memory schema registry.

Although if you prefer mocking, I won't stand in your way 😉

lovromazgon: It will call the gRPC service, that's right, I'm saying that in tests we might want to forward the calls to this ~fake~ in-memory registry instead of a mock. In other words, instead of a mock we'd have a small adapter type, which implements the gRPC service interface that the connector is using and calls the in-memory registry.

pkg/schema/instance.go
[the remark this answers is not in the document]
hariso: Gotcha, and I was about to say that a connector developer probably doesn't need a nice Go type besides an Avro schema itself, but then I remember the schema version for example. So we'll definitely need one.

pkg/plugin/processor/builtin/impl/avro/schemaregistry/fake.go
hariso: I think you're making a good, but at least in this moment, I'm not quite convinced it's worth. So maybe we just leave the FakeRegistry, and move it into commons when/if we realize it's easier to test with it? That way, we'll hopefully keep the commons repo a bit cleaner.

pkg/plugin/processor/builtin/impl/avro/schemaregistry/fake.go
```diff
@@ -94,10 +94,10 @@ const (
...
 // schemas in memory and supports only the basic functionality needed in our
 // tests and supported by our client.
-type fakeRegistry struct {
+type FakeRegistry struct {
 	schemas            []sr.SubjectSchema
```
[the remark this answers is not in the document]
lovromazgon: So the idea is to completely rely on Apicurio? I was actually talking about changing this struct so it becomes an actual schema registry. All it's missing is just a way to permanently store the schemas, that's what I mean by giving it access to a [`database.DB`](https://github.com/ConduitIO/conduit/blob/bf18b8c4649f0c418ba5bc93f92cfbdacd2ae09a/pkg/foundation/database/database.go#L31) object and let it store the schemas there.

What worries me is that if we don't have a schema registry in here, then Conduit will have a hard dependency on Apicurio (well, depending on the connectors used in the pipeline) and won't be a completely standalone tool anymore, which is one of the biggest appeals of Conduit. I'm not saying that we shouldn't use an external registry, we should, that's how schemas can be shared across Conduit instances. But I'm trying to prevent the external registry becoming a hard dependency, things should ideally still work if there is none.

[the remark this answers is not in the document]
hariso: Leaving the no-op (in memory) implementation around for development purposes is a good idea, we can do that. :+1: 

As for storing the schemas in the Conduit DB, I think that too is a good idea, but we should discuss it as a separate item, so that expectations are correctly, especially given that we plan a Conduit schema registry long term anyway.

pkg/schema/service.go
```diff
@@ -0,0 +1,79 @@
...
+	}, nil
+}
+
+func (s *Service) Get(_ context.Context, id string) (schema.Instance, error) {
+	idInt, err := strconv.Atoi(id)
+	if err != nil {
+		return schema.Instance{}, cerrors.Errorf("invalid schema id: %w", err)
+	}
+
+	sch, found := s.fakeReg.SchemaByID(idInt)
```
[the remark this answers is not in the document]
hariso: Yup, however, in the Apicurio Registry, IDs are strings. If the caller doesn't supply one, they are UUIDs (we may or may not use that functionality). Record metadata values are strings too, so I believe that all in all it's better to leave them strings: it gives us more flexibility.

pkg/plugin/processor/builtin/impl/avro/schemaregistry/fake.go
```diff
@@ -94,10 +94,10 @@ const (
 	errorCodeSchemaNotFound  = 40403
 )
 
-// fakeRegistry is a simple fake registry meant to be used in tests. It stores
+// FakeRegistry is a simple fake registry meant to be used in tests. It stores
 // schemas in memory and supports only the basic functionality needed in our
 // tests and supported by our client.
-type fakeRegistry struct {
+type FakeRegistry struct {
```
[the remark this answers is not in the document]
hariso: > Now that I'm thinking about it, maybe we can get away without having a schema.Service, but just having a schema.Registry.

We might be able to do that... But I think we can have a clearer picture once we're done integrating with Apicurio.:)

pkg/plugin/processor/builtin/impl/avro/schemaregistry/fake.go
```diff
@@ -106,7 +106,7 @@ type fakeRegistry struct {
 	initOnce sync.Once
 }
 
-func (fr *fakeRegistry) init() {
+func (fr *FakeRegistry) Init() {
```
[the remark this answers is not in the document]
hariso: Moved to the ctor method.

pkg/schema/instance.go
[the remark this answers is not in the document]
hariso: Added to commons: https://github.com/ConduitIO/conduit-commons/commit/31b922714201dae61542657bf4364f125c33b314.

pkg/plugin/processor/builtin/impl/avro/decode_examples_test.go
```diff
@@ -12,8 +12,6 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-//go:build !integration
```
hariso, line 15: The fake registry now needs to be part of the build.

pkg/schema/fake_registry.go
```diff
@@ -12,9 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-//go:build !integration
```
lovromazgon, line 15: The fake schema registry was previously meant to only run in unit tests, now we include it in Conduit because we use it as the in-memory implementation. This all makes sense. However, previously _integration_ tests were running against a real schema registry, meaning that the same test was ensuring that it works in a real-world use case, while allowing the developer to run fast unit tests without spinning up a schema registry. Notice that [`integration.go`](https://github.com/ConduitIO/conduit/blob/main/pkg/plugin/processor/builtin/impl/avro/schemaregistry/integration.go) and [`fake.go`](https://github.com/ConduitIO/conduit/blob/main/pkg/plugin/processor/builtin/impl/avro/schemaregistry/fake.go) both define the functions `ExampleSchemaRegistryURL` and `TestSchemaRegistryURL`, so depending on the flag `--tags=integration` the test either targets the fake or a real schema registry. This behavior should be retained.

Essentially we need to extract everything between lines 31 and 88 into a separate file that's guarded with the `go:build` tag I'm commenting on, and then have another file like [`integration.go`](https://github.com/ConduitIO/conduit/blob/main/pkg/plugin/processor/builtin/impl/avro/schemaregistry/integration.go) that just returns the URL to the real schema registry.

pkg/schema/fake_registry.go
```diff
@@ -94,33 +91,29 @@ const (
 	errorCodeSchemaNotFound  = 40403
 )
 
-// fakeRegistry is a simple fake registry meant to be used in tests. It stores
+// FakeRegistry is a simple fake registry meant to be used in tests. It stores
 // schemas in memory and supports only the basic functionality needed in our
 // tests and supported by our client.
-type fakeRegistry struct {
+type FakeRegistry struct {
```
lovromazgon, line 97: I wonder if `FakeRegistry` is a suitable name now that we don't use it solely as a fake in tests but also as a real in-memory registry.

pkg/schema/in_memory.go
```diff
@@ -0,0 +1,71 @@
...
+func (s *InMemoryService) Create(_ context.Context, name string, bytes []byte) (schema.Instance, error) {
+	created := s.fakeReg.CreateSchema(name, sr.Schema{
+		Schema: string(bytes),
+		Type:   sr.TypeAvro,
+	})
+
+	return schema.Instance{
+		ID:      strconv.Itoa(created.ID),
+		Name:    "",
+		Version: 0,
```
lovromazgon, line 47:
```suggestion
		Name:    created.Subject,
		Version: created.Version,
```

pkg/schema/in_memory.go
```diff
@@ -0,0 +1,71 @@
...
+
+	sch, found := s.fakeReg.SchemaByID(idInt)
+	if !found {
+		return schema.Instance{}, ErrSchemaNotFound
+	}
+
+	return schema.Instance{
+		ID:      id,
+		Name:    "",
+		Version: 0,
```
lovromazgon, line 67:
```suggestion
		Name:    created.Subject,
		Version: created.Version,
```

pkg/schema/service.go
```diff
@@ -0,0 +1,35 @@
...
+// limitations under the License.
+
+package schema
+
+//go:generate mockgen -destination=mock/schema_service.go -package=mock -mock_names=Service=Service . Service
+
+import (
+	"context"
+
+	commschema "github.com/conduitio/conduit-commons/schema"
```
lovromazgon, line 22: Why are we renaming this to `commschema`? Other files don't do that 🤔

Btw, given the many package name clashes between conduit-commons, protocol, and conduit, did we consider a different naming approach? After thinking about this for a while, this is my proposal:
- Rename this package to `schemaregistry`. I know, not ideal, but it removes the naming conflict.
- The conduit-commons package stays as it is, simply `schema`.
- For the protocol I would introduce the naming convention of prefixing packages with `p` as in "protocol", so the package would become `pschema`. Similarly, we would rename `cplugin` to `pconnector`, to indicate it contains the protocol code for the `connector` proto files.

WDYT? If we do this, I can include the `cplugin` -> `pconnector` change in my PRs.

pkg/schema/service.go
```diff
@@ -0,0 +1,35 @@
...
+	"github.com/conduitio/conduit/pkg/foundation/cerrors"
+)
+
+var (
+	ErrSchemaNotFound = cerrors.New("schema not found")
+)
+
+type Service interface {
```
lovromazgon, line 30: Not for this PR, but I'll mention it before I forget. I realized that having a builtin schema registry could create the expectation that the [`avro.decode`](https://conduit.io/docs/processors/builtin/avro.decode) and [`avro.encode`](https://conduit.io/docs/processors/builtin/avro.encode) processors would be able to use the internal schema registry. It would be a bit strange if we had to configure Conduit to point to a schema registry and then again supply the URL to the processors separately. So ideally the processors should use this service to create and get schemas. Not sure if this was included in the design or not.

pkg/plugin/processor/builtin/impl/avro/schemaregistry/internal/cache.go
```diff
@@ -32,7 +33,7 @@ func newSubjectVersion(subject string, version int) subjectVersion {
 }
 
 func newSubjectFingerprint(subject string, text string) subjectFingerprint {
-	fingerprint := Rabin([]byte(text))
+	fingerprint := schema.Rabin([]byte(text))
```
lovromazgon, line 36: This is a bit of a smell if you ask me, the `Rabin` fingeprint is an implementation detail, that's why it was in the `internal` package before. Seems to me that this cache belongs in the `schema` package. I won't dwell on it though, we can refactor it in a PR down the road when we have a better picture what goes where.

pkg/schema/fake_registry.go
hariso: Things are easy to miss with the build tags.:)

pkg/schema/fake_registry.go
hariso: Good point, I'll rename it!

pkg/schema/service.go
hariso: I created an issue for this so we can discuss it there: https://github.com/ConduitIO/conduit/issues/1653.

pkg/plugin/processor/builtin/impl/avro/schemaregistry/internal/cache.go
hariso: I'll add a todo for us here, so we don't forget about it before merging the feature branch to `main`.

pkg/schema/service.go
hariso: > Why are we renaming this to `commschema`? Other files don't do that 🤔

`cschema` didn't read accurate to me.:) So I went with `commschema` to try indicating it comes from `commons`.

As for the name changes: yeah, it does look like we're bumping into this more frequently. I would suggest we rename it once this and Raul's PR are merged, so we don't derail the current work too much.

As for the names themselves: how about we add a `c` prefix to the `commons` packages?

pkg/plugin/processor/builtin/impl/avro/schemaregistry/internal/cache.go
hariso: FTR, Raul will fix this in his PR.

pkg/schema/fake_registry.go
hariso: Done!

pkg/schema/service.go
raulb: For context, I'm making some pkg changes in https://github.com/ConduitIO/conduit/pull/1630

pkg/plugin/processor/builtin/impl/avro/schemaregistry/internal/cache.go
raulb: Yup https://github.com/ConduitIO/conduit/pull/1630

pkg/schema/service.go
lovromazgon: > As for the names themselves: how about we add a `c` prefix to the `commons` packages?

I wouldn't do that for all packages in commons. We do it for _some_ already (`csync`, `cchan`, `ccontext`) where we anticipate conflicts, but I wouldn't make it a rule for all. I'd also keep `schema` named `schema` in commons, as it can be used by processors and connectors as well, so we expose a nice API for them.

hariso: Sounds good to me. I can make the changes once this PR has been merged and once Raul merges his changes too.

pkg/conduit/runtime.go
```diff
@@ -461,7 +469,38 @@ func (r *Runtime) serveGRPCAPI(ctx context.Context, t *tomb.Tomb) (net.Addr, err
...
+	healthServer := api.NewHealthServer(
+		map[string]api.Checker{
+			"SchemaService": r.schemaService,
+		},
+		r.logger,
+	)
+	grpc_health_v1.RegisterHealthServer(grpcServer, healthServer)
+
+	// todo make port random
+	return r.serveGRPC(ctx, t, grpcServer, ":8184")
```
lovromazgon, line 503: Hint: `":0"` will choose a random free port. Or if we want to give the user some control over this, we could do the same as go-plugin and let the user define [min and max ports](https://github.com/hashicorp/go-plugin/blob/721087dad1e61cd41a19a30e5ae740dceb55061d/client.go#L188-L191).

pkg/schema/service.go
```diff
@@ -0,0 +1,35 @@
...
+import (
+	"context"
+
+	commschema "github.com/conduitio/conduit-commons/schema"
+	"github.com/conduitio/conduit/pkg/foundation/cerrors"
+)
+
+var (
+	ErrSchemaNotFound = cerrors.New("schema not found")
+)
```
lovromazgon, line 28: I'd move this to `errors.go`, we use this in quite a few packages in Conduit, makes it easier to discover them.

lovromazgon (approved):
A few nitpicks, but approving to unblock this and move on 👍

pkg/schema/in_memory.go
```diff
@@ -44,7 +44,7 @@ func (s *InMemoryService) Create(_ context.Context, name string, bytes []byte) (
 	return schema.Instance{
 		ID:      strconv.Itoa(created.ID),
 		Name:    created.Subject,
-		Version: created.Version,
+		Version: int32(created.Version),
```
lovromazgon, line 47: Btw, why do we use `int32` in `schema.Instance`? I anticipate that it will cause a bunch of casts in our code, also given that we'll expose this to the connector and processor this could simply be an `int` to make it easier to work with.

pkg/schema/service.go
lovromazgon: Alright, I'll add the rename of `cplugin` -> `pconnector` after I'm done with my current work 👍

pkg/schema/in_memory.go
lovromazgon: I just noticed you changed it to `int` in https://github.com/ConduitIO/conduit-commons/pull/67 👍

pkg/conduit/runtime.go
hariso: Yup, yup, the port stuff is going to be handled in #1563.

Issue #1433, opened 2024-03-12
Feature: Update conduit internals to use conduit-commons


simonl2002:
### Feature description

This work includes removing config parameters and validations in the connector SDK in favor of using the config parameters in conduit-commons (see https://github.com/ConduitIO/conduit-commons/pull/18 and https://github.com/ConduitIO/conduit-commons/pull/19). This allows us to reuse the same code in connectors and processors. Also paramgen can then be removed from the SDK in favor of paramgen in conduit-commons, which will then work for both connectors and processors.

Additionally we need to remove `sdk.Record` (connector SDK) and `record.Record` (Conduit) in favor of `opencdc.Record` in conduit-commons. This will simplify our codebase as well as have a positive impact on performance, since we won't have to do as many transformations between the types, which is currently happening in the hot path (for every record).

lovromazgon:
Migration script for connectors regarding the changes done in this issue:

```sh
sed -i '' 's/conduit-connector-sdk\/cmd\/paramgen/conduit-commons\/paramgen/g' tools.go
find . -type f -iname "*.go" -exec sed -i '' 's/sdk.Record/opencdc.Record/g' {} +
find . -type f -iname "*.go" -exec sed -i '' 's/sdk.Operation/opencdc.Operation/g' {} +
find . -type f -iname "*.go" -exec sed -i '' 's/sdk.Position/opencdc.Position/g' {} +
find . -type f -iname "*.go" -exec sed -i '' 's/sdk.Change/opencdc.Change/g' {} +
find . -type f -iname "*.go" -exec sed -i '' 's/sdk.Data/opencdc.Data/g' {} +
find . -type f -iname "*.go" -exec sed -i '' 's/sdk.RawData/opencdc.RawData/g' {} +
find . -type f -iname "*.go" -exec sed -i '' 's/sdk.StructuredData/opencdc.StructuredData/g' {} +
find . -type f -iname "*.go" -exec sed -i '' 's/sdk.Metadata/opencdc.Metadata/g' {} +
find . -type f -iname "*.go" -exec sed -i '' 's/sdk.Parameter/config.Parameter/g' {} +
find . -type f -iname "*.go" -exec sed -i '' 's/sdk.Validation/config.Validation/g' {} +
find . -type f -iname "*.go" -exec sed -i '' 's/Parameters() map\[string\]config.Parameter/Parameters() config.Parameters/g' {} +
[...]

PR #1622, opened 2024-05-28
Use connector protocol v2

lovromazgon:
### Description

This PR includes code changes to use `opencdc.Record` everywhere. It removes the old `record` package entirely. It also updates the connector protocol and SDK to the latest version, so we use the same record across the board. The code is updated according to the changes that were already done in the protocol.

Most of the code changes are straightforward, it's just that there are a lot. I posted a few comments below which call out the important bits, **I strongly recommend reading them through before reviewing this PR**.

The next PR will remove a bunch of code in `pkg/foundation` and use the packages in `conduit-commons` instead.

Closes https://github.com/ConduitIO/conduit/issues/1433.

### Quick checks:


pkg/plugin/connector/builtin/source.go
```diff
@@ -0,0 +1,127 @@
...
+}
+
+func (s *sourcePluginAdapter) withLogger(ctx context.Context) context.Context {
+	return s.ctxLogger.WithContext(ctx)
+}
+
+func (s *sourcePluginAdapter) Configure(ctx context.Context, in pconnector.SourceConfigureRequest) (pconnector.SourceConfigureResponse, error) {
+	s.logger.Debug(ctx).Any("request", in).Msg("calling Configure")
+	out, err := runSandbox(s.impl.Configure, s.withLogger(ctx), in.Clone(), s.logger)
+	return out.Clone(), err
```
lovromazgon, line 59: Also all methods that call directly into the builtin plugin clone the request and response.

pkg/plugin/connector/acceptance_testing.go
```diff
@@ -319,68 +264,74 @@ func testSource_Read_Success(t *testing.T, tdf testDispenserFunc) {
 	source, err := dispenser.DispenseSource()
 	is.NoErr(err)
 
-	err = source.Start(ctx, nil)
+	stream := source.NewStream()
```
lovromazgon, line 267: All tests (and generally all code) that were using `Start` were changed as they need to create the stream and use it directly. We have less abstraction now.

pkg/pipeline/lifecycle_test.go
```diff
@@ -175,13 +176,20 @@ func TestServiceLifecycle_PipelineSuccess(t *testing.T) {
...
 	is.Equal("", pl.Error)
+
+	// stop pipeline before ending test
+	err = ps.Stop(ctx, pl.ID, false)
+	is.NoErr(err)
+	is.NoErr(pl.Wait())
 }
 
 func TestServiceLifecycle_PipelineError(t *testing.T) {
+	t.Skipf("this test fails, see github.com/ConduitIO/conduit/issues/1659")
```
lovromazgon, line 187: This is the test mentioned in https://github.com/ConduitIO/conduit/issues/1659. It's not really a regression, the previous test just had a misconfigured mock. Now that we have less abstraction the tests are closer to reality and thus can find this issue.

pkg/pipeline/lifecycle_test.go
```diff
@@ -414,34 +435,23 @@ func generateRecords(count int) []record.Record {
...
-			return record.Record{}, ctx.Err()
-		}
-		r := records[position]
-		position++
-		return r, nil
-	}).MinTimes(recordCount + 1)
+	if stop {
+		sourcePluginOptions = append(sourcePluginOptions, pmock.SourcePluginWithStop())
+	}
+	sourcePlugin := pmock.NewConfigurableSourcePlugin(ctrl, sourcePluginOptions...)
```
lovromazgon, line 454: This is where we use `ConfigurableSourcePlugin` and `ConfigurableDestinationPlugin` mentioned in the other comment.

pkg/connector/destination.go
```diff
@@ -18,11 +18,12 @@ import (
...
+	"github.com/conduitio/conduit-commons/opencdc"
+	"github.com/conduitio/conduit-connector-protocol/pconnector"
 	"github.com/conduitio/conduit/pkg/foundation/cerrors"
 	"github.com/conduitio/conduit/pkg/foundation/log"
 	"github.com/conduitio/conduit/pkg/plugin"
 	connectorPlugin "github.com/conduitio/conduit/pkg/plugin/connector"
-	"github.com/conduitio/conduit/pkg/record"
 )
 
 type Destination struct {
```
lovromazgon, line 29: The code in `connector.Destination` and `connector.Source` needs a bit more attention than the rest. We removed an abstraction layer in `pkg/plugin/connector/...`, this code now does just a bit more, as it communicates more directly with the plugin.

pkg/pipeline/stream/destination.go
```diff
@@ -104,7 +105,7 @@ func (n *DestinationNode) Run(ctx context.Context) (err error) {
 		n.logger.Trace(msg.Ctx).Msg("writing record to destination connector")
 
 		writeTime := time.Now()
-		err = n.Destination.Write(msg.Ctx, msg.Record)
+		err = n.Destination.Write(msg.Ctx, []opencdc.Record{msg.Record})
```
lovromazgon, line 108: Notice that the plugin protocol now supports batching, but the node is not taking advantage of that yet.

pkg/pipeline/stream/base.go
```diff
@@ -173,7 +173,9 @@ func (n *pubNodeBase) Trigger(
 						}
 						return
 					}
-					n.msgChan <- msg
+					for _, msg := range msgs {
+						n.msgChan <- msg
+					}
```
lovromazgon, line 178: Same here, the source node breaks up the batch into separate messages.

pkg/plugin/connector/builtin/registry.go
```diff
@@ -19,31 +19,27 @@ import (
...
 	// built-in plugins. The key of the map is the import path of the module
 	// containing the connector implementation.
 	DefaultDispenserFactories = map[string]DispenserFactory{
-		"github.com/conduitio/conduit-connector-file":      NewDispenserFactory(file.Connector),
-		"github.com/conduitio/conduit-connector-kafka":     NewDispenserFactory(kafka.Connector),
-		"github.com/conduitio/conduit-connector-generator": NewDispenserFactory(generator.Connector),
-		"github.com/conduitio/conduit-connector-s3":        NewDispenserFactory(s3.Connector),
-		"github.com/conduitio/conduit-connector-postgres":  NewDispenserFactory(postgres.Connector),
-		"github.com/conduitio/conduit-connector-log":       NewDispenserFactory(connLog.Connector),
+		// TODO update connectors to latest SDK and uncomment
```
lovromazgon, line 36: The rest of the plugins will be updated shortly.

lovromazgon (commented):
Here's some pointers to make the review easier.

pkg/foundation/database/database.go
```diff
@@ -12,7 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-//go:generate mockgen -destination=mock/database.go -package=mock -mock_names=DB=DB . DB
+//go:generate mockgen -typed -destination=mock/database.go -package=mock -mock_names=DB=DB . DB
```
lovromazgon, line 15: I changed all `mockgen` invocations to include the `-typed` flag introduced in mockgen. This makes our tests a bit safer and easier to write, since autocomplete helps you out now.

This is also where the bulk of the additional code comes from. Without this change we would lose quite a bit of lines because of the removal of `pkg/record` and the simplification of the plugin packages.

hariso: Was there a specific reason why we didn't use typed mocks before?

pkg/plugin/connector/builtin/source.go
hariso: What's the issue what cloning solves? We should probably leave a comment somewhere about that.

pkg/foundation/database/database.go
lovromazgon: This was added in the `uber-go/mock` fork of the library relatively recently (september last year). We should use it going forward.

pkg/plugin/connector/builtin/source.go
lovromazgon: We wouldn't have to clone _all_ of them, as some are empty. The point is that we don't want to share memory between Conduit and the plugin so changes are isolated. For instance, the plugin could manipulate the config map, but we don't want that to change the map in Conduit (that actually happens, when the plugin sets the defaults). In the end, this ensures the plugin works the same way as standalone or builtin.

lovromazgon: Added a comment https://github.com/ConduitIO/conduit/pull/1622/commits/b708694acd1ee6d25962b5dd3d0e8228c71153e7 👍

pkg/connector/destination.go
```diff
@@ -248,13 +279,37 @@ func (d *Destination) preparePluginCall() (func(), error) {
...
+		return cerrors.Errorf("could not open destination connector plugin: %w", err)
+	}
+	return nil
+}
+
+func (d *Destination) run(ctx context.Context) error {
+	d.Instance.logger.Trace(ctx).Msg("running destination connector plugin")
+	ctx, d.stopStream = context.WithCancel(ctx)
+	stream := d.plugin.NewStream()
+	err := d.plugin.Run(ctx, stream)
```
hariso, line 302: Can the `Run` method return the stream?

pkg/connector/source.go
```diff
@@ -282,13 +288,39 @@ func (s *Source) state() SourceState {
...
+
+func (s *Source) run(ctx context.Context) error {
+	s.Instance.logger.Trace(ctx).Msg("running source connector plugin")
+	ctx, s.stopStream = context.WithCancel(ctx)
+	stream := s.plugin.NewStream()
+	err := s.plugin.Run(ctx, stream)
+	if err != nil {
+		s.stopStream()
+		s.stopStream = nil
+		s.stream = nil
```
hariso, line 317: Hmm, in which cases would `s.stream` be non-nil? Only if `run()` was already called?

pkg/pipeline/stream/destination_acker.go
```diff
@@ -131,22 +131,22 @@ func (n *DestinationAckerNode) worker(
...
-			}
-
-			err = n.handleAck(msg, err)
+			acks, err := n.Destination.Ack(ctx)
 			if err != nil {
-				errChan <- err
+				handleError(msg, cerrors.Errorf("error while fetching acks: %w", err))
 				return
 			}
+			for _, ack := range acks {
```
hariso, line 139: Only 1 ack is expected here I guess?

pkg/plugin/connector/builtin/destination.go
```diff
@@ -0,0 +1,129 @@
...
+	"fmt"
+	"io"
+
+	"github.com/conduitio/conduit-connector-protocol/pconnector"
+	"github.com/conduitio/conduit/pkg/foundation/log"
+	"github.com/conduitio/conduit/pkg/plugin/connector"
+	"github.com/rs/zerolog"
+)
+
+// destinationPluginAdapter implements the destination plugin interface used
```
hariso, line 28: Nitpick really: the destination plugin -> DestinationPlugin (or with the path) to make it easier to find.

pkg/plugin/connector/builtin/destination.go
```diff
@@ -0,0 +1,129 @@
...
+	return out.Clone(), err
+}
+
+func (s *destinationPluginAdapter) Open(ctx context.Context, in pconnector.DestinationOpenRequest) (pconnector.DestinationOpenResponse, error) {
+	s.logger.Debug(ctx).Any("request", in).Msg("calling Open")
+	out, err := runSandbox(s.impl.Open, s.withLogger(ctx), in.Clone(), s.logger)
+	return out.Clone(), err
+}
+
+func (s *destinationPluginAdapter) Run(ctx context.Context, stream pconnector.DestinationRunStream) error {
```
hariso, line 70: Nitpick: `d` would be a better receiver name.

pkg/plugin/connector/builtin/destination.go
```diff
@@ -0,0 +1,129 @@
...
+	}
+
+	inmemStream.Init(ctx)
+
+	s.logger.Debug(ctx).Msg("calling Run")
+	go func() {
+		err := runSandboxNoResp(s.impl.Run, s.withLogger(ctx), stream, s.logger)
+		if err != nil {
+			if inmemStream.Close(err) {
+				s.logger.Err(ctx, err).Msg("stream already stopped")
```
hariso, line 86: It looks like `inmemStream.Close(err)` will return `true` if the stream was closed in that call. So from that point, we should either change the condition or change the log message.

pkg/connector/destination.go
lovromazgon: That would be great 😄 Although then we would need that abstraction layer (the one we removed now). The idea behind this change is that the interface is _exactly_ the same as what the connector implements (well, the connector SDK), so it's hopefully simpler to reason about. The connector needs to get a stream in `Run`, while Conduit needs a stream after it calles `Run`. This is now the middle ground, Conduit creates the stream and gives it to the connector, then both need to fetch the right "side" of the stream (conduit gets the client, the connector gets the server).

pkg/connector/source.go
lovromazgon: Good point, this is redundant.

pkg/pipeline/stream/destination_acker.go
lovromazgon: For now, this will be a single ack, yes. But this specific code should already work with multiple acks.

pkg/plugin/connector/builtin/destination.go
lovromazgon: copy pasta 🍝

pkg/pipeline/dlq.go
```diff
@@ -42,37 +41,28 @@ func (d *DLQDestination) Open(ctx context.Context) error {
...
 	if err != nil {
 		return err
 	}
-	if !bytes.Equal(rec.Position, pos) {
-		return cerrors.Errorf("received unexpected ack, expected position %q but got %q", rec.Position, pos)
+
+	if len(ack) != 1 {
+		return cerrors.Errorf("expected 1 ack but got %d", len(ack))
+	}
+
```
raulb, line 60:
```suggestion
```

pkg/pipeline/dlq.go
```diff
@@ -42,37 +41,28 @@ func (d *DLQDestination) Open(ctx context.Context) error {
...
 	d.lastPosition = rec.Position
 
-	err = wg.Wait(ctx)
+	ack, err := d.Destination.Ack(ctx)
 	if err != nil {
 		return err
 	}
-	if !bytes.Equal(rec.Position, pos) {
-		return cerrors.Errorf("received unexpected ack, expected position %q but got %q", rec.Position, pos)
+
```
raulb, line 56:
```suggestion
```

pkg/plugin/connector/builtin/stream.go
```diff
@@ -0,0 +1,181 @@
...
+	"sync"
+
+	"github.com/conduitio/conduit-connector-protocol/pconnector"
+)
+
+type InMemoryDestinationRunStream struct {
+	stream *inMemoryStream[pconnector.DestinationRunRequest, pconnector.DestinationRunResponse]
+}
+
+func (s *InMemoryDestinationRunStream) Init(ctx context.Context) {
```
raulb, line 29: Small nit. What do you think about naming these receivers? `im` instead?

raulb (approved):
@lovromazgon Looks great. Only left some nits

pkg/plugin/connector/builtin/destination.go
```diff
@@ -0,0 +1,129 @@
...
+		} else {
+			inmemStream.Close(io.EOF)
+		}
+		d.logger.Debug(ctx).Msg("Run stopped")
+	}()
+
+	return nil
+}
+
+func (d *destinationPluginAdapter) Stop(ctx context.Context, in pconnector.DestinationStopRequest) (pconnector.DestinationStopResponse, error) {
```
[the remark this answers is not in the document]
hariso: Sometimes we use `in` and `out` for what comes into the method (arguments, requests) and for what goes out (results, responses).

hariso (approved):

pkg/connector/destination.go
```diff
@@ -175,30 +190,46 @@ func (d *Destination) Teardown(ctx context.Context) error {
...
+	err = d.stream.Send(pconnector.DestinationRunRequest{Records: r})
 	if err != nil {
 		return cerrors.Errorf("error writing record: %w", err)
 	}
 
 	return nil
 }
 
-func (d *Destination) Ack(ctx context.Context) (record.Position, error) {
+func (d *Destination) Ack(context.Context) ([]DestinationAck, error) {
```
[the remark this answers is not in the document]
lovromazgon: Yes, you can skip the variable names of all parameters if none of them are needed. Useful when you just need a parameter to satisfy an interface.

pkg/plugin/connector/builtin/stream.go
lovromazgon: I don't feel strongly one way or the other. I can give you the reasoning why I used `s` - it's a stream 😉

Issue #1562, opened 2024-05-14
[Schemas] Apicurio Registry as the backend implementation


hariso:
Part of #1560.

This issue is a continuation of #1561. 

Here we change the connector utility service to use an Apicurio Registry instance to manage schemas. In `pkg/plugin/processor/builtin/impl/avro/schemaregistry/client.go` we have a caching client that can be used for this service.

We should also add a Docker compose file to install Apicurio Registry for local development purposes.

hariso:
Done in #1630

PR #1630, opened 2024-05-31
feat(schema): Add Apicurio backend

raulb:
### Description

Fixes https://github.com/ConduitIO/conduit/issues/1562

### Quick checks:


pkg/conduit/config.go
```diff
@@ -21,13 +21,17 @@ import (
 	"github.com/conduitio/conduit/pkg/foundation/database"
 	"github.com/conduitio/conduit/pkg/foundation/log"
 	"github.com/conduitio/conduit/pkg/plugin/connector/builtin"
+	"github.com/conduitio/conduit/pkg/schema"
```
raulb, line 24: Oops!

pkg/conduit/config.go
```diff
@@ -21,13 +21,17 @@ import (
...
 	"github.com/rs/zerolog"
 )
 
 const (
 	DBTypeBadger   = "badger"
 	DBTypePostgres = "postgres"
 	DBTypeInMemory = "inmemory"
+
+	SchemaTypeApicurio = "apicurio"
```
lovromazgon, line 33: We shouldn't align ourselves directly with apicurio, rather call this `external` or even `confluent`, as it needs to be a schema registry compatible with the confluent schema registry API.

test/docker-compose-apicurio.yml
```diff
@@ -0,0 +1,33 @@
+# docker compose -f test/docker-compose-apicurio.yml up --quiet-pull -d --wait
+# docker compose -f test/docker-compose-apicurio.yml down
+version: "3"
+
+services:
+  registry:
+    image: apicurio/apicurio-registry:latest-snapshot
```
lovromazgon, line 7: Btw, we already have a compose file with the confluent schema registry that we can use in tests.

pkg/conduit/config.go
raulb: @lovromazgon good point. I wasn't aware of such compatibility. I'll rename

test/docker-compose-apicurio.yml
raulb: Right. As I mentioned above, I wasn't familiar with that compatibility. In that case, it definitely makes sense to rely on that one instead. I'll remove. Thanks for the callout.

pkg/conduit/config_test.go
```diff
@@ -160,6 +160,7 @@ func TestConfig_Validate(t *testing.T) {
 			validConfig.Log.Level = "info"
 			validConfig.Log.Format = "cli"
 			validConfig.Pipelines.Path = "./pipelines"
+			validConfig.Schema.Type = SchemaTypeInMemory
```
hariso, line 163: I _think_ the above lines can be replaced with:
```go
validConfig := DefaultConfig()
validConfig.DB.Postgres.ConnectionString = "postgres://user:pass@localhost:5432/mydb?sslmode=disable"
```

raulb: @hariso good point! Much easier to maintain.

pkg/schemaregistry/confluent.go
```diff
@@ -0,0 +1,107 @@
...
+		return schema.Instance{}, err
+	}
+
+	return schema.Instance{
+		ID:    id,
+		Bytes: []byte(s.Schema),
+	}, nil
+}
+
+func (a *ConfluentService) Check(ctx context.Context) error {
```
hariso, line 85: Nanonitpick: the receiver name could be something like `s`.

pkg/schemaregistry/confluent.go
```diff
@@ -0,0 +1,107 @@
...
+	url := fmt.Sprintf("%s%s", a.connString, a.healthCheckPath)
```
hariso, line 86: I'm thinking, if let's say the connection string ends with a slash, and the health check path starts with a slash (or none of them have slashes), that could become a trouble.

hariso (approved):
Great work, @raulb! Let me think what you think about building the health check URL. All other comments are purely nitpicks.

pkg/conduit/runtime.go
```diff
@@ -146,6 +146,22 @@ func NewRuntime(cfg Config) (*Runtime, error) {
 
 	orc := orchestrator.NewOrchestrator(db, logger, plService, connService, procService, connPluginService, procPluginService)
 
+	var schemaService schemaregistry.Service
+	schemaService = cfg.Schema.Service
```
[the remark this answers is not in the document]
raulb: @hariso I wondered about the same. I felt this could be useful, like we do for DB being able to replace the driver altogether, but I could leave that one out for now.

pkg/schemaregistry/confluent.go
raulb: @hariso yeah, good point. Let me improve that

raulb: @hariso let me know what you think https://github.com/ConduitIO/conduit/pull/1630/commits/36bbc2e31ba5ea74192862cc9ad6ee9a84337a3f.

pkg/schemaregistry/confluent_test.go
```diff
@@ -0,0 +1,127 @@
...
+	bytes := []byte(`{"type":"record","name":"test","fields":[{"name":"field1","type":"string"}]}`)
+
+	instance, err := service.Create(ctx, expectedName, bytes)
+	assert.NoError(t, err)
+
+	gotInstance, err := service.Get(ctx, instance.ID)
+	assert.NoError(t, err)
+	assert.Equalf(t, gotInstance.ID, instance.ID, "expected id %s, got %s", instance.ID, gotInstance.ID)
+	assert.Emptyf(t, gotInstance.Name, "expected empty name, got %s", gotInstance.Name)
+	assert.Equalf(t, gotInstance.Bytes, instance.Bytes, "expected bytes %s, got %s", instance.Bytes, gotInstance.Bytes)
```
hariso, line 75: It's great to have these checks! One small thing, there's a convention that we stick to github.com/matryer/is, since it's a simple and, together with go-cmp it handles pretty much all cases.

PR #1658, opened 2024-06-14
Configure dependabot for Docker

hariso:
### Description

This covers our Dockerfile for the Conduit Docker image. Updates for it (Go, Alpine) are relatively infrequent so I've put the schedule interval to "weekly".

This doesn't cover the Docker Compose files we have, as dependabot doesn't support them: https://github.com/dependabot/dependabot-core/issues/390

A solution that's been often mentioned in the above issue is [Renovate](https://github.com/renovatebot/renovate).

### Quick checks:


.github/dependabot.yml
```diff
@@ -25,3 +25,11 @@ updates:
...
     commit-message:
       prefix: "npm:"
+
+  # Maintain dependencies for GitHub Actions
+  - package-ecosystem: "docker"
+    directory: "/"
+    schedule:
+      interval: "weekly"
+    commit-message:
+      prefix: ".docker:"
```
lovromazgon, line 35: Nitpick re prefix, how about just `docker`?
```suggestion
      prefix: "docker:"
```

Btw, we should add an exclusion for this to [goreleaser](https://github.com/ConduitIO/conduit/blob/3552ad5c2e03c63f23bf1e031a3984b12aa611c5/.goreleaser.yml#L41-L46).

lovromazgon: I'm not exactly sure where we got that from, I assume that `.github` points towards updating the contents of the `.github` folder in the root of the repo 🤷 I wouldn't mind if that would simply be `github` tbh, but I don't think it's worth changing now.

lovromazgon (approved):

PR #1678, opened 2024-06-27
support empty .payload.After / .payload.Before in avro.decode

samirketema:
### Description

The avro decode processor states the following:
> If a specific field is set to nil the processor won't have enough information to determine the type and will default to a nullable string.

In the case that the entire `.payload.after` is nil, the processor fails with the following error:
```
failed getting structured data: failed unmarshalling JSON from raw data: expected {} for map
```

This attempts to alleviate the issue, by instead returning an empty `sdk.StructuredData` in this scenario.

Fixes # (issue)

### Quick checks:


raulb (commented):
Makes sense to me!

raulb (approved):

lovromazgon (approved):

PR #1682, opened 2024-06-28
Database implementation for SQLite3

lyuboxa:
### Description

Uses non-ego driver, only sets some default pragmas.
* Ensure journal mode is set to WAL
* Add options to CLI/Config

### Quick checks:


pkg/foundation/database/acceptance_testing.go
```diff
@@ -184,6 +187,79 @@ func testTransactionVisibility(t *testing.T, db DB) {
 	})
 }
 
+func testConcurrency(t *testing.T, db DB) {
```
lovromazgon, line 190: I added a concurrency test that ensures the implementation can handle concurrent access. We need to fix this test, then sqlite can become an option.

pkg/foundation/database/sqlite/db.go
```diff
@@ -0,0 +1,240 @@
...
+
+var _ database.DB = (*DB)(nil)
+
+// New initializes the database structures needed by DB.
+func New(ctx context.Context, l log.CtxLogger, path, table string) (*DB, error) {
+	dbpath, err := dburl(path)
+	if err != nil {
+		return nil, cerrors.Errorf("failed to construct db path: %w", err)
+	}
+	fmt.Println("path", dbpath)
```
lovromazgon, line 47: Can we change this to use the logger `l.Info()...`? Also, if we do that, we should first do `l = l.WithComponent("sqlite.DB")` so the correct component is included in the output, and we can then remove that in line 66.

lovromazgon (approved):

PR #1683, opened 2024-06-28
add getSchema and createSchema for wasm processors

maha-hajja:
### Description

Related to https://github.com/ConduitIO/conduit-processor-sdk/issues/45

processor SDK PR https://github.com/ConduitIO/conduit-processor-sdk/pull/60

### Quick checks:


pkg/plugin/processor/standalone/host_module.go
```diff
@@ -145,3 +153,101 @@ func (m *hostModuleInstance) commandResponse(ctx context.Context, buf types.Byte
...
+// createSchema is the exported function that is called by the WASM module to
+// create a schema and set the buffer bytes to the marshalled response.
+// It returns the response size on success, or an error code on error.
+func (m *hostModuleInstance) createSchema(ctx context.Context, buf types.Bytes) types.Uint32 {
+	m.logger.Trace(ctx).Msg("executing create_schema")
+
+	var req processorv1.CreateSchemaRequest
+	var resp processorv1.CreateSchemaResponse
+	length := len(buf)
+	if m.parkedCreateSchemaRequest != nil {
```
lovromazgon, line 166: We don't need to park the create schema request, the processor will always send us the whole request, as it knows exactly how big it is. However, the processor has no idea about how big the response will be, so we might need to park the response.

That said, we need to store the request bytes (or a hash + size, but that might be overkill), so that if we have a parked response we can compare the requests and see if the processor actually repeated the same request.

Pseudocode for how I think this should work:
```
var buf types.Bytes
if parkedResponse != nil {
    if buf[:len(lastRequest)] == lastRequest {
        clear parkedResponse and lastRequest for next call
        return parkedResponse
    }
}
unmarshal request
response := schemaService.CreateSchema(unmarshalledRequest)
if len(response) > len(buf) {
    parkedResponse = response
    copy(lastRequest, buf)
    return len(response)
}
marshal response into buf
return len(response)
```

We could create a generic utility for this stuff so we can reuse it in all host functions.

pkg/plugin/processor/standalone/host_module.go
```diff
@@ -139,9 +155,137 @@ func (m *hostModuleInstance) commandResponse(ctx context.Context, buf types.Byte
...
+
+	var req procutilsv1.CreateSchemaRequest
+	length := len(buf)
+	if m.parkedCreateSchemaBuffer != nil {
+		length = len(m.parkedCreateSchemaBuffer)
+	}
+	err := proto.Unmarshal(buf[:length], &req)
+	if err != nil {
+		m.logger.Err(ctx, err).Msg("failed unmarshalling protobuf create schema request")
+		return pprocutils.ErrorCodeStart
```
hariso, line 194: Is there a reason for not returning a specific error here?

maha-hajja: because an error about failed unmarshalling or marshaling will not be useful for the user when trying to get or create a schema, this would be a problem from the processor SDK itself or conduit, and will be logged if we got a problem and have to fix it, but the user shouldn't encounter this error.
Other errors that will mean something to the user are the ones returned from "`create`" and "`get`" schema, like an invalid schema or a subject that doesn't exist.

hariso: I see... So maybe we can then have something like an "internal error"? `ErrorCodeStart` feels a bit unspecific.

pkg/plugin/processor/standalone/host_module.go
```diff
@@ -139,9 +155,137 @@ func (m *hostModuleInstance) commandResponse(ctx context.Context, buf types.Byte
...
+	m.parkedCreateSchemaResponse = nil
+	m.parkedCreateSchemaBuffer = nil
+
+	return types.Uint32(len(out))
+}
+
+// getSchema is the exported function that is called by the WASM module to
+// get a schema and set the buffer bytes to the marshalled response.
+// It returns the response size on success, or an error code on error.
+func (m *hostModuleInstance) getSchema(ctx context.Context, buf types.Bytes) types.Uint32 {
```
hariso, line 232: getSchema and createSchema work almost identically (check for parked response, get request, unmarshall, do something with it, etc.). WDYT about trying to refactoring this into "something" (a struct or a function) that does that?

lovromazgon: Makes sense, similarly to how we extracted the utility function in the SDK 👍

maha-hajja: thought about that, but the problem is with the number of different variables needed and functions and types, so I don't feel like it would look pretty, but will give it a go

pkg/plugin/processor/standalone/host_module.go
maha-hajja: we'd need to rename it for the processor SDK first, but I agree with the change, will crate a PR for it after this one

lovromazgon (approved):
In the spirit of the Olympic games, you win the gold medal for this PR 🥇 👏

PR #1631, opened 2024-05-31
 [Schemas] Connector SDK: Create and get a schema from Conduit

hariso:
### Description

TBD

### Quick checks:

- [ ] I have followed the [Code Guidelines](https://github.com/ConduitIO/conduit/blob/main/docs/code_guidelines.md).
- [ ] There is no other [pull request](https://github.com/ConduitIO/conduit/pulls) for the same update/change.
- [ ] I have written unit tests.
- [ ] I have made sure that the PR is of reasonable size and can be easily reviewed.

Issue #1653, opened 2024-06-12
[Schemas] Use configured schema service in Avro processors


hariso:
Part of #1560.

With schema support in Conduit, we come to the questions:
* Should the Avro processors be able to use Conduit's schema service?
* If yes, how?

hariso:
Comment by @lovromazgon:

> Not for this PR, but I'll mention it before I forget. I realized that having a builtin schema registry could create the expectation that the [`avro.decode`](https://conduit.io/docs/processors/builtin/avro.decode) and [`avro.encode`](https://conduit.io/docs/processors/builtin/avro.encode) processors would be able to use the internal schema registry. It would be a bit strange if we had to configure Conduit to point to a schema registry and then again supply the URL to the processors separately. So ideally the processors should use this service to create and get schemas. Not sure if this was included in the design or not.

https://github.com/ConduitIO/conduit/pull/1605#discussion_r1635023006

PR #1694, opened 2024-07-03
[Schemas] Connector SDK: Create and get a schema from Conduit

lovromazgon:
### Description

> [!NOTE]
> This PR was created as a clean version of https://github.com/ConduitIO/conduit/pull/1631, the first commit contains a cleaned up squashed commit with the latest state of that PR.

Part of https://github.com/ConduitIO/conduit-connector-sdk/issues/132.

### Quick checks:


.golangci.yml
```diff
@@ -1,5 +1,10 @@
 run:
   timeout: 3m
+  exclude-dirs-use-default: false
+  exclude-dirs:
+    - ^examples/
+    - ^ui/
+    - ^pkg/plugin/processor/builtin/internal/diff # external code
```
hariso, line 7: This should be under `issues`.

pkg/conduit/config.go
```diff
@@ -110,9 +110,9 @@ func DefaultConfig() Config {
 	cfg.Connectors.Path = "./connectors"
 	cfg.Processors.Path = "./processors"
 	cfg.Pipelines.Path = "./pipelines"
-	cfg.Schema.Type = SchemaTypeInMemory
+	cfg.SchemaRegistry.Type = SchemaRegistryTypeBuiltin
 
-	cfg.PluginDispenserFactories = builtin.DefaultDispenserFactories
+	cfg.ConnectorPlugins = builtin.DefaultBuiltinConnectors
```
hariso, line 115: Nitpick: maybe we rename `DefaultBuiltinConnector` to just `DefaultConnectors` (it's in the `builtin` package anyway)?

pkg/conduit/runtime.go
```diff
@@ -495,8 +492,9 @@ func (r *Runtime) startConnectorUtils(ctx context.Context, t *tomb.Tomb) (net.Ad
 		grpc.StatsHandler(r.gRPCStatsHandler),
 	)
 
-	schemaServiceAPI := utils.NewSchemaServiceAPIv1(r.schemaService)
-	schemaServiceAPI.RegisterInServer(grpcServer)
+	// TODO
```
hariso, line 495: This PR or another one?

pkg/schemaregistry/service.go
```diff
@@ -14,17 +14,61 @@
...
+	}
+	return pconduit.CreateSchemaResponse{
+		Schema: toschema.SrSubjectSchema(ss),
+	}, nil
+}
+
+func (s *Service) GetSchema(ctx context.Context, req pconduit.GetSchemaRequest) (pconduit.GetSchemaResponse, error) {
+	ss, err := s.registry.SchemaBySubjectVersion(ctx, req.Subject, req.Version)
+	if err != nil {
+		// TODO: convert error
```
hariso, line 67: This PR or another one?

pkg/conduit/runtime.go
lovromazgon: Whoops, this one!

pkg/schemaregistry/service.go
lovromazgon: This one for sure 😬

pkg/conduit/config.go
lovromazgon: Idk about this, I get that we have `builtin` in the package name, but somehow `DefaultConnectors` feels odd without "builtin" in there. What is a "default connector"? 🤔

hariso: To me it still reads nice: `DefaultConnectors` are the connectors you get by default. But as I said, just a nitpick, `builtin.DefaultBuiltinConnectors` is a good name as well!

pkg/schemaregistry/service.go
lovromazgon: I changed the conduit schema registry to [return `sr.ResponseError`](https://github.com/ConduitIO/conduit-schema-registry/pull/2) same as the client, then convert the errors accordingly. We need to test it end-to-end to know if it works as expected. I'll add unit tests separately.

pkg/plugin/connector/connutils/schema.go
```diff
@@ -0,0 +1,96 @@
...
+	conduitschemaregistry "github.com/conduitio/conduit-schema-registry"
+	"github.com/conduitio/conduit/pkg/foundation/cerrors"
+	"github.com/conduitio/conduit/pkg/foundation/log"
+	"github.com/conduitio/conduit/pkg/schemaregistry"
+	"github.com/conduitio/conduit/pkg/schemaregistry/fromschema"
+	"github.com/conduitio/conduit/pkg/schemaregistry/toschema"
+	"github.com/twmb/franz-go/pkg/sr"
+)
+
+type SchemaService struct {
```
lovromazgon, line 30: Btw, notice I moved `schemaregistry.Service` to `pkg/plugin/connector/connutils`, as this service should only be used by connectors. We will probably have something similar for processors. Now we have a central place to put future connector utilities (`StorageService` for instance).

It's similar to what you created before my refactoring, only that this service is used both for standalone and builtin plugins. This is also the type where we can add the connector token checks.

pkg/conduit/config.go
```diff
@@ -29,8 +30,8 @@ const (
 	DBTypePostgres = "postgres"
 	DBTypeInMemory = "inmemory"
 
-	SchemaTypeConfluent = "confluent"
-	SchemaTypeInMemory  = "inmemory"
+	SchemaRegistryTypeConfluent = "confluent"
+	SchemaRegistryTypeBuiltin   = "builtin"
```
lovromazgon, line 34: Hmm, thinking out loud - I changed the `inmemory` option to `builtin`, since it won't necessarily be in-memory, the goal is to use the same persistence layer as Conduit for storing schemas. Not I'm thinking if the other one should be called `standalone` (to mimic the terminology in plugins) or `external` (as the opposite of "builtin")? 🤔 cc @hariso @maha-hajja

pkg/plugin/connector/connutils/schema.go
```diff
@@ -0,0 +1,96 @@
...
+}
+
+func unwrapSrError(e *sr.ResponseError) error {
+	switch e.ErrorCode {
+	case conduitschemaregistry.ErrorCodeSubjectNotFound,
+		conduitschemaregistry.ErrorCodeVersionNotFound,
+		conduitschemaregistry.ErrorCodeSchemaNotFound:
+		return pconduit.ErrSchemaNotFound
+	case conduitschemaregistry.ErrorCodeInvalidSchema:
+		return pconduit.ErrInvalidSchemaBytes // TODO change to ErrInvalidSchema
```
lovromazgon, line 91: Sidenote: apparently, we don't get the info from the schema registry if the bytes are invalid, the type or the subject. We should change the errors in the protocol to match this behavior.

pkg/plugin/processor/builtin/impl/avro/internal/decoder.go
```diff
@@ -0,0 +1,59 @@
...
+
+func NewDecoder(registry schemaregistry.Registry, logger log.CtxLogger) *Decoder {
+	return &Decoder{
+		registry: registry,
+		logger:   logger.WithComponent("avro.internal.Decoder"),
+	}
+}
+
+func (d *Decoder) Decode(ctx context.Context, b opencdc.RawData) (opencdc.StructuredData, error) {
+	id, data, err := (&sr.ConfluentHeader{}).DecodeID(b.Bytes())
```
lovromazgon, line 41: The `avro.decode` and `avro.encode` processors work same as before, which means they encode/decode the Confluent wire format. Maybe we should think about renaming them to something like `confluent.avro.decode`? I'm not convinced, just pointing out that internally we don't need the wire format, because we store the required fields in the record metadata.

pkg/conduit/config.go
hariso: `builtin` sounds good to me. `confluent` is still a better option to me, because it makes it clear what kind of schema registry is supported.

pkg/plugin/connector/connutils/schema.go
hariso: Can you elaborate a bit more on the change in the protocol?

pkg/plugin/processor/builtin/impl/avro/decode.go
```diff
@@ -136,12 +137,13 @@ func (p *decodeProcessor) Configure(ctx context.Context, m map[string]string) er
 	return nil
 }
 
-func (p *decodeProcessor) Open(ctx context.Context) error {
+func (p *decodeProcessor) Open(context.Context) error {
+	// TODO: somehow inject the schemaregistry.Registry that Conduit created.
```
hariso, line 141: FYI https://github.com/ConduitIO/conduit/issues/1653

hariso (approved):
:raised_hands:

pkg/conduit/config.go
lovromazgon: Alright, I'll leave it as it is, I also like that `confluent` makes it explicit what API is expected.

pkg/plugin/connector/connutils/schema.go
lovromazgon: It's about the errors that are defined in the [connecor protocol](https://github.com/ConduitIO/conduit-connector-protocol/blob/35d371d9907ac9b0053b6405c514600118146e86/pconduit/errors.go#L22-L25) vs the ones in the [schema registry](https://github.com/ConduitIO/conduit-schema-registry/blob/277080c2e0462ac2c41d02bf467cc43dc39c8d65/errors.go#L43-L62). We should have the same errors (or a subset) in the protocol, since that's all the errors we get from the registry.

PR #1701, opened 2024-07-09
[Schemas] Limit schema service usage to connectors

### LINKED ISSUE #1563: [Schemas] Limit schema service usage to connectors

hariso:
### Description

Closes #1563.

Depends on:
- https://github.com/ConduitIO/conduit-connector-protocol/pull/112
- https://github.com/ConduitIO/conduit-connector-sdk/pull/150

### Quick checks:


go.mod
```diff
@@ -8,15 +8,15 @@ require (
...
 	github.com/conduitio/conduit-connector-file v0.6.1-0.20240621111422-221c138201d3
 	github.com/conduitio/conduit-connector-generator v0.6.1-0.20240621111436-e9fa3464f7b2
 	github.com/conduitio/conduit-connector-kafka v0.8.1-0.20240621111431-87c01cf39a06
 	github.com/conduitio/conduit-connector-log v0.3.1-0.20240621111440-e2f0f04a35a4
 	github.com/conduitio/conduit-connector-postgres v0.7.6-0.20240630172132-84b5a6e6104f
-	github.com/conduitio/conduit-connector-protocol v0.6.1-0.20240705154009-b938cfa7f251
+	github.com/conduitio/conduit-connector-protocol v0.6.1-0.20240711112538-4e32f3a34aa3
 	github.com/conduitio/conduit-connector-s3 v0.5.2-0.20240630172807-e278fde1fb46
-	github.com/conduitio/conduit-connector-sdk v0.9.2-0.20240705162050-971c5f7facc2
+	github.com/conduitio/conduit-connector-sdk v0.9.2-0.20240711113046-c35080163f50
```
hariso, line 19: Reminder to update the dependencies.

pkg/plugin/connector/standalone/registry.go
```diff
@@ -60,30 +63,32 @@ func NewRegistry(logger log.CtxLogger, pluginDir string) *Registry {
 			r.logger.Warn(context.Background()).Err(err).Msg("could not extract absolute connector plugins path")
 		} else {
 			r.pluginDir = absPluginDir // store plugin dir for hot reloads
-			r.reloadPlugins()
```
hariso, line 63: Moved into the `Init` method, so that we have two separate steps: one for just creating a registry and one for initializing it (loading all the plugins). This makes the registry more similar to other services constructed in runtime.go.

pkg/plugin/connector/builtin/dispenser.go
```diff
@@ -49,12 +49,12 @@ func (d *Dispenser) DispenseSpecifier() (connector.SpecifierPlugin, error) {
 	return newSpecifierPluginAdapter(d.specifierPlugin(), d.pluginLogger("specifier")), nil
 }
 
-func (d *Dispenser) DispenseSource() (connector.SourcePlugin, error) {
-	return newSourcePluginAdapter(d.sourcePlugin(), d.pluginLogger("source")), nil
+func (d *Dispenser) DispenseSource(connectorID string) (connector.SourcePlugin, error) {
```
hariso, line 52: It looks like a dispenser is always dispensing either sources or destinations, so I think I can get the connector ID into the dispenser as a field, and keep the method signature as it was (without any params).

go.mod
```diff
@@ -357,3 +357,7 @@ require (
 	mvdan.cc/gofumpt v0.6.0 // indirect
 	mvdan.cc/unparam v0.0.0-20240528143540-8a5130ca722f // indirect
 )
+
+replace (
+	github.com/conduitio/conduit-connector-postgres => ../conduit-connector-postgres
+)
```
hariso, line 363: Reminder to remove this before merging.

pkg/connector/instance.go
```diff
@@ -86,7 +86,7 @@ type Connector interface {
 
 // PluginDispenserFetcher can fetch a plugin dispenser.
 type PluginDispenserFetcher interface {
-	NewDispenser(logger log.CtxLogger, name string) (connectorPlugin.Dispenser, error)
+	NewDispenser(logger log.CtxLogger, name string, connectorID string) (connectorPlugin.Dispenser, error)
```
hariso, line 89: Random thing I noticed: a dispenser is able to dispense sources and destinations, but in practice a dispenser dispenses either sources or destinations.

pkg/plugin/connector/connutils/schema.go
```diff
@@ -38,6 +40,7 @@ func NewSchemaService(logger log.CtxLogger, registry schemaregistry.Registry) *S
 	return &SchemaService{
 		registry: registry,
 		logger:   logger.WithComponent("connutils.SchemaService"),
+		token:    uuid.NewString(),
```
lovromazgon, line 43: Looks like we create a global token for all connectors here. However, we said the token would be used to identify the connector that called the connector utility. This will be useful for logging, but also for things like namespacing when creating schemas (to prevent naming conflicts) or in a storage service in the future to keep data between connectors separate.

In my mind I imagined something like this:
- There should be some service that is tracking running connector plugins.
  - Maybe we can add that directly in `pkg/plugin/connector.PluginService`, although not sure if that would create circular dependencies.
- When creating a new dispenser, the registry would create a token specific for that connector and store it in a map of running connectors. That map allows us to know which token is used by which connector.
  - Or the dispenser could get a reference to `connector.PluginService` and it can register and deregister itself when it starts and when it stops. Registering would create the token and add it to the map of running connectors, deregistering would remove it from that map. That might be easier, since the dispenser actually knows when the plugin is started/stopped.
- The connector utility services (like this `SchemaService`) would also have a reference to the `connector.PluginService` so it can validate the token and get the info about which connector the token belongs to.

That's the high-level flow, the devil is in the detail of course, how to structure the code without circular dependencies. The point is that we'll end up with a place where we know what connector plugins are currently running and which token belongs to which plugin. The connector utilities can use that mapping to figure out if a token is valid and which connector sent the request.

Another approach would be to create a JWT token for each connector plugin so that the connector ID is stored in the token itself. This way we wouldn't need that mapping in a central place, since we don't have a different use case for that mapping right now other than connector utilities.

pkg/plugin/connector/service.go
```diff
@@ -26,29 +26,41 @@ import (
...
 	List() map[plugin.FullName]pconnector.Specification
 }
 
+// The built-in registry creates a dispenser which dispenses a plugin adapter
+// that communicates with the plugin directly as if it was a library. These
+// plugins are baked into the Conduit binary and included at compile time.
+type builtinReg interface {
+	registry
+
+	Init(context.Context)
```
lovromazgon, line 41: Maybe we can make our lives easier and just add `Init(ctx context.Context, connUtilsAddr string, connUtilsToken string)` to `registry`, the builtin one can ignore the params it doesn't need 🤷

pkg/plugin/connector/standalone/registry.go
```diff
@@ -60,30 +64,32 @@ func NewRegistry(logger log.CtxLogger, pluginDir string) *Registry {
...
 
-	r.logger.Info(context.Background()).
-		Str(log.PluginPathField, r.pluginDir).
-		Int("count", len(r.List())).
-		Msg("standalone connector plugins initialized")
-
 	return r
 }
 
-func (r *Registry) reloadPlugins() {
```
lovromazgon, line 75: Reload plugins was created purposefully, so we could create a goroutine that calls it repeatedly with a sleep interval, this way we could hot-reload plugins that are dropped in the folder while Conduit is running. Not a big thing though, we can easily extract it when we add that, just explaining why it was there.

pkg/plugin/connector/service.go
hariso: I started with that, but given that the connector utils params are specific for the standalone, I went this route.

pkg/plugin/connector/connutils/schema.go
hariso: Yup, you're right about the static token, and for this PR it was actually intentional, because it's enough to limit the usage of the schema service to the connectors. I'm about to implement proper authentication with tokens when handling name conflicts.

pkg/conduit/runtime.go
```diff
@@ -94,7 +94,6 @@ type Runtime struct {
 	connectorPluginService *conn_plugin.PluginService
 	processorPluginService *proc_plugin.PluginService
 
-	schemaRegistry    schemaregistry.Registry
```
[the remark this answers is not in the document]
hariso: Like when Conduit is embedded?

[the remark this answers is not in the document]
lovromazgon: Yes.

pkg/plugin/connector/service.go
```diff
@@ -57,24 +69,29 @@ func NewPluginService(
...
+	s.builtinReg.Init(ctx)
+	s.standaloneReg.Init(ctx, connUtilsAddr, connUtilsToken)
+}
+
 func (s *PluginService) Check(context.Context) error {
 	return nil
 }
 
-func (s *PluginService) NewDispenser(logger log.CtxLogger, name string) (Dispenser, error) {
+func (s *PluginService) NewDispenser(logger log.CtxLogger, name string, connectorID string) (Dispenser, error) {
```
lovromazgon, line 81: Seems like in this function we have all the required information to create `pconnector.PluginConfig` - in the next PR I propose managing the config and token here (create, register so it can be used in conn utils, deregister when connector stops), otherwise we'll do the same work twice in the builtin and standalone dispenser.

pkg/conduit/runtime.go
```diff
@@ -211,12 +200,31 @@ func initServices(r *Runtime) error {
 	r.processorService = procService
 	r.connectorPluginService = connPluginService
 	r.processorPluginService = procPluginService
-	r.schemaRegistry = schemaRegistry
 	r.connSchemaService = connSchemaService
 
 	return nil
 }
 
+func createSchemaService(config Config, logger log.CtxLogger) (*connutils.SchemaService, error) {
```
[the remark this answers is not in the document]
lovromazgon: Only this part is missing.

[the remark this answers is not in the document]
hariso: Ah, didn't commit, doing that now!

pkg/plugin/connector/service.go
hariso: Yup, makes sense!

lovromazgon (approved):

PR #1718, opened 2024-07-23
[Schemas] Handle name conflicts

### LINKED ISSUE #1660: [Schemas] Handle name conflicts
Part of #1560.

Currently, our API allows connector developers to specify a schema name to be used. Given that Conduit's schema registry is shared by multiple connectors and pipelines, we need to handle name conflicts.

In other words: different connectors should be allowed to use same schema names, but that shouldn't have any side effects (such as one connector modify the schema from another connector).

This might be useful: [Use Schema Contexts in Confluent Platform](https://docs.confluent.io/platform/current/schema-registry/schema-contexts-cp.html).

Pull requests:
- https://github.com/ConduitIO/conduit-connector-protocol/pull/115 
- https://github.com/ConduitIO/conduit-connector-sdk/pull/153
- #1718

hariso:
### Description

Closes #1660.

Depends on:
- https://github.com/ConduitIO/conduit-connector-sdk/pull/153
- https://github.com/ConduitIO/conduit-connector-protocol/pull/115

### Quick checks:


go.mod
```diff
@@ -1,22 +1,22 @@
...
+	github.com/conduitio/conduit-connector-file v0.6.1-0.20240709112929-6207f9f8efcf
 	github.com/conduitio/conduit-connector-generator v0.6.1-0.20240621111436-e9fa3464f7b2
 	github.com/conduitio/conduit-connector-kafka v0.8.1-0.20240621111431-87c01cf39a06
 	github.com/conduitio/conduit-connector-log v0.3.1-0.20240621111440-e2f0f04a35a4
 	github.com/conduitio/conduit-connector-postgres v0.7.6-0.20240630172132-84b5a6e6104f
-	github.com/conduitio/conduit-connector-protocol v0.6.1-0.20240716144613-8ef04cf70e73
+	github.com/conduitio/conduit-connector-protocol v0.6.1-0.20240723164726-270199a40f09
 	github.com/conduitio/conduit-connector-s3 v0.5.2-0.20240630172807-e278fde1fb46
-	github.com/conduitio/conduit-connector-sdk v0.9.2-0.20240716152132-20953984628f
+	github.com/conduitio/conduit-connector-sdk v0.9.2-0.20240724125119-3399efefc218
```
hariso, line 19: Reminder for myself to update the dependencies.

pkg/plugin/connector/cleanup_dispenser.go
```diff
@@ -0,0 +1,82 @@
...
+
+func (c *CleanupDispenser) DispenseSpecifier() (SpecifierPlugin, error) {
+	return c.Target.DispenseSpecifier()
+}
+
+func (c *CleanupDispenser) DispenseSource() (SourcePlugin, error) {
+	plugin, err := c.Target.DispenseSource()
+	if err != nil {
+		return nil, err
+	}
```
lovromazgon, line 38: Maybe I'm paranoid, but if `c.Cleanup` is `nil` this could potentially `panic`. Can we skip wrapping the plugin if that's the case? 🙏 
```suggestion
	if err != nil || c.Cleanup == nil {
		return plugin, err
	}
```

pkg/plugin/connector/service.go
```diff
@@ -81,22 +85,24 @@ func (s *PluginService) Check(context.Context) error {
...
-	case plugin.PluginTypeAny:
-		d, err := s.standaloneReg.NewDispenser(logger, fullName, connectorID)
-		if err != nil {
-			s.logger.Debug(context.Background()).Err(err).Msg("could not find standalone plugin dispenser, falling back to builtin plugin")
-			d, err = s.builtinReg.NewDispenser(logger, fullName, connectorID)
-		}
-		return d, err
-	default:
-		return nil, cerrors.Errorf("invalid plugin name prefix %q", fullName.PluginType())
+	// todo deregister token when connector stops running
```
lovromazgon, line 88: It's done 👍

pkg/plugin/connector/standalone/registry.go
```diff
@@ -166,8 +164,9 @@ func (r *Registry) loadSpecifications(pluginPath string) (pconnector.Specificati
 	dispenser, err := NewDispenser(
 		zerolog.Nop(),
 		pluginPath,
+		// todo are the grpc target and token really needed?
```
lovromazgon, line 167: We can keep them in case there's some code that expects a value there, but most probably the connector won't need them here. The only use case I see its the grpc target, if the connector wants to check that a service is available or something.

pkg/plugin/connector/connutils/auth.go
```diff
@@ -0,0 +1,79 @@
...
+)
+
+var ErrInvalidToken = cerrors.New("invalid token")
+
+type Token struct {
+	Value       string `json:"value"`
+	ConnectorID string `json:"connector_id"`
+}
+
+type TokenService struct {
```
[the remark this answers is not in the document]
hariso: Yup, AuthManager sounds good. :+1:

pkg/plugin/connector/cleanup_dispenser.go
hariso: Hmm... Maybe not even create a cleanup dispenser if there's no cleanup function?

lovromazgon: Right, that should be the case, but I was looking at the cleanup dispenser in isolation and saw the potential for a panic when it's not used correctly. It's unlikely to happen, assuming we have the check elsewhere, but I would sleep better if we write code that doesn't make such assumptions 😄

hariso: It's not quite up to best practices, because it returns an interface instead of a struct, but I'd say it pays out.

lovromazgon: I'd be up for that, if we rename the method. It's unexpected for a constructor to do this, but not for something like `WithCleanup` or even `OnTeardown`, since the function is called on teardown.

hariso: Done!

lovromazgon (approved):

PR #1720, opened 2024-07-24
Replace pkg/foundation/database with commons, use schema registry with DB

### LINKED ISSUE #1719: [Schemas] Replace pkg/foundation/database with database package in conduit-commons
As part of this issue we also need to update the conduit schema registry to the latest code in `main` and make sure we pass the database to the registry so the schemas get persisted.

hariso:
### Description

Closes #1719. Depends on https://github.com/ConduitIO/conduit-schema-registry/pull/5.

### Quick checks:


pkg/foundation/ctxutil/transaction.go
```diff
@@ -17,7 +17,7 @@ package ctxutil
 import (
 	"context"
 
-	"github.com/conduitio/conduit/pkg/foundation/database"
+	"github.com/conduitio/conduit-commons/database"
 )
 
 // transactionCtxKey is used as the key when saving the transaction in a context.
```
lovromazgon, line 23: We should also remove this file and use the corresponding functions in the `database` package instead.

lovromazgon (approved):
💪

PR #1739, opened 2024-08-01
Update Buf actions

### LINKED ISSUE #1710: CI: Authenticate Buf actions
Email from Buf:

> As of **August 1, 2024**, Buf will begin limiting remote plugin requests from unauthenticated users on [buf.build](https://buf.build/). If you send more than 10 unauthenticated requests per hour using [remote plugins](https://buf.build/docs/bsr/remote-plugins/overview), you’ll begin to see rate limit errors.
> 
> We’ll start by failing a small percentage of unauthenticated requests, increasing over time until 100% of unauthenticated requests are blocked. To prevent service interruptions, please update your CI jobs and local usages of the Buf CLI to authenticate with the Buf Schema Registry (BSR) before making these requests.

We use remote plugins in our config and need to authenticate in the action that generates the buf code ([`validate-generated-files.yml`](https://github.com/ConduitIO/conduit/blob/main/.github/workflows/validate-generated-files.yml)). See [Buf docs](https://buf.build/docs/bsr/authentication) on how to achieve that.

We need to do the same in all repos that use buf (at least conduit-commons, conduit-connector-protocol and conduit-processor-sdk, investigate if there are any others).

Repositories on https://buf.build/conduitio:

- [x] conduit https://github.com/ConduitIO/conduit/pull/1770
- [ ] conduit-commons https://github.com/ConduitIO/conduit-commons/pull/93
- [x] conduit-connector-protocol https://github.com/ConduitIO/conduit-connector-protocol/pull/124
- [x] conduit-processor-sdk https://github.com/ConduitIO/conduit-processor-sdk/pull/72

hariso:
### Description

Closes #1710.

- Updates the Buf jobs (uses the new action)
- Splits existing workflow into two separate ones:
  - one that validates changes in a pull request
  - one that pushes the changes into BSR (once they've been merged into our `main` branch) 
- Uses a token (that we already have set up in a secret) to authenticate requests to Buf.

The validation currently fails, because we haven't updated `buf.lock` when we made changes to conduit-commons (part of the schema related work in 0.11). A workflow that regularly runs `buf update` is implemented in https://github.com/ConduitIO/conduit/pull/1772.

### Quick checks


hariso:
Looks like Buf released a new action that unifies all of these actions. I'll convert this PR to a draft.

lovromazgon:
Also keep in mind that we need the token specifically when generating the proto code in [`validate-generated-files.yml`](https://github.com/ConduitIO/conduit/blob/2da3a0b378d0e7b188984d2d6f8d2e02b3614ba1/.github/workflows/validate-generated-files.yml#L22) because we're using remote code generation.

hariso:
> Also keep in mind that we need the token specifically when generating the proto code in [`validate-generated-files.yml`](https://github.com/ConduitIO/conduit/blob/2da3a0b378d0e7b188984d2d6f8d2e02b3614ba1/.github/workflows/validate-generated-files.yml#L22) because we're using remote code generation.

I was looking at that too, but interestingly, `buf` isn't complaining about the token there. I plan to take a closer at that (I mean, it doesn't hurt just giving it the token, but still).

proto/api/v1/api.swagger.json
```diff
@@ -2192,6 +2192,9 @@
         },
         "description": {
           "type": "string"
+        },
+        "descriptionAgain": {
+          "type": "string"
```
hariso, line 2197: Test change that needs to be removed.

proto/api/v1/api.proto
hariso, line 1: File reformated with `buf format`.

.github/workflows/buf-validate.yaml
```diff
@@ -0,0 +1,33 @@
+name: buf-validate
+
+on:
+  pull_request:
+    # 'labeled' and 'unlabeled' are included so we can use the 'Buf Skip Breaking' label
+    # and skip a breaking change if that's required.
+    # See: https://github.com/bufbuild/buf-action?tab=readme-ov-file#skip-the-breaking-change-detection-step
```
lovromazgon, line 7: So I guess this PR is a good candidate to add this label now?

.github/workflows/buf-push.yaml
```diff
@@ -0,0 +1,33 @@
...
+        id: buf-version
+        run: |
+          BUF_VERSION=$( go list -m -f '{{.Version}}' github.com/bufbuild/buf )
+          # remove the leading 'v' in front of the version
+          BUF_VERSION=${BUF_VERSION#v}
+          echo "v=$BUF_VERSION" >> "$GITHUB_OUTPUT"
+      - uses: bufbuild/buf-action@v1
+        with:
+          version: ${{ steps.buf-version.outputs.v }}
+          token: ${{ secrets.BUF_TOKEN }}
```
lovromazgon, line 28: So regarding the buf token in the validate action, try to execute this and you'll see the rate limiting errors:
```
$ for i in {1..100}; do make proto-generate; done
```

The error looks like this:

```
Failure: Unauthenticated remote plugin rate limit exceeded. For info on how to resolve this, visit: https://buf.build/b/auth-required
make: *** [proto-generate] Error 1
```

Currently it doesn't happen often, but according to Buf they will lower the limit until no unauthenticated requests are allowed.

lovromazgon: Oh and before doing that, execute `buf registry logout` to make sure you are not logged in.

hariso: I see it now... It looks like logs I was looking for were actually coming from the buf GitHub actions not the tool itself. I'll fix this too, thanks for calling out.

lovromazgon (approved):
LGTM 💪  Please do the same in the processor SDK, connector protocol and conduit commons, I believe we are using buf there as well.

PR #1742, opened 2024-08-02
Update processor SDK, add middleware

lovromazgon:
### Description

This PR includes the following changes:

- apply processor middleware to builtin processors
- use `config.Config` as configuration map type
- take into account middleware parameters in tests

Depends on https://github.com/ConduitIO/conduit-processor-sdk/pull/68.

Fixes https://github.com/ConduitIO/conduit/issues/1707

### Quick checks

- [X] I have followed the [Code Guidelines](https://github.com/ConduitIO/conduit/blob/main/docs/code_guidelines.md).
- [X] There is no other [pull request](https://github.com/ConduitIO/conduit/pulls) for the same update/change.
- [X] I have written unit tests.
- [X] I have made sure that the PR is of reasonable size and can be easily reviewed.

PR #1743, opened 2024-08-05
[Schemas] Use configured schema service in Avro processors

### LINKED ISSUE #1653: [Schemas] Use configured schema service in Avro processors

hariso:
### Description

Closes #1653.

Depends on https://github.com/ConduitIO/conduit/pull/1742.

Tested with Confluent's Schema Registry and two pipelines (one that encodes data from a file, and another that reads the encoded and decodes it back).

### Quick checks

### DIFF
--- a/pkg/plugin/processor/builtin/impl/avro/encode.go
+++ b/pkg/plugin/processor/builtin/impl/avro/encode.go
@@ -21,7 +21,6 @@ import (
 	"context"
 
 	"github.com/conduitio/conduit-commons/config"
-	"github.com/conduitio/conduit-commons/lang"
 	"github.com/conduitio/conduit-commons/opencdc"
 	sdk "github.com/conduitio/conduit-processor-sdk"
 	"github.com/conduitio/conduit/pkg/foundation/cerrors"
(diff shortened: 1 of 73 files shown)


pkg/plugin/processor/builtin/registry.go
```diff
@@ -74,10 +81,10 @@ type ProcessorPluginConstructor func(log.CtxLogger) sdk.Processor
 func NewRegistry(
 	logger log.CtxLogger,
 	constructors map[string]ProcessorPluginConstructor,
-	schemaService pprocutils.SchemaService,
+	schemaRegistry schemaregistry.Registry,
```
hariso, line 84: I don't like this change much, but the `avro.encode` and `avro.decode` processors need a schema registry, not a schema service.

pkg/plugin/processor/builtin/impl/avro/encode.go
```diff
@@ -107,15 +71,33 @@ func parseEncodeConfig(ctx context.Context, c config.Config) (encodeConfig, erro
...
+	return []sdk.ProcessorMiddlewareOption{
+		sdk.ProcessorWithSchemaDecodeConfig{
+			PayloadEnabled: lang.Ptr(false),
+			KeyEnabled:     lang.Ptr(false),
+		},
+		sdk.ProcessorWithSchemaEncodeConfig{
+			PayloadEnabled: lang.Ptr(false),
+			KeyEnabled:     lang.Ptr(false),
+		},
+	}
```
lovromazgon, line 94: As mentioned in slack, I think disabling this could lead to some unexpected behavior 😕

pkg/plugin/processor/builtin/registry.go
lovromazgon: The thing that bothers me a bit is that the standalone and builtin registry now have each their own schema service. But I don't think it's a big issue right now, since the service is stateless, so I think we can live with this.

lovromazgon (approved):
Looks good, just that one thing about the disabled middleware, approving in advance 👍

PR #1745, opened 2024-08-05
Tests and examples for converting time using `field.convert`

lovromazgon:
### Description

The title pretty much describes it. One small but important change is the line where we ensure the timezone is set to `UTC` when converting the value.

### Quick checks

### DIFF
--- a/pkg/plugin/processor/builtin/impl/field/convert_examples_test.go
+++ b/pkg/plugin/processor/builtin/impl/field/convert_examples_test.go
@@ -15,6 +15,8 @@
 package field
 
 import (
+	"time"
+
 	"github.com/conduitio/conduit-commons/config"
 	"github.com/conduitio/conduit-commons/opencdc"
 	sdk "github.com/conduitio/conduit-processor-sdk"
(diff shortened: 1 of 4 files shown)


pkg/plugin/processor/builtin/impl/field/convert_examples_test.go
```diff
@@ -144,3 +146,46 @@ func ExampleConvertProcessor_floatToString() {
...
 }
+
+//nolint:govet // a more descriptive example description
+func ExampleConvertProcessor_intTotime() {
+	p := NewConvertProcessor(log.Nop())
+
+	timeObj := time.Date(2024, 1, 2, 12, 34, 56, 123456789, time.UTC)
+
+	exampleutil.RunExample(p, exampleutil.Example{
+		Summary:     "Convert `string` to `time`",
```
hariso, line 157: I guess this is a slip, should be int to time?

hariso (approved):
LGTM!

pkg/plugin/processor/builtin/impl/field/convert_examples_test.go
lovromazgon: Absolutely, good catch! 👀

PR #1765, opened 2024-08-12
[Design] Recover from pipeline errors

lovromazgon:
Design document for recovering from pipeline errors.

Discussion: https://github.com/ConduitIO/conduit/discussions/1559


docs/design-documents/20240812-recover-from-pipeline-errors.md
```diff
@@ -0,0 +1,194 @@
...
+and control to the user. This
+includes [record deduplication](#record-deduplication) and
+the [audit log](#audit-log), which are bigger topics and will require a separate
+design document.
+
+## Open questions
+
+- **Should failed processors trigger a retry, or should the pipeline be
+  considered failed immediately?**
+  - I am leaning towards making processor faults fatal by default, as most of
```
maha-hajja, line 178: +1

docs/design-documents/20240812-recover-from-pipeline-errors.md
```diff
@@ -0,0 +1,194 @@
...
+implement a backoff delay strategy. This strategy introduces a waiting period
+before attempting to restart the pipeline, allowing time for issues to resolve
+themselves. The backoff parameters will be configurable to allow customization
+based on specific production needs.
+
+- **Configurable Parameters:**
+  - **Minimum Delay Before Restart:** Default: 1 second
+  - **Maximum Delay Before Restart:** Default: 1 minute
+  - **Backoff Factor:** Default: 2
+  - **Maximum Number of Retries:** Default: Infinite
```
hariso, line 67: IMHO, we should limit the number of retries. If, after many retries, an error is still not gone, then it's very likely not a transient error (which the pipeline engine thinks it is).

docs/design-documents/20240812-recover-from-pipeline-errors.md
```diff
@@ -0,0 +1,194 @@
...
+tracks records that have been successfully processed by the destination, but the
+acknowledgment wasn't processed by the source yet. If such a record is
+encountered after restarting the pipeline, Conduit should acknowledge it without
+sending it to the destination.
+
+## Audit log
+
+If a pipeline is automatically recovered, a user needs to check the Conduit logs
+to see the cause and what actually happened. Ideally that information would be
+retrievable through the API and would be attached to the pipeline itself so a
```
hariso, line 113: >  Ideally that information would be retrievable through the API

I guess you mean the Conduit API? How about we add this to the metrics? It might be easier to integrate with (because of existing tools that use those metrics), will provide value and it's also relatively easy to implement.

docs/design-documents/20240812-recover-from-pipeline-errors.md
```diff
@@ -0,0 +1,194 @@
...
+before attempting to restart the pipeline, allowing time for issues to resolve
+themselves. The backoff parameters will be configurable to allow customization
+based on specific production needs.
+
+- **Configurable Parameters:**
+  - **Minimum Delay Before Restart:** Default: 1 second
+  - **Maximum Delay Before Restart:** Default: 1 minute
+  - **Backoff Factor:** Default: 2
+  - **Maximum Number of Retries:** Default: Infinite
+  - **Delay After Which Fail Count is Reset:** Default: 5 minutes (or on first
```
hariso, line 68: I'm not sure I understand this part... Does this mean that after 5 minutes, we start counting the failures from 0?

docs/design-documents/20240812-recover-from-pipeline-errors.md
```diff
@@ -0,0 +1,194 @@
...
+- **Should failed processors trigger a retry, or should the pipeline be
+  considered failed immediately?**
+  - I am leaning towards making processor faults fatal by default, as most of
+    the processors are stateless and always behave the same. Processors, which
+    connect to a 3rd party resource and can experience transient errors, should
+    implement retrials internally (like the `webhook.http` processor). Once we
+    implement [error classification](#error-classification), processors should
+    be allowed to return transient errors which should be treated as such.
+- **How does this functionality interact with a DLQ, especially with a nack
+  threshold?**
```
raulb, line 185: To me, this is the biggest question. How to pair the pipeline recovery feature with a DLQ. Considering that I value more providing good defaults on pipeline recovery, keeping the default of a nack threshold of 0 on DLQs makes perfect sense. We need to be careful of documenting well where this pipeline recovery takes effect so it's clear to the user that there are two different levels of tolerance. One for records that can't be processed successfully, and another for once a pipeline errors in a more _expected_ way.

docs/design-documents/20240812-recover-from-pipeline-errors.md
lovromazgon: In case of a fatal error, you're completely right, we will be trying again and again forever without the possibility to recover, which is not ideal. But let's say we do limit the maximum number of retries - what would be a sane default for this limit?

This number essentially defines the time window in which the issue should be resolved, otherwise the pipeline will be degraded and an operator needs to step in to resume the pipeline. If I set up a pipeline in production, I probably want it to run forever without my intervention (or at least with as little intervention as possible). Wouldn't that speak for retrying every now and then?

How about we increase the maximum delay to something like 10 minutes instead? 🤔

docs/design-documents/20240812-recover-from-pipeline-errors.md
```diff
@@ -0,0 +1,194 @@
...
+
+To avoid endless restarts, Conduit will include an option to permanently stop
+the pipeline after a configurable number of consecutive failures. Once the
+pipeline successfully processes at least one record or runs without encountering
+an error for a predefined time frame, the count of consecutive failures will
+reset.
+
+## Pipeline state management
+
+A new pipeline state, `recovering`, will be introduced to indicate that a
```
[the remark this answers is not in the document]
lovromazgon: Makes sense, stopping a recovering pipeline should definitely be possible 👍

docs/design-documents/20240812-recover-from-pipeline-errors.md
lovromazgon: Yes, I meant the Conduit API (HTTP, gRPC). I'm not following though, how would you return that information through metrics? 🤔 I'm talking about having a log of events that happened on a pipeline. Or maybe a global log of events, so that actions like deleting a pipeline are also recorded, even if the pipeline doesn't exist anymore.

docs/design-documents/20240812-recover-from-pipeline-errors.md
lovromazgon: > pipeline is running for at least 5 minutes, i.e. that's when we consider it healthy again

Yes, the purpose of this is that a pipeline might be healthy even if no data is flowing, since the source just may have no records to produce. In that case, we would mark the pipeline as healthy after some period of time.

This is not ideal of course, but unless we add a specific "health" method on the connector to tell us if everything is fine, we have no way of knowing if a pipeline is doing fine or still trying to recover.

We could just as well skip this, stay on the cautious side and leave the pipeline in `recovering` until we get the first record, although in that case a healthy pipeline could be marked as `recovering` for a potentially long time.

docs/design-documents/20240812-recover-from-pipeline-errors.md
```diff
@@ -0,0 +1,194 @@
...
+
+If a pipeline is automatically recovered, a user needs to check the Conduit logs
+to see the cause and what actually happened. Ideally that information would be
+retrievable through the API and would be attached to the pipeline itself so a
+user could get insights into how a pipeline is behaving.
+
+To add this feature, we would need to track what is happening on a pipeline and
+log any important events so we can return them to the user. When done properly,
+this could serve as an audit log of everything a pipeline went through (
+provisioning, starting, fault, recovery, stopping, deprovisioning etc.).
```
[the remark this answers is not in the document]
lovromazgon: I was mostly using it to illustrate what this could potentially look like. Although I think it would be nice to have more detailed state management, I don't think it's strictly needed for this specific feature.

hariso (approved):
LGTM!

maha-hajja (approved):
lgtm!

docs/design-documents/20240812-recover-from-pipeline-errors.md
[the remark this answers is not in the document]
hariso: For the record, in one of the team syncs we talked about this and mentioned that we'll try to make all the new changes as generic as possible so they allow new pipeline states, but we won't introduce any other new pipeline states in the scope of the recovery epic.

docs/design-documents/20240812-recover-from-pipeline-errors.md
hariso: We discussed this in an "offline" meeting, the default can stay `infinite`, as this will be configurable any way and called out in the docs.

docs/design-documents/20240812-recover-from-pipeline-errors.md
raulb: Update: We discussed this the other day at standup at length. I don't think we came up with a great conclusion, so we'll need to continue thinking about this one while in development.

docs/design-documents/20240812-recover-from-pipeline-errors.md
hariso: We discussed this in a meeting, we'll add some metrics in the first iteration.

PR #1794, opened 2024-08-19
Handle concurrent access to a pipeline's status field

### LINKED ISSUE #1778: [Recovery] Make pipeline status an atomic value
Currently the `pipeline.Instance` field `Status` can change at any point in time when the pipeline is running, therefore it's not safe to access this value. We should change the type into [`atomic.Pointer[Status]`](https://pkg.go.dev/sync/atomic#Pointer) so the value can be accessed safely.

hariso:
### Description

Closes #1778.

The approach that we initially planned didn't quite work and was changed to use a `sync.Mutex` because of following:
1. Having an exported field (and this is part of the "public" Conduit API) that's an `atomic` is sub-optimal user experience because it forces the user to work with the low-level constructs. In other words we should either expose a field or methods that work with just the `pipeline.Status` field.
2. With a field, it's possible to simply overwrite it: `myPipeline.Status = atomic.Pointer{}`
3. We still need to provide a getter and setter because otherwise working with `atomic.Pointer` would involve a few steps:
    ```go
    // set status
    myPipeline.Status.Store(lang.Ptr(theStatus))
    // get status
    theStatus := *myPipeline.Status.Load()
    ```
5. A pointer, we need to make sure the field is properly initialized.

For the reasons above, this PR uses a `sync.Mutex` and adds a getter and setter method for the status field, that is now unexported.

### Quick checks


pkg/pipeline/instance.go
```diff
@@ -58,10 +58,20 @@ type Instance struct {
 	ConnectorIDs []string
 	ProcessorIDs []string
 
+	status     Status
+	statusLock sync.RWMutex
```
raulb, line 62: @hariso do you think this lock would be relevant only to status or we could have the opportunity to have a lock for anything that would mutate the pipeline? Something tells me that this is basically it, but maybe with the need of a broader naming? What do you think?

For context, this line made me think of that https://github.com/ConduitIO/conduit/pull/1794/files#diff-576b2656f57e455d4f1e5405fd78db341ddef5d190447eb2e15b5088bacc0980R96.

pkg/pipeline/instance.go
```diff
@@ -58,10 +58,20 @@ type Instance struct {
...
+
 	n map[string]stream.Node
 	t *tomb.Tomb
 }
 
+// encodableInstance is an encodable "view" of Instance
+// through which we can also encode an Instance's unexported fields.
+type encodableInstance struct {
+	*Instance
+	Status Status
```
raulb, line 72: Do we want `Status` to be unexported here as well to ensure we're not mutating an encoded status?

raulb (approved):
@hariso LGTM! I only posted a non blocking question.

pkg/pipeline/instance.go
hariso: I imagine we can do it! We would need to think about what exactly needs to be locked since there might be cases where we don't want to lock the whole pipeline for a single field.

PR #1797, opened 2024-08-20
Add 'recovering' status

### LINKED ISSUE #1779: [Recovery] Add `recovering` state
Add the state to the proto enum [`Status`](https://github.com/ConduitIO/conduit/blob/d4e617775fb05e20261389da31a7506566e8b47b/proto/api/v1/api.proto#L14-L22) as well as to the [converter](https://github.com/ConduitIO/conduit/blob/d4e617775fb05e20261389da31a7506566e8b47b/pkg/web/api/toproto/pipeline.go#L45).

As part of this issue we should also allow the state in the [`Stop`](https://github.com/ConduitIO/conduit/blob/d4e617775fb05e20261389da31a7506566e8b47b/pkg/pipeline/lifecycle.go#L127) function (a recovering pipeline should be able to stop), also check if there are other places where we need to include it.

hariso:
### Description

Closes #1779.

### Quick checks


pkg/pipeline/lifecycle.go
```diff
@@ -124,7 +124,7 @@ func (s *Service) Stop(ctx context.Context, pipelineID string, force bool) error
 		return err
 	}
 
-	if pl.Status != StatusRunning {
+	if pl.Status != StatusRunning && pl.Status != StatusRecovering {
```
hariso, line 127: It should be possible to stop a recovering pipeline.

Makefile
```diff
@@ -51,7 +51,7 @@ run:
 
 .PHONY: proto-generate
 proto-generate:
-	rm -rf proto/gen && cd proto && buf generate
```
raulb, line 54: Good catch

pkg/pipeline/lifecycle.go
```diff
@@ -138,7 +138,10 @@ func (s *Service) Stop(ctx context.Context, pipelineID string, force bool) error
 }
 
 func (s *Service) stopGraceful(ctx context.Context, pl *Instance, reason error) error {
-	s.logger.Info(ctx).Str(log.PipelineIDField, pl.ID).Msg("gracefully stopping pipeline")
+	s.logger.Info(ctx).
```
raulb, line 141: I wonder if we need a `Stopping` pipeline status.

.golangci.yml
```diff
@@ -71,7 +71,7 @@ linters:
     - errcheck
     - errname
     # - errorlint
-    # - exhaustive
+    - exhaustive
```
raulb, line 74: @hariso What made you realize we needed to enable this back?

hariso: In one part of the code, there was a switch statement that was transforming proto statuses into Go statuses.

raulb (approved):

Issue #1780, opened 2024-08-14
[Recovery] Introduce fatal errors


lovromazgon:
We need a specific error type so the recovery mechanism will be able to detect which errors are fatal and which ones can be retried.

A fatal error should be returned in the [processor node](https://github.com/ConduitIO/conduit/blob/d4e617775fb05e20261389da31a7506566e8b47b/pkg/pipeline/stream/processor.go#L113) to signal that a processor nacking a record is a fatal error, as well as in the [DLQ](https://github.com/ConduitIO/conduit/blob/d4e617775fb05e20261389da31a7506566e8b47b/pkg/pipeline/stream/dlq.go#L149-L152) when the nack threshold is reached _and_ the nack threshold is more than 0.

It should be possible to convert any existing error into a fatal error, for example:

```
err := foo()
if err != nil {
	return cerrors.FatalError(err)
}
```

The fatal error should implement `Unwrap() error` so that the wrapped error can be accessed. We should also provide a convenience function to check if it's a fatal error like `cerrors.IsFatalError(err)`.

PR #1811, opened 2024-08-28
feat: add fatal errors

raulb:
### Description

Fixes https://github.com/ConduitIO/conduit/issues/1780

### Quick checks


pkg/foundation/cerrors/fatal.go
```diff
@@ -0,0 +1,41 @@
...
+}
+
+// Unwrap returns the wrapped error.
+func (f *FatalError) Unwrap() error {
+	return f.Err
+}
+
+// Error returns the error message.
+func (f *FatalError) Error() string {
+	return f.Err.Error()
```
maha-hajja, line 34: we should first check if `f.Err` is not `nil`  or this will cause a panic

pkg/foundation/cerrors/fatal_test.go
```diff
@@ -0,0 +1,78 @@
...
+	"testing"
+
+	"github.com/conduitio/conduit/pkg/foundation/cerrors"
+)
+
+func TestNewFatalError(t *testing.T) {
+	err := cerrors.New("test error")
+	fatalErr := cerrors.NewFatalError(err)
+
+	if fatalErr.Error() != err.Error() {
```
maha-hajja, line 27: we should use `"github.com/matryer/is"` for asserting tests instead

pkg/foundation/cerrors/fatal_test.go
```diff
@@ -0,0 +1,78 @@
...
+		want bool
+	}{
+		{
+			name: "FatalError",
+			err:  cerrors.NewFatalError(err),
+			want: true,
+		},
+		{
+			name: "No Fatal Error",
+			err:  cerrors.New("test error"),
```
maha-hajja, line 47:
```suggestion
			err:  err,
```

pkg/foundation/cerrors/fatal_test.go
```diff
@@ -0,0 +1,78 @@
...
+			want: false,
+		},
+	}
+
+	for _, tc := range testCases {
+		t.Run(tc.name, func(t *testing.T) {
+			got := cerrors.IsFatalError(tc.err)
+			if got != tc.want {
```
maha-hajja, line 55:
```suggestion
			is.Equal(got, tc.want)
```

pkg/foundation/cerrors/fatal_test.go
maha-hajja: same for all the tests please

pkg/foundation/cerrors/fatal_test.go
```diff
@@ -0,0 +1,78 @@
...
+				t.Errorf("IsFatalError(%v) = %v; want %v", tc.err, got, tc.want)
+			}
+		})
+	}
+}
+
+func TestUnwrap(t *testing.T) {
+	err := cerrors.New("test error")
+	fatalErr := cerrors.NewFatalError(err)
```
maha-hajja, line 64: notices that all the tests are using `cerrors.New` as the input for `NewFatalError`, we should maybe have one case where it's a normal error, cuz not all error we're gonna have are `cerrors`

pkg/foundation/cerrors/fatal.go
```diff
@@ -0,0 +1,41 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package cerrors
+
+// FatalError is an error type that will diferentiate these from other errors that could be retried.
```
maha-hajja, line 17:
```suggestion
// FatalError is an error type that will differentiate these from other errors that could be retried.
```

pkg/foundation/cerrors/fatal.go
```diff
@@ -0,0 +1,41 @@
...
+}
+
+// Error returns the error message.
+func (f *FatalError) Error() string {
+	return f.Err.Error()
+}
+
+// IsFatalError checks if the error is a FatalError.
+func IsFatalError(err error) bool {
+	_, ok := err.(*FatalError)
```
maha-hajja, line 39: after some research, this check wouldn't work if the fatalError is wrapped, something like 
> err := fmt.Errorf("something went wrong: %w", NewFatalError(errors.New("fatal error")))

will return false for this method.
looks like `errors.As` should be used instead, which is just `As` in this package

pkg/foundation/cerrors/fatal.go
```diff
@@ -0,0 +1,41 @@
...
+
+package cerrors
+
+// FatalError is an error type that will diferentiate these from other errors that could be retried.
+type FatalError struct {
+	Err error
+}
+
+// NewFatalError creates a new FatalError.
+func NewFatalError(err error) *FatalError {
```
maha-hajja, line 23: I believe the go convention for errors is just having `FatalError` instead of `NewFatalError`, I also noticed the main issue has it as `FatalError` too, maybe think of a way we could accomplish this? maybe renaming the struct to `FatalErr` instead? not sure of the right way to do so but it's worth the thought

raulb: I started defining `FatalError` as the constructor, but I wasn't in love with any name I gave to the struct type. I also noticed we _break_ that convention in places such as:

- https://github.com/ConduitIO/conduit/blob/06bd6e58cb1ff17841319897109dfade369a5ddd/pkg/plugin/connector/mock/plugin.go#L33
- https://github.com/ConduitIO/conduit/blob/76c25f1d1c140e65164e5ec5147a0fbe5d7e2830/pkg/plugin/connector/service.go#L62
- https://github.com/ConduitIO/conduit/blob/817d52c3184f33e7b4b542afc8b794196a68c420/pkg/conduit/runtime.go#L109

And a long etc.

I'll make the change however to meet the criteria of the issue, as you mentioned, but I'd be curious to see what's the real convention around this.

raulb: Looks like the type actually needs to end as `Error`: https://github.com/ConduitIO/conduit/actions/runs/10613375269/job/29417037570?pr=1811#step:5:25.

I think I'll change it back the way it was before.

pkg/foundation/cerrors/fatal_test.go
raulb: @maha-hajja That was my thinking as well, and I tried it, but it seems we can't use that pkg at all. See https://github.com/ConduitIO/conduit/actions/runs/10596314041/job/29364014509#step:5:44.

pkg/foundation/cerrors/fatal.go
raulb: @maha-hajja in a second thought, I made [this modification](https://github.com/ConduitIO/conduit/pull/1811/commits/44652268b6a635c1a20a965ab61977c6ac1bf075). Let me know what you think.

pkg/foundation/cerrors/fatal.go
raulb: Same as with https://github.com/ConduitIO/conduit/pull/1811#discussion_r1735998845, this would require importing the `errors` package which is not allowed:

```
pkg/foundation/cerrors/fatal.go:18:2: import 'errors' is not allowed from list 'main': To accommodate for custom error creation and handling use Conduit's 'cerrors' package instead. (depguard)
	"errors"
```

I'll check what else I can do.

pkg/foundation/cerrors/fatal.go
raulb: Interesting. It seems like this is still required:

```
pkg/foundation/cerrors/fatal.go:22:6: the type name `Fatal` should conform to the `XxxError` format (errname)
```

pkg/foundation/cerrors/fatal.go
raulb: Oh! I misunderstood what you said @maha-hajja 🤦 . I see that `As` is actually defined in the `cerrors` pkg, and that's exactly what you meant! 😓 . I'll make that change. Thanks for the pointer.

raulb: @maha-hajja pushed a fix https://github.com/ConduitIO/conduit/pull/1811/commits/e812e1c64655e000cc6df350c3f05c7fb59d88aa

raulb:
@maha-hajja I address all feedback I could apply, however there are still two open comments I responded to. Let me know what you think about those. Thanks for taking a look at this!

pkg/foundation/cerrors/fatal_test.go
maha-hajja: we could use `fmt.Errorf` instead, doesn't have to be the `errors` package, that should work right?

pkg/foundation/cerrors/fatal.go
```diff
@@ -0,0 +1,51 @@
...
+	return &FatalError{Err: err}
+}
+
+// Unwrap returns the wrapped error.
+func (f *FatalError) Unwrap() error {
+	if f == nil {
```
maha-hajja, line 33: I don't think we need this, `f` (which is the receiver for this method) can't be `nil` when the method is called

pkg/foundation/cerrors/fatal.go
maha-hajja: thank you for trying! ok let's keep it as `NewFatalError`, since there's no pretty way to do that now

pkg/foundation/cerrors/fatal.go
raulb: @maha-hajja d'oh, you're right 🤦🏻‍♂️

pkg/foundation/cerrors/fatal_test.go
raulb: @maha-hajja absolutely! I actually did that after I had left my comment 😅. You can find it [here](https://github.com/ConduitIO/conduit/blob/e812e1c64655e000cc6df350c3f05c7fb59d88aa/pkg/foundation/cerrors/fatal_test.go#L51).

raulb:
@maha-hajja addressed feedback again! Please take another look 🙏🏻

pkg/foundation/cerrors/fatal.go
```diff
@@ -0,0 +1,48 @@
...
+// limitations under the License.
+
+package cerrors
+
+import (
+	"fmt"
+)
+
+// FatalError is an error type that will differentiate these from other errors that could be retried.
+type FatalError struct {
```
hariso, line 22: It looks like we can make this type unexported, similar to `errors.New` (that returns an `*errorString`) or `fmt.Errorf` (that returns an `*wrapError`).

raulb: @hariso good point.

hariso (approved):
:zap:

pkg/foundation/cerrors/fatal_test.go
raulb: @maha-hajja I'm shipping this PR, because I'm assuming your only comment left was already addressed, but let me know if that wasn't the case, and I'll open a new PR.

pkg/foundation/cerrors/fatal.go
```diff
@@ -0,0 +1,48 @@
...
+	"fmt"
+)
+
+// fatalError is an error type that will differentiate these from other errors that could be retried.
+type fatalError struct {
+	Err error
+}
+
+// NewFatalError creates a new fatalError.
+func NewFatalError(err error) *fatalError {
```
maha-hajja, line 27: @raulb since the struct is not exported now, we can have this method as `FatalError` 👀

raulb: @maha-hajja LOL, you're right. I just opened https://github.com/ConduitIO/conduit/pull/1817.

pkg/foundation/cerrors/fatal.go
lovromazgon: FYI, the receiver can actually be `nil` in Go if the method is called on a `nil` value 😉  https://go.dev/play/p/M1-eMFxvJgT

PR #1828, opened 2024-09-04
[Recovery] Metrics

### LINKED ISSUE #1783: [Recovery] Metrics
Add metrics regarding pipeline recovery:

- `conduit_pipeline_status` should be a [gauge](https://prometheus.io/docs/concepts/metric_types/#gauge) that's set every time a pipeline is created and every time its status changes. The gauge value should always be set to 1. When the pipeline gets deleted, the gauge with the corresponding `pipeline_name` should be [deleted](https://pkg.go.dev/github.com/prometheus/client_golang/prometheus#MetricVec.Delete). It should contain two labels:
  - `status` holds the status of the pipeline (e.g. `running`, `stopped`, `degraded`, `recovering`)
  - `pipeline_name` holds the pipeline name (note that we use the pipeline name, because the API generates a random pipeline ID, the user is only in charge of the pipeline name, this might change in the future)
- `pipeline_recovering_count` should be a [counter](https://prometheus.io/docs/concepts/metric_types/#counter) which is created and set to 0 when a pipeline is created. The counter should increase every time the recovery mechanism is triggered (before the backoff retry starts waiting). It should contain one label:
  - `pipeline_name` holds the pipeline name

hariso:
### Description

Closes #1783.

### Quick checks


pkg/foundation/metrics/measure/measure.go
```diff
@@ -28,6 +28,16 @@ var (
 	PipelinesGauge = metrics.NewLabeledGauge("conduit_pipelines",
 		"Number of pipelines by status.",
 		[]string{"status"})
+	PipelineStatusGauge = metrics.NewLabeledGauge(
+		"conduit_pipeline_status",
+		"Pipeline statuses.",
+		[]string{"pipeline_name", "status"},
+	)
+	PipelineRecoveringCount = metrics.NewLabeledCounter(
```
hariso, line 36: @raulb This is relevant to your work.

pkg/foundation/metrics/measure/measure.go
```diff
@@ -28,6 +28,16 @@ var (
 	PipelinesGauge = metrics.NewLabeledGauge("conduit_pipelines",
 		"Number of pipelines by status.",
 		[]string{"status"})
+	PipelineStatusGauge = metrics.NewLabeledGauge(
```
hariso, line 31: Note: I considered another approach, where the value would be the integer value of status itself, e.g. `conduit_pipeline_status{pipeline_name="testpipeline1"} 3` (where for example 3=degraded). While that's a bit more readable and meaningful (reads as "the status of pipeline testpipeline1 is '3') than `conduit_pipeline_status{pipeline_name="pipeline1",status="degraded"} 1` (the gauge value '1' has no meaning), the integer value spoils it all.:)

lovromazgon: Hmm, interesting, that could work if we describe the meaning of the integers in the description of the metric. The whole problem is that Prometheus doesn't allow string values for metrics, so anything we do here is a workaround, but yours might be easier to handle when visualizing the metrics in Grafana or other tools.

hariso: But it would still be numbers when visualizing, right? Unless some custom mapping is done. And we would need to be careful when adding new statuses and/or refactoring.

pkg/foundation/metrics/measure/measure.go
```diff
@@ -28,6 +28,16 @@ var (
...
+		"conduit_pipeline_status",
+		"Pipeline statuses.",
+		[]string{"pipeline_name", "status"},
+	)
+	PipelineRecoveringCount = metrics.NewLabeledCounter(
+		"pipeline_recovering_count",
+		"Number of times pipelines have been recovering (by by pipeline name)",
```
raulb, line 38: This would refer to a specific pipeline -by name- (and not pipelines, right?). Also fixing a double `by`

```suggestion
		"Number of times a pipeline have been recovering (by pipeline name)",
```

pkg/foundation/metrics/metrics.go
```diff
@@ -68,7 +68,12 @@ type Gauge interface {
 
 // LabeledGauge describes a gauge that must have values populated before use.
 type LabeledGauge interface {
+	// WithValues returns the Gauge for the given slice of label
+	// values (same order as the label names used when creating this LabeledGauge).
+	// If that combination of label values is accessed for the first time,
+	// a new Gauge is created.
 	WithValues(labels ...string) Gauge
+	DeleteLabels(labels ...string) bool
```
raulb, line 76: I think we should document this method here (like `WithValues`) or below.

pkg/foundation/metrics/prometheus/gauge.go
```diff
@@ -27,6 +27,10 @@ func (lg *labeledGauge) WithValues(labels ...string) metrics.Gauge {
 	return &gauge{pg: lg.pg.WithLabelValues(labels...)}
 }
 
+func (lg *labeledGauge) DeleteLabels(labels ...string) bool {
+	return lg.pg.DeleteLabelValues(labels...)
```
raulb, line 31: could `pg` be nil?

pkg/pipeline/service.go
```diff
@@ -363,3 +366,15 @@ func (s *Service) validatePipeline(cfg Config, id string) error {
...
+	status := strings.ToLower(pl.GetStatus().String())
+	measure.PipelinesGauge.WithValues(status).Dec()
+	measure.PipelineStatusGauge.DeleteLabels(pl.Config.Name, status)
+}
+
+func (s *Service) updateNewStatusMetrics(pl *Instance) {
+	status := strings.ToLower(pl.GetStatus().String())
+	measure.PipelinesGauge.WithValues(status).Inc()
+	measure.PipelineStatusGauge.WithValues(pl.Config.Name, status).Set(1)
+}
```
raulb, line 380: These two refer to `before` and `after`?

pkg/foundation/metrics/measure/measure.go
hariso: I looked a bit into how the data can be used within Grafana, for visualizations and alerting. For visualization, I built a simple table dashboard showing a pipeline's current and/or historical statuses.
* Option with the metric value representing the integer status (e.g. `conduit_pipeline_status{pipeline_name="testpipeline1"} 3` (where for example 3=degraded)
  * dashboard: needs value mappings, that can be done with field overrides (5 of them)
  * alert: same, the alert checks if the value of `conduit_pipeline_status{pipeline_name="testpipeline1"}`
* Option with the status being set as a label name (e.g. `conduit_pipeline_status{pipeline_name="pipeline1",status="degraded"} 1`)
  * dashboard: nothing special needs to be done. However, other visualizations (such as charts) could be more difficult to build, because the actual status needs to be taken from a label.
  * alert: the alert checks if the value for `conduit_pipeline_status{pipeline_name="pipeline1",status="degraded"}` is above 0. This is counter-intuitive since the actual value plays no role (we're checking for the mere presence of the metric).

I'm still a bit on the fence, because of the value mappings for option 1, but overall it _feels_ like it's more in the spirit of Prometheus.

pkg/pipeline/service.go
hariso: Exactly!

pkg/foundation/metrics/measure/measure.go
hariso: We discussed this on today's standup, we're going with option 1 (the gauge value is the pipeline's status as integer).

pkg/pipeline/service.go
```diff
@@ -363,3 +366,14 @@ func (s *Service) validatePipeline(cfg Config, id string) error {
 
 	return cerrors.Join(errs...)
 }
+
+func (s *Service) updateOldStatusMetrics(pl *Instance) {
+	status := strings.ToLower(pl.GetStatus().String())
+	measure.PipelinesGauge.WithValues(status).Dec()
```
lovromazgon, line 372: Wait, shouldn't `WithValues` get `pl.Name`? And the gauge then needs to be set to the correct status value.

hariso: Not really, because this is actually this line: https://github.com/ConduitIO/conduit/pull/1828/files#diff-4d290fa258ba1313fd6cb7e6393c2efa98c5fae36e7e0d7ace77f31bc6c53557L596

It's updating the metrics for the old status value, before the new value is set and then the metrics for the new value are updated.

lovromazgon: Oh I see, that's the existing pipelines gauge, not the pipeline status gauge, my bad!

lovromazgon (approved):

PR #1835, opened 2024-09-06
refactor: Add Lifecycle service 

raulb:
### Description

This pull request creates a new lifecycle service that allows for the following:

-  Decoupling the existing pipeline service from others, such as ConnectorPlugin and Processor.
-  The main goal of this pull request is to enable the restart of a pipeline in isolation, without involving other services. This is part of the pipeline recovery work.

### Quick checks


pkg/lifecycle/service.go
```diff
@@ -154,33 +216,43 @@ func (s *Service) stopGraceful(ctx context.Context, pl *Instance, reason error)
...
+	rp.t.Kill(pipeline.ErrForceStop)
+	for _, n := range rp.n {
 		if node, ok := n.(stream.ForceStoppableNode); ok {
 			// stop all pub nodes
 			s.logger.Trace(ctx).Str(log.NodeIDField, n.ID()).Msg("force stopping node")
 			node.ForceStop(ctx)
 		}
 	}
+
+	s.runningPipelines[rp.pipeline.ID] = nil
```
lovromazgon, line 242: We should rather delete the key entirely.

```suggestion
	delete(s.runningPipelines, rp.pipeline.ID)
```

pkg/pipeline/service.go
```diff
@@ -366,3 +336,27 @@ func (s *Service) validatePipeline(cfg Config, id string) error {
 
 	return cerrors.Join(errs...)
 }
+
+// GetInstances returns all pipeline instances.
+func (s *Service) GetInstances() map[string]*Instance {
```
lovromazgon, line 341: This functionality is already provided by `List`, let's use that instead.

pkg/pipeline/service.go
```diff
@@ -366,3 +336,27 @@ func (s *Service) validatePipeline(cfg Config, id string) error {
...
+	return s.instances
+}
+
+// UpdateStatus updates the status of a pipeline by the ID.
+func (s *Service) UpdateStatus(ctx context.Context, id string, status Status, errMsg string) error {
+	pipeline, err := s.Get(ctx, id)
+	if err != nil {
+		return err
+	}
+	measure.PipelinesGauge.WithValues(strings.ToLower(pipeline.GetStatus().String())).Dec()
```
lovromazgon, line 351: Heads-up - this code changed slightly in the [metrics PR](https://github.com/ConduitIO/conduit/pull/1828), we need to update it. Also, be ready to resolve conflicts from `main` because of this.

lovromazgon (commented):
Not bad, not bad at all 👍

pkg/provisioning/test/dest-file.txt
```diff
@@ -0,0 +1 @@
+{"position":"MjU=","operation":"create","metadata":{"conduit.source.connector.id":"pipeline1:con1","file.path":"./test/source-file.txt"},"key":"MQ==","payload":{"before":null,"after":"anVzdCBzb21lIHRlc3RpbmcgZGF0YSA6RA=="}}
```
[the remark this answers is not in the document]
raulb: Yup, I also noticed we had this one too https://github.com/ConduitIO/conduit/blob/0553aaf21bd4de800ab33f2fcac0fb49f3747ea3/pkg/provisioning/test/source-file.txt.

pkg/lifecycle/service.go
```diff
@@ -681,8 +681,6 @@ func (s *Service) runPipeline(ctx context.Context, rp *runnablePipeline) error {
 		nodesWg.Wait()
 		err := rp.t.Err()
 
-		measure.PipelinesGauge.WithValues(strings.ToLower(rp.pipeline.GetStatus().String())).Dec()
```
raulb: These will be updated via `UpdateStatus`.

pkg/pipeline/service.go
raulb: @lovromazgon thank you for the heads up. I just merged and resolve their conflicts.

pkg/provisioning/test/dest-file.txt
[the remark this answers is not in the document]
lovromazgon: That one is actually needed iirc, that's used as the input in the test.

pkg/lifecycle/service.go
```diff
@@ -12,7 +12,11 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-package pipeline
+// Package lifecycle wires up everything under the hood of a Conduit instance
+// including metrics, telemetry, logging, and server construction.
+// It should only ever interact with the Orchestrator, never individual
+// services. All of that responsibility should be left to the Orchestrator.
+package lifecycle
```
hariso, line 19: The comment isn't quite accurate, it's probably been taken from somewhere else.

raulb: @hariso Good catch

pkg/pipeline/service.go
```diff
@@ -51,7 +43,6 @@ type Service struct {
 
 	instances     map[string]*Instance
 	instanceNames map[string]bool
-	handlers      []FailureHandler
 	backoffCfg    *backoff.Backoff
```
lovromazgon, line 46: This can now be removed.

pkg/lifecycle/service.go
```diff
@@ -46,25 +101,37 @@ type ProcessorService interface {
...
+
+// OnFailure registers a handler for a lifecycle.FailureEvent.
+// Only errors which happen after a pipeline has been started
+// are being sent.
+func (s *Service) OnFailure(handler FailureHandler) {
+	s.handlers = append(s.handlers, handler)
+}
+
 // Run runs pipelines that had the running state in store.
 func (s *Service) Run(
```
lovromazgon, line 124: I'd think about renaming this to `Init`. Previously this was on the `pipeline.Service` where we already had an `Init` method, so we had to find a different name. In fact, all services have an `Init` method, which initializes the service (e.g. loads stuff from the store), that's why I think this name could be a better fit.

pkg/lifecycle/service.go
```diff
@@ -119,31 +184,31 @@ func (s *Service) Start(
 // instead the context for all nodes will be canceled which causes them to stop
 // running as soon as possible.
 func (s *Service) Stop(ctx context.Context, pipelineID string, force bool) error {
-	pl, err := s.Get(ctx, pipelineID)
-	if err != nil {
-		return err
+	rp, ok := s.runningPipelines[pipelineID]
```
lovromazgon, line 187: Watch out, we need to lock the mutex before accessing `s.runningPipelines`.

pkg/lifecycle/service.go
```diff
@@ -176,11 +242,13 @@ func (s *Service) stopForceful(ctx context.Context, pl *Instance) error {
 // StopAll will ask all the pipelines to stop gracefully
 // (i.e. that existing messages get processed but not new messages get produced).
 func (s *Service) StopAll(ctx context.Context, reason error) {
-	for _, pl := range s.instances {
-		if pl.GetStatus() != StatusRunning && pl.GetStatus() != StatusRecovering {
+	instances := s.pipelines.List(ctx)
+	for _, pl := range instances {
+		rp, ok := s.runningPipelines[pl.ID]
```
lovromazgon, line 247: Again here, we need to hold the lock if we are accessing `s.runningPipelines`. Also, I'm not sure we even need to list the pipelines, we should be able to just range over the ones in `s.runningPipelines`.

pkg/lifecycle/service.go
```diff
@@ -211,49 +279,61 @@ func (s *Service) Wait(timeout time.Duration) error {
...
 	}
 }
 
 // waitInternal blocks until all pipelines are stopped and returns an error if any of
 // the pipelines failed to stop gracefully.
 func (s *Service) waitInternal() error {
 	var errs []error
-	for _, pl := range s.instances {
-		err := pl.Wait()
+	for _, rp := range s.runningPipelines {
```
lovromazgon, line 290: Missing lock 😬

pkg/lifecycle/service.go
```diff
@@ -593,69 +666,77 @@ func (s *Service) runPipeline(ctx context.Context, pl *Instance) error {
...
 		// since the source of the truth is the actual pipeline (stored in memory).
-		s.updateNewStatusMetrics(pl)
-
-		storeErr := s.store.Set(ctx, pl.ID, pl)
-		if storeErr != nil {
-			return cerrors.Errorf("pipeline not updated: %w", storeErr)
-		}
-
-		return err
+		return s.pipelines.UpdateStatus(ctx, rp.pipeline.ID, rp.pipeline.GetStatus(), "")
```
lovromazgon, line 725: This call is redundant now, since we call `UpdateStatus` above with the corresponding status, we can simply return `err` here.

pkg/lifecycle/service.go
hariso: Maybe it would be easier and safer to use a `sync.Map`?

lovromazgon: I'm not a fan, since we then need to do type assertions 😕 That said, we could provide our own safe implementation in `csync` that uses generics, but not sure it's needed in this PR.

hariso: Cool, didn't know that about types (I didn't use it myself yet). :+1: So yeah, locks be it.:)

lovromazgon: Man, nowadays with AI's help such a struct is implemented and tested insanely fast. https://github.com/ConduitIO/conduit-commons/pull/116 🙃

pkg/lifecycle/service.go
raulb: Good point. This will be a way to reutilize the runningPipelines more effectively.

pkg/lifecycle/service.go
raulb: Good catch.

pkg/lifecycle/service.go
```diff
@@ -173,18 +241,26 @@ func (s *Service) stopForceful(ctx context.Context, pl *Instance) error {
...
 
-// StopAll will ask all the pipelines to stop gracefully
+// StopAll will ask all the running pipelines to stop gracefully
 // (i.e. that existing messages get processed but not new messages get produced).
 func (s *Service) StopAll(ctx context.Context, reason error) {
-	for _, pl := range s.instances {
-		if pl.GetStatus() != StatusRunning && pl.GetStatus() != StatusRecovering {
+	s.m.Lock()
+	instances := s.runningPipelines
+	s.m.Unlock()
```
lovromazgon, line 249: This only copies the pointer to the map, so the map could still change while we iterate over it. We'll need to hold the lock while we iterate over it (an alternative would be to make a copy of the map, but I don't think that's needed):
```suggestion
	s.m.Lock()
	defer s.m.Unlock()
```

pkg/lifecycle/service.go
```diff
@@ -173,18 +241,26 @@ func (s *Service) stopForceful(ctx context.Context, pl *Instance) error {
...
+
+	for _, rp := range instances {
+		p := rp.pipeline
+		if p.GetStatus() != pipeline.StatusRunning && p.GetStatus() != pipeline.StatusRecovering {
+			s.m.Lock()
+			delete(s.runningPipelines, p.ID)
+			s.m.Unlock()
```
lovromazgon, line 256: Wait, why are we deleting the pipeline here? 🤔 This should happen in the cleanup goroutine.

pkg/lifecycle/service.go
```diff
@@ -211,49 +287,66 @@ func (s *Service) Wait(timeout time.Duration) error {
...
 // waitInternal blocks until all pipelines are stopped and returns an error if any of
 // the pipelines failed to stop gracefully.
 func (s *Service) waitInternal() error {
 	var errs []error
-	for _, pl := range s.instances {
-		err := pl.Wait()
+
+	s.m.Lock()
+	pipelines := s.runningPipelines
+	s.m.Unlock()
```
lovromazgon, line 301: Here we have a similar issue, although here we can't really hold the lock, since stopping a pipeline also requires the lock, and we are waiting for pipelines to stop. So here we would actually need to make a copy in advance so we can safely iterate over all the pipelines.

This would indeed be easier if we used the map from [this PR](https://github.com/ConduitIO/conduit-commons/pull/116), as we could simply do `s.runningPipelines.Copy()`.

pkg/lifecycle/service.go
raulb: Oh, shoot, you're right. My thinking was that, since we're iterating over the running pipelines, if it's stopped, then, it shouldn't be part of the running pipelines at all, but I didn't realize that this could also be part of a different goroutine that maybe still need be cleaned up.

pkg/lifecycle/service.go
raulb: @lovromazgon https://github.com/ConduitIO/conduit/pull/1835/commits/3596fb82428f71d8d990aed272d2407fd2344157

pkg/lifecycle/service.go
raulb: https://github.com/ConduitIO/conduit/pull/1835/commits/3596fb82428f71d8d990aed272d2407fd2344157

lovromazgon (approved):

PR #1837, opened 2024-09-09
Use own registry for the stats handler

lyuboxa:
### Description

Use own registry for the stats handler to create separation between the global/default prom client registry.
Conduit metrics are already using a separate registry. 

When testing Conduit embedded, `..MustRegister()` may fail due to the stats handler already registered in the global registry. 

Screenshot 2024-09-09 at 13 28 07

### Quick checks


lyuboxa:
Closing, superseded by #1842

Issue #1781, opened 2024-08-14
[Recovery] Recovery mechanism


lovromazgon:
Implement a recovery mechanism (probably around [here](https://github.com/ConduitIO/conduit/blob/d4e617775fb05e20261389da31a7506566e8b47b/pkg/pipeline/lifecycle.go#L600-L602)) that checks if the pipeline errored out and restarts the pipeline after the backoff delay, unless the returned error is a fatal error.

Make sure to listen if the context gets cancelled so a pipeline can stop even while waiting for the backoff retry. Hint: see how we do it in the [http processor](https://github.com/ConduitIO/conduit/blob/d4e617775fb05e20261389da31a7506566e8b47b/pkg/plugin/processor/builtin/impl/webhook/http.go#L222-L236).

Also be careful because the tomb in `pipeline.Instance` will be alive and will [prevent](https://github.com/ConduitIO/conduit/blob/d4e617775fb05e20261389da31a7506566e8b47b/pkg/pipeline/lifecycle.go#L534-L536) the pipeline to be restarted until it's dead.

PR #1853, opened 2024-09-16
feat: restart pipeline with recovery

raulb:
### Description

Fixes https://github.com/ConduitIO/conduit/issues/1781 relying on its [design document](https://github.com/ConduitIO/conduit/blob/main/docs/design-documents/20240812-recover-from-pipeline-errors.md).

### Quick checks


pkg/lifecycle/service.go
```diff
@@ -176,6 +185,77 @@ func (s *Service) Start(
...
+	timer := time.NewTimer(duration)
+	select {
+	case <-ctx.Done():
+		return ctx.Err()
+	case <-time.After(duration):
+		<-timer.C
+	}
+
+	// Get status of pipeline, if it recovers (it's running),
+	pl, err := s.pipelines.Get(ctx, rp.pipeline.ID)
```
lovromazgon, line 241: Btw, `rp.pipeline` is already the pipeline, no need to fetch it again through the service 😉

pkg/lifecycle/service.go
```diff
@@ -176,6 +185,77 @@ func (s *Service) Start(
 	return nil
 }
 
+// Restart stops an existing pipeline and replaces their nodes.
+func (s *Service) Restart(ctx context.Context, rp *runnablePipeline) error {
+	rp, ok := s.runningPipelines.Get(rp.pipeline.ID)
```
lovromazgon, line 190: Also here, `rp` is the runnable pipeline, I don't think we need to fetch it again.

raulb: Oops, that's a leftover from when I was using a pipeline instance. Thanks for the catch.

raulb:
Created this separate issue to tackle what's remaining on this project https://github.com/ConduitIO/conduit/issues/1862. Without that, we're setting the pipeline back to running as soon as we restarted the pipeline.

pkg/conduit/config_test.go
```diff
@@ -196,12 +197,12 @@ func TestConfig_Validate(t *testing.T) {
...
 		{
-			name: "error recovery: negative max-retries",
+			name: "error recovery: max-retries smaller than -1",
 			setupConfig: func(c Config) Config {
-				c.Pipelines.ErrorRecovery.MaxRetries = -1
+				c.Pipelines.ErrorRecovery.MaxRetries = lifecycle.InfiniteRetriesErrRecovery - 1
 				return c
 			},
-			want: cerrors.New(`invalid error recovery config: "max-retries" config value mustn't be negative (got: -1)`),
+			want: cerrors.New(`invalid error recovery config: "max-retries" can't be smaller than -1 (infinite retries)`),
```
hariso, line 205: Nitpick: would be good to have a test for 0 retries (to make sure that's an accepted and valid option).

pkg/conduit/entrypoint.go
```diff
@@ -71,6 +71,16 @@ func (e *Entrypoint) Serve(cfg Config) {
 	}
 }
 
+func deprecatedFlag(name string) bool {
```
hariso, line 74: Nitpick: maybe you can turn this into a map, and just use the map:
```go
deprecatedFlags := map[string]bool{
    "pipelines.exit-on-error": true
}
if deprecatedFlags["something"] {
    ...
}
```

raulb: @hariso good idea!

pkg/lifecycle/service.go
```diff
@@ -176,6 +187,69 @@ func (s *Service) Start(
...
+	nodes, err := s.buildNodes(ctx, rp.pipeline)
+	if err != nil {
+		return cerrors.Errorf("could not build new nodes for pipeline %s: %w", rp.pipeline.ID, err)
+	}
+
+	// Replaces the old nodes with the new ones.
+	rp.n = nodes
+
+	// clears out the tomb
+	rp.t.Kill(cerrors.New("restarting pipeline"))
```
hariso, line 204: Shouldn't the tomb already be dead here, because the pipeline failed?

pkg/lifecycle/service.go
```diff
@@ -176,6 +187,69 @@ func (s *Service) Start(
 	return nil
 }
 
+// Restart replaces the nodes of a pipeline and starts it again.
+func (s *Service) Restart(ctx context.Context, rp *runnablePipeline) error {
```
hariso, line 191: Thinking aloud... `Restart()` should actually be `Start()`, but with the proper backoff config. It would also be good that a restart starts with a clean slate, which is why I think it's better to just create a new `runningPipeline` then fix the existing one. It's also safer if we add other fields to `runningPipeline` in future.

`Restart()` can call `Start()` and give it the existing `runningPipeline`'s backoff config. Or `Start()` can check if a pipeline is already running, and copy over the backoff config.

Not a must have for this PR, I think it already looks great and we're ready to merge it!

raulb: I'm giving this one a try and I'm thinking that this is not ideal. In the start method, if the pipeline is running, it will return an error saying we can't start it again. So, if the start method returns that error, the retries will be stopped. We could put it into a degraded state if that was the last attempt

So, in other words, even if when trying to recover it, we return an error when we shouldn't, running should be a good thing. The runnablePipeline also contains the backoff configuration would also require some refactoring so we preserve the backoff altogether.

pkg/lifecycle/service.go
```diff
@@ -176,6 +211,44 @@ func (s *Service) Start(
 	return nil
 }
 
+// StartWithBackoff starts a pipeline with a backoff.
+// It'll check the number of times the pipeline has been restarted and the duration of the backoff.
+// When the pipeline has reached out the maximum number of retries, it'll return a fatal error.
+func (s *Service) StartWithBackoff(ctx context.Context, rp *runnablePipeline) error {
```
hariso, line 217: @raulb WDYT about just calling it `Restart`?

pkg/lifecycle/service.go
```diff
@@ -176,6 +211,44 @@ func (s *Service) Start(
...
+	s.logger.Info(ctx).Str(log.PipelineIDField, rp.pipeline.ID).Msg("restarting with backoff")
```
hariso, line 218: @raulb FYI, changed this, because a restart might be a significant event. I'm open to changing back if you think it's too high of a logging level.

pkg/lifecycle/service.go
```diff
@@ -226,7 +299,9 @@ func (s *Service) stopForceful(ctx context.Context, rp *runnablePipeline) error
 		Str(log.PipelineIDField, rp.pipeline.ID).
 		Any(log.PipelineStatusField, rp.pipeline.GetStatus()).
 		Msg("force stopping pipeline")
-	rp.t.Kill(pipeline.ErrForceStop)
+
+	// Creates a FatalError to prevent the pipeline from recovering.
+	rp.t.Kill(cerrors.FatalError(pipeline.ErrForceStop))
```
hariso, line 304: I don't think this is needed anymore, right?

pkg/lifecycle/service.go
raulb: @hariso That could work

pkg/lifecycle/service.go
raulb: @hariso makes sense!

pkg/lifecycle/service.go
raulb: @hariso actually, ..., in my opinion, `Restart` would be `Stop` + `Start`. This method doesn't stop, but mainly start with an additional backoff so I think this method's name works.

hariso: Ah, you're right, StartWithBackoff definitely makes more sense now that you said that.:)

pkg/lifecycle/service_test.go
```diff
@@ -250,24 +248,24 @@ func TestServiceLifecycle_PipelineSuccess(t *testing.T) {
 	// create mocked connectors
 	ctrl := gomock.NewController(t)
 	wantRecords := generateRecords(10)
-	source, sourceDispenser := generatorSource(ctrl, persister, wantRecords, nil, true)
-	destination, destDispenser := asserterDestination(ctrl, persister, wantRecords)
-	dlq, dlqDispenser := asserterDestination(ctrl, persister, nil)
+	source, srcDispenser := asserterSource(ctrl, persister, wantRecords, nil, true, 1)
```
hariso, line 251: Is there a special reason for changing to asserterSource?

hariso (approved):
:muscle:

pkg/lifecycle/service_test.go
hariso: I was wondering if the behavior was different, but looks like not :shrug: . Looks good to me then!

PR #1856, opened 2024-09-16
Upgrade Postgres connector (includes schema support)

hariso:
### Quick checks


lovromazgon:
Any specific reason we want to do this now, and not when the Postgres connector is properly released?

hariso:
> Any specific reason we want to do this now, and not when the Postgres connector is properly released?

Yeah, it was to have it ready for the Platform team to try it out. But since this was 2 weeks and now we're about to release Conduit, it doesn't actually make much sense.:)

PR #1877, opened 2024-10-04
 Pipeline recovery: add test case 

hariso:
### Description

Adds: Recovery not triggered - user force-stopped pipeline.

### Quick checks


raulb:
@hariso sorry I missed this one. Do you think it's still relevant?

hariso:
@raulb Yes, I think we should always be sure that a pipeline that a user force-stopped shouldn't be recovered.

raulb (approved):

PR #1880, opened 2024-10-04
feat: checks `HealthyAfter` to set pipeline status to running

### LINKED ISSUE #1862: [Recovery] Add delay after which fail count is reset
### Feature description

As described in this [design document](https://github.com/ConduitIO/conduit/blob/main/docs/design-documents/20240812-recover-from-pipeline-errors.md), to control intermittent errors, the recovery mechanism will have a parameter named `HealthyAfter` that  

[This pull-request](https://github.com/ConduitIO/conduit/pull/1853) implemented the restart of a pipeline to be recovered, but it was considered tackling `HealthyAfter` separately considering it'd increase its scope notably.

## Some details

- The `backoff` will be reset only when a pipeline processes at least one record or runs without encountering an error for a predefined time frame.
- Each pipeline will control its own `HealthyAfter` duration https://github.com/ConduitIO/conduit/blob/raul/restart-with-recovery/pkg/lifecycle/service.go#L55.
- This will need to be updated once we have this delay in place https://github.com/ConduitIO/conduit/blob/raul/restart-with-recovery/pkg/lifecycle/service.go#L752-L756.

## Possible solution

Spawn a goroutine that resets the backoff after a certain time.

raulb:
### Description

Fixes #1862 

### Quick checks

### DIFF
--- a/pkg/lifecycle/service.go
+++ b/pkg/lifecycle/service.go
@@ -745,14 +745,12 @@ func (s *Service) runPipeline(ctx context.Context, rp *runnablePipeline) error {
 		})
 	}
 
-	// TODO: When it's recovering, we should only update the status back to running once HealthyAfter has passed.
-	// now:
-	//		running -> (error) -> recovering (restart) -> running
-	// future (with the HealthyAfter mechanism):
-	//		running -> (error) -> recovering (restart) -> recovering (wait for HealthyAfter) -> running
-	err := s.pipelines.UpdateStatus(ctx, rp.pipeline.ID, pipeline.StatusRunning, "")
-	if err != nil {
-		return err
+	// If we're not trying to recover it, we update the status to running
+	if rp.backoffCfg.Attempt() == 0 {
+		err := s.pipelines.UpdateStatus(ctx, rp.pipeline.ID, pipeline.StatusRunning, "")
+		if err != nil {
+			return err
+		}
 	}
 
 	// cleanup function updates the metrics and pipeline status once all nodes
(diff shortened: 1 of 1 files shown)


lovromazgon:
IMO We should approach this differently, let me explain.

Every time we recover, that recovery attempt should be "active" for a certain time period, let's call it `HealthyAfter` (in the end I'll explain why I think we should use a different name). We would track the number of "active" recovery attempts and set a threshold on the number of recovery attempts `MaxRetries`. If we go over that threshold, the pipeline doesn't recover anymore, but rather goes into the degraded state.

The resulting behavior is that you can define a time window (`HealthyAfter`) in which you can tolerate a certain number of recovery attempts (`MaxRetries`).

For example, imagine the following scenario and notice how the number of "active" recovery attempts changes over time:

- Conduit runs a pipeline, `HealthyAfter` is set to 5 minutes, `MaxRetries` is set to 2.
- The pipeline encounters an error at some point in time.
  - Active recovery attempts increases to 1.
- The pipeline recovers, runs for 2 minutes, then fails again.
  - Active recovery attempts increases to 2.
- The pipeline recovers, runs fine.
- After 3 more minutes it has been 5 minutes since the first recovery attempt.
  - Active recovery attempts decreases to 1.
- After 1 minute the pipeline fails again.
  - Active recovery attempts increases to 2.
- It tries to recover but fails right away.
  - Active recovery attempts increases to 3.
- This time the pipeline doesn't try to recover, because the number of attempts just exceeded 2, pipeline gets degraded.

In code this could be done using a field on `runnablePipeline` of type `atomic.Int32` that is increased every time a recovery is started, and a [`time.AfterFunc`](https://pkg.go.dev/time#AfterFunc) call that decreases the integer after `HealthyAfter`.

Now, regarding the name `HealthyAfter`. If we go with this algorithm, I would propose renaming this to something that incldues the word "window", since it's essentially defining the time window in which the number of recovery attempts `MaxRetries` can't be exceeded. It's the same behavior that we use in the DLQ (see [animation](https://conduit.io/docs/features/dead-letter-queue#nack-window-and-threshold)), except that the window in the DLQ is defined using a number of records, while here the window is defined by a time duration.

Let me know what you think and if anything is unclear.

lovromazgon:
I opened a PR here that should do what I described above https://github.com/ConduitIO/conduit/pull/1881.

raulb:
Closing in favor of https://github.com/ConduitIO/conduit/pull/1881

PR #1881, opened 2024-10-04
Track recovery attempts using leaky bucket

### LINKED ISSUE #1862: [Recovery] Add delay after which fail count is reset

lovromazgon:
### Description

This PR implements tracking of recovery attempts using a [leaky bucket](https://en.wikipedia.org/wiki/Leaky_bucket). Once the bucket is full (i.e. we exceed the `MaxRetries` threshold) the pipeline is degraded instead of recovered.

Fixes #1862 

### Quick checks


raulb (approved):

pkg/lifecycle/service.go
```diff
@@ -220,7 +220,12 @@ func (s *Service) StartWithBackoff(ctx context.Context, rp *runnablePipeline) er
 		Int64(log.AttemptField, attempt).
 		Msg("restarting with backoff")
 
-	time.AfterFunc(duration+s.errRecoveryCfg.HealthyAfter, func() {
+	time.AfterFunc(duration+s.errRecoveryCfg.MaxRetriesWindow, func() {
+		s.logger.Info(ctx).
```
lovromazgon, line 224: Maybe this should be a debug log?

lovromazgon (commented):
LGTM, I'd approve if I could, but I can't because I'm officially the author. Posted a nitpick, but that's about it 👍

pkg/lifecycle/service.go
raulb: @lovromazgon I questioned myself about it too, so I'll change it to debug.

PR #1904, opened 2024-10-10
 Update version in README.md 

hariso:
### Quick checks

### DIFF
--- a/docs/releases.md
+++ b/docs/releases.md
@@ -57,10 +57,14 @@ the same directory.
 ## Documentation
 
 1. Write a blog post.
-2. Regenerate processor documentation on [`conduit-site`](https://github.com/ConduitIO/conduit-site).
-3. Update banner on the web-site ([example](https://github.com/ConduitIO/conduit-site/pull/47/files#diff-cc8abb6104e21d495dc8f64639c7b03419226d920d1c545df51be9b0b73b2784)).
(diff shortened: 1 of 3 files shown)


docs/releases.md
```diff
@@ -58,7 +58,12 @@ the same directory.
 
 1. Write a blog post.
 2. Regenerate processor documentation on [`conduit-site`](https://github.com/ConduitIO/conduit-site).
-3. Update banner on the web-site ([example](https://github.com/ConduitIO/conduit-site/pull/47/files#diff-cc8abb6104e21d495dc8f64639c7b03419226d920d1c545df51be9b0b73b2784)).
+3. Update banner on the
+   web-site ([example](https://github.com/ConduitIO/conduit-site/pull/47/files#diff-cc8abb6104e21d495dc8f64639c7b03419226d920d1c545df51be9b0b73b2784)).
+4. Search for the latest version in `conduit-site` and replace with new
+   version (e.g. search for 0.11.1 and replace with 0.12.0)
```
maha-hajja, line 64: nitpick: sometimes we use 0.11 instead of 0.11.0 in the docs, so it's better to search for that

maha-hajja (approved):

docs/releases.md
hariso: Makes sense!

PR #1927, opened 2024-10-24
Basic `conduit init` and `conduit pipelines init` commands

### LINKED ISSUE #1924: A basic `conduit init` command
The purpose of this command is to provide a better onboarding experience.

This should be a basic command that makes it possible to:

- create a demo pipeline with `conduit init`
- create a pipeline with a given source, destination, and possibly a processor, e.g. `conduit init --source SRC_PLUGIN --destination DEST_PLUGIN --processor PROC_PLUGIN`

This version is not part of #1911 and some parts of it will be rewritten to use the new CLI code. The logic (that generates a pipeline given a source, destination, etc.) should be re-usable.

hariso:
### Description

Closes #1924.

High-level overview of changes:

* There's a new package `cli` with the CLI-related code.
* `main.go` now invokes the CLI.
* The CLI adds 2 new commands, `init` and `pipelines`, and one sub-command, `pipelines init`.
* Running the server has been moved to the root command. The existing flags for Conduit have been moved to the root command.
* The default value of `--pipelines.path` in `conduit pipelines init` is now `./pipelines`. This makes it easier to get started with Conduit (`conduit init && conduit pipelines init`, without specifying the path). IMO, it makes sense also because the pipelines' directory is relative to the binary/working directory.

Tested commands:

```shell
./conduit

./conduit --http.address :8888

./conduit init

./conduit init --config.path /tmp/conduit
tree /tmp/conduit

./conduit pipelines init
cat pipeline-*yaml

./conduit pipelines init --source postgres
cat pipeline-*yaml

./conduit pipelines init --destination kafka
cat pipeline-*yaml

./conduit pipelines init my-pipeline --source postgres --destination kafka --pipelines.path /tmp/conduit-pipelines/
cat /tmp/conduit-pipelines/pipeline-my-pipeline.yaml

./conduit help

./conduit --version

./conduit --invalid
```

### Quick checks


raulb:
@hariso I think `init` is ypically used to set up the initial structure, configuration, and dependencies for a new project or application. Here are some industry examples:

- `npm init` The npm init command is used to set up a new or existing npm package
- `tsc --init` Initializes a TypeScript project and creates a tsconfig.json file.
- `git init`: Initializes a new Git repository in the current directory, creating the necessary files and directories to start version controlling your project.
- `webpack init`: Helps you set up a new Webpack configuration by asking you a series of questions and generating a basic Webpack configuration file (`webpack.config.js`).
- `vue init`: Generates a new Vue.js project based on a specific template. It prompts the user with questions to customize the project setup.
- `terraform init`: Initializes a new Terraform working directory, creating the necessary files and directories to start managing infrastructure as code.
- `yarn init`: Initializes a new Yarn project, creating a `package.json` file and setting up the initial project structure.
- `kubectl init`: Initializes a new Kubernetes cluster, setting up the necessary components and configuration files.

By having `--build-pipeline` I also see the disadvantage of not distinguishing the command with flags. `--build-pipeline` as it is now looks like a global flag for the `conduit` CLI. 

For that reason, I feel strongly opinionated that the commands needs to be `init` and nothing else.

hariso:
We discussed this at our daily sync, here's the summary
* `conduit init` will initialize a "working directory" for Conduit, i.e. create `connectors`, `processors` and `pipelines` directory, and the `conduit.yaml` config gile
* once `conduit init` is done, it will inform the user to run `conduit pipelines init`, to initialize a pipeline.
* `conduit pipelines init` does what we already mentioned.

raulb:
> We discussed this at our daily sync, here's the summary
> 
> * `conduit init` will initialize a "working directory" for Conduit, i.e. create `connectors`, `processors` and `pipelines` directory, and the `conduit.yaml` config gile
> * once `conduit init` is done, it will inform the user to run `conduit pipelines init`, to initialize a pipeline.
> * `conduit pipelines init` does what we already mentioned.

Yes, I was convinced today. There’s something that wasn’t mentioned above, which hopefully addresses @nickchomey’s latest comment.

The idea behind `conduit init` is that it will accept an optional flag `--path`, with the current directory as its default value. This will allow users to initialize different environments if needed.

While we don’t expect `conduit init` to be run more than once, `conduit pipelines init` is a command that could (and likely will) be executed multiple times. Therefore, these two actions should remain separate.

raulb:
I forgot to mention something above. The intention behind the `conduit init` command includes an educational component. This command will initialize a working conduit environment, demonstrating what it comprises (it's more than just a pipeline YAML file). We believe this is helpful for illustrating the different components of a conduit project.

pkg/cli/conduit_init.go
```diff
@@ -0,0 +1,79 @@
...
+
+	fmt.Println("Conduit has been initialized!")
+	fmt.Println("To quickly create an example pipeline, run `conduit pipelines init`.")
+	fmt.Println("To see how you can customize your first pipeline, run `conduit pipelines init --help`.")
+	return nil
+}
+
+func (i *ConduitInit) createConfigYAML() error {
+	return nil
+}
```
hariso, line 58: I'm still working on this part, should be ready soon.

pkg/cli/cli.go
```diff
@@ -0,0 +1,148 @@
...
+	)
+	pipelinesInitCmd.Flags().StringVar(
+		&pipelinesInitArgs.Destination,
+		"destination",
+		"",
+		"Destination connector (any of the built-in connectors).",
+	)
+	pipelinesInitCmd.Flags().StringVar(
+		&pipelinesInitArgs.Path,
+		"path",
```
hariso, line 142: I'll change this to `pipelines.path` as the CLI design doc: https://github.com/ConduitIO/conduit/pull/1931/files#diff-942a9708b7ff04439e2aeef700f12133f188e578bda106a4e62dd17ebbdc8626R150.

hariso:
> Not that my opinion should matter much here, but I'm curious why you dont just have the one command do both things?

Opinions of users always matter.:)

Can you elaborate a bit more on "one command doing both things"?

cmd/cli/pipelines_init.go
```diff
@@ -0,0 +1,335 @@
...
+	case value == "":
+		return `""`
+	case strings.Contains(value, "\n"):
+		// specifically used in the javascript processor
+		formattedValue := formatMultiline(value, "            ", 10000)
+		return fmt.Sprintf("|\n%s", formattedValue)
+	default:
+		return fmt.Sprintf(`'%s'`, value)
+	}
+}
```
hariso, line 135: The code for filling out the template has been taken from `processorgen`, except for one small change, which is that values are printed with single quotes (so that they can contain double quotes).

pkg/conduit/entrypoint.go
```diff
@@ -27,17 +27,21 @@ import (
 	"github.com/peterbourgon/ff/v3/ffyaml"
 )
 
-// Serve is a shortcut for Entrypoint.Serve.
-func Serve(cfg Config) {
-	e := &Entrypoint{}
-	e.Serve(cfg)
+var DeprecatedFlags = map[string]bool{
```
hariso, line 30: Exported, because the CLI needs to hide some flags.

cmd/cli/pipelines_init.go
```diff
@@ -0,0 +1,335 @@
...
+		}
+	}
+
+	return connectorTemplate{}, cerrors.Errorf("%v: %w", pi.destinationConnector(), plugin.ErrPluginNotFound)
+}
+
+func (pi *PipelinesInit) configFilePath() string {
+	path := pi.args.Path
+	if path == "" {
+		path = "."
```
hariso, line 303: Thinking a bit more about this, I believe that the default value should be `./pipelines`, because that's the default directory for pipelines, and also the directory created by `conduit init`. With this, getting started with Conduit would boil down to `conduit init && conduit pipelines init && conduit`

raulb (changes requested):
@hariso nice work! 

Only one thing I'd like to add before we ship this. Looks like there's no output when running `pipelines init`. I think we should do something similar like you did on `conduit init`

```
Created directory: /Users/rb/code/conduitio/conduit/processors
Directory '/Users/rb/code/conduitio/conduit/connectors' already exists, skipping...
Directory '/Users/rb/code/conduitio/conduit/pipelines' already exists, skipping...
Configuration file written to /Users/rb/code/conduitio/conduit/conduit.yaml

Conduit has been initialized!

To quickly create an example pipeline, run 'conduit pipelines init'.
To see how you can customize your first pipeline, run 'conduit pipelines init --help'.
```

cmd/cli/pipelines_init.go
```diff
@@ -183,6 +183,14 @@ func (pi *PipelinesInit) Run() error {
 	if err != nil {
 		return cerrors.Errorf("could not write pipeline: %w", err)
 	}
+
+	fmt.Printf(`Your pipeline has been initialized and written to %s.
```
raulb, line 187:
```suggestion
	fmt.Printf(`Your pipeline has been initialized and created at %s.
```

raulb (approved):
@hariso only one suggestion. Otherwise, looks good to me!

PR #1931, opened 2024-10-25
doc: add cli design document for 0.13

### LINKED ISSUE #1926: doc: Add design document for Conduit CLI to finalize what will be part of 0.13
The goal of this issue is to clarify what part of [this design document](https://github.com/ConduitIO/conduit/discussions/1642) will be part of the upcoming [0.13 release](https://github.com/ConduitIO/conduit/milestone/15). 

I will use this document to also state what we've recently discussed and conduct meetings for posterity.

raulb:
### Description

Fixes #1926

It can be seen at

### Quick checks

### DIFF
--- a/docs/design-documents/20241024-conduit-cli.md
+++ b/docs/design-documents/20241024-conduit-cli.md
@@ -1,0 +1,86 @@
+# Conduit CLI
+
+## Introduction
+
+The goal of this design document is to clarify which part of this [Conduit discussion](https://github.com/ConduitIO/conduit/discussions/1642) will be part of the upcoming [0.13 release](https://github.com/ConduitIO/conduit/milestone/15). The intention is to provide a good starting experience and evolve it over time with more complex and advanced features.
+
+### Goals
+
+- We aim to provide our users with an interface that empowers them to use the **main** Conduit features effortlessly. While we have envisioned some complex features, they will be available in future releases.
+- We will maintain the current installation methods by incorporating the CLI through the same established processes.
+- The CLI needs to feel intuitive and align with our standards.
+
+## Background
+
+Here's how the existing Conduit CLI looks like:
+
+```shell
+$ conduit --help
+Usage of conduit:
+  -api.enabled
+    	enable HTTP and gRPC API (default true)
+  -config string
+    	global config file (default "conduit.yaml")
+  -connectors.path string
+    	path to standalone connectors' directory (default "./connectors")
+  -db.badger.path string
+    	path to badger DB (default "conduit.db")
+  -db.postgres.connection-string string
+    	postgres connection string
+  -db.postgres.table string
+    	postgres table in which to store data (will be created if it does not exist) (default "conduit_kv_store")
+  -db.type string
+    	database type; accepts badger,postgres,inmemory (default "badger")
+  -grpc.address string
+    	address for serving the gRPC API (default ":8084")
+  -http.address string
+    	address for serving the HTTP API (default ":8080")
+  -log.format string
+    	sets the format of the logging; accepts json, cli (default "cli")
+  -log.level string
+    	sets logging level; accepts debug, info, warn, error, trace (default "info")
+  -pipelines.exit-on-error
+    	exit Conduit if a pipeline experiences an error while running
+  -pipelines.path string
+    	path to the directory that has the yaml pipeline configuration files, or a single pipeline configuration file (default "./pipelines")
+  -processors.path string
+    	path to standalone processors' directory (default "./processors")
+  -version
+    	prints current Conduit version
+```
+
+## Implementation plan
+
+### `conduit init`
+
+- This command will initialize a conduit pipeline that demonstrates some basic features of Conduit. 
+- It won't require flags or arguments.
+- Additional flags could be provided to specify a source and a destination
+- Optional argument will be for
+
+#### Arguments
+
+|Argument name | Description  | Required  |
+|---|---|---|
+| Name  | Pipeline name | No |
+
+#### Flags
+
+|Flag name | Description | Required  |
+|---|---|---|
+| source  |  Connector name | No |
+| destination  |   |  No |
+
+
+#### `--help`
+
+```bash
+$ conduit init [NAME] [--source plugin@version] [--destination plugin@version]
+
+EXAMPLES
+  $ conduit init
+  $ conduit init my-first-pipeline
+```
+
+
+### Global flags
--- a/docs/design-documents/20241024-conduit-cli.md
+++ b/docs/design-documents/20241024-conduit-cli.md
@@ -53,34 +53,25 @@ Usage of conduit:
 
 ### `conduit init`
 
-- This command will initialize a conduit pipeline that demonstrates some basic features of Conduit. 
+- This command will initialize a Conduit working environment. 
 - It won't require flags or arguments.
-- Additional flags could be provided to specify a source and a destination
-- Optional argument will be for
-
-#### Arguments
-
-|Argument name | Description  | Required  |
-|---|---|---|
-| Name  | Pipeline name | No |
+- Additional flags could be provided to specify the path.
 
 #### Flags
 
-|Flag name | Description | Required  |
-|---|---|---|
-| source  |  Connector name | No |
-| destination  |   |  No |
+|Flag name | Description | Required  | Default  |
+|---|---|---|---|
+| path  |  Connector name | No | `.` (current directory) |
 
 
 #### `--help`
 
 ```bash
-$ conduit init [NAME] [--source plugin@version] [--destination plugin@version]
+$ conduit init [--path PATH]
 
 EXAMPLES
   $ conduit init
-  $ conduit init my-first-pipeline
+  $ conduit init --path ~/code/conduit-dir
 ```
-
 
 ### Global flags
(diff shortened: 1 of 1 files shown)


docs/design-documents/20241024-conduit-cli.md
```diff
@@ -0,0 +1,412 @@
...
+</details>
+
+<br/>
+
+<details>
+<summary><code style="font-size: 19px; font-weight:bold;">conduit config</code></summary>
+
+#### Description
+
+- This command will output the [Conduit configuration](https://github.com/ConduitIO/conduit/blob/05dbc275a724526f02779abb47b0ecc53f711485/pkg/conduit/config.go#L34) based on the default values and the user's configured settings that Conduit would use.
```
hariso, line 91: i.e. the effective configuration? If so, then the CLI should then be accessing a running Conduit instance, because of the command line flags, custom config files, etc.

docs/design-documents/20241024-conduit-cli.md
```diff
@@ -0,0 +1,412 @@
...
+|------|-------------|----------|---------------|
+| name  |  Pipeline file name and pipeline name  | No | `pipeline-#` (`pipeline-#.yaml`) |
+
+
+#### Flags
+
+| Name | Description | Required | Default Value |
+|------|-------------|----------|---------------|
+| destination  |  Plugin name of the destination connector  | No | `log` |
+| pipelines.path  |  Where to initialize a new pipeline | No | `.` (current directory) |
```
hariso, line 174: WDYT about path? For two reasons I'd say: (1) because of the other flags (destination and source don't have that prefix), and because of the command itself (pipelines).

raulb: @hariso my thinking is that we'll have other paths out there and would like to make it explicit. Especially considering there will be a global `config.path`. Want to make sure they're not confused.

docs/design-documents/20241024-conduit-cli.md
raulb: @hariso yup, that's the idea. Conduit should be running. Is that what you meant?

docs/design-documents/20241024-conduit-cli.md
```diff
@@ -0,0 +1,333 @@
...
+| grpc.address | address for serving the gRPC API | No | ":8084" |
+
+#### `--help`
+
+```bash
+$ conduit doctor
+# returns version, checks if there's a newer version of conduit, plugin versions, etc.
+```
+
+### conduit version
```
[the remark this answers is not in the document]
raulb: @hariso yeah, both should be compatible. `--version` would remain as global flag, but `conduit version ` would be there for visibility purposes.

docs/design-documents/20241024-conduit-cli.md
raulb: @hariso good catch. Fixed

docs/design-documents/20241024-conduit-cli.md
raulb: @hariso I think that if we want to preserve consistency, maybe adding `pipelines.path` to a pipelines command is the way to go considering the global flag for configuration will have to be `--config.path`. I also think that would likely make a better design if all flags were simply `--path`, but I'm leaning towards being explicit without limiting ourselves in the future in the event of having to combine both path flags for some reason.

raulb: Hopefully, by using configuration files, this is not a flag users need to type much anyways, so I would think it's a decent tradeoff.

docs/design-documents/20241024-conduit-cli.md
```diff
@@ -0,0 +1,375 @@
...
+#### `--help`
+
+```bash
+$ conduit pipelines ls
+NAME            STATUS
+my-pipeline     running
+my-other        degraded      
+```
+
+### `conduit pipelines describe ID`
```
hariso, line 207: I know this is about topology, but it would be good to either squeeze in the status or have a command for getting the status.

raulb: I mean, it should return everything we could have available from a running pipeline.

docs/design-documents/20241024-conduit-cli.md
```diff
@@ -0,0 +1,375 @@
...
+my-destination    conduit-connector-http@0.1.0.   standalone   my-other-pipeline
+```
+
+### `conduit connectors describe [--plugin] [--id]`
+
+This command will describe the connector configuration available.
+
+When using `--plugin` will describe the configuration for that connector plugin.
+
+When using `--id` will require having Conduit previously running since it'll describe the existing running connector.
```
hariso, line 256: So these 2 are mutually exclusive?

docs/design-documents/20241024-conduit-cli.md
lovromazgon: If we make the output of the command structured (e.g. JSON) we could for now offload this to another tool like `jq`.

Btw, that's one thing I like about the `gh` command tool for instance - some commands [accept the flag `--json`](https://cli.github.com/manual/gh_help_formatting) which will output JSON, while the default is human readable. This allows me to use it both manually as well as in a script.

docs/design-documents/20241024-conduit-cli.md
```diff
@@ -0,0 +1,375 @@
...
+
+```bash
+$ conduit processors ls
+NAME          TYPE      PIPELINE
+avro.decode   builtin   my-pipeline
+avro.encode    builtin   my-pipeline
+base64.decode  builtin   my-other-pipeline
+```
+
+### `conduit processor describe [NAME]`
```
hariso, line 305: Maybe we can make it work the same as `conduit connectors describe`, to have the `--plugin` and `--id`.

docs/design-documents/20241024-conduit-cli.md
raulb: @hariso correct

docs/design-documents/20241024-conduit-cli.md
raulb: @lovromazgon yup. `--json` is a global flag, so that should be an option.

docs/design-documents/20241024-conduit-cli.md
lovromazgon: I'm not sure I like conflating these two things in a single command 😕

lovromazgon: How about `conduit connectors plugins ...` for accessing plugins? 🤔 Or is this too verbose?

raulb: If having mutually exclusive flags is an issue, then I think we could have the separation of `connectors` and `connector-plugins`. I'm not sure having `plugins` as a subcommand of `connectors` helped with discoverability.

hariso: I too think it would be good if we find something that doesn't combine these two things, since one is about describing a connector plugin, and the other one is about a running connector.

How about we have `get-specs` and `describe`?

raulb: Maybe it's not too bad after all. That could open the possibility of listing all conector and processor plugins:

```bash
$ conduit connector-plugins ls
$ conduit processor-plugins ls
```

hariso: Would we then move (un)install under this? Looks good to me. :+1:

raulb: @lovromazgon @hariso just updated it for connectors. Let me know what you think

lovromazgon: We might as well alias it under `conduit connectors plugins` and provide both options, then we have a subcommand that is directly discoverable, while also having the one that makes sense hierarchically 🤷  WDYT?

docs/design-documents/20241024-conduit-cli.md
```diff
@@ -0,0 +1,410 @@
...
+#### `--help`
+
+```bash
+$ conduit connector-plugins ls
+PLUGIN                          
+postgres@v0.2.0                 
+conduit-connector-http@0.1.0
+```
+
+### `conduit connector-plugins PLUGIN`
```
lovromazgon, line 299: Shouldn't this then be `conduit connector-plugins describe PLUGIN`?

raulb: @lovromazgon yeah, just fixed

docs/design-documents/20241024-conduit-cli.md
raulb: @hariso 

> Would we then move (un)install under this? Looks good to me. 👍

Not planning to include `install` or `uninstall` as part of this release, but you're right. It'd be under those commands.

docs/design-documents/20241024-conduit-cli.md
```diff
@@ -0,0 +1,447 @@
...
+...
+```
+
+### `conduit doctor`
+
+This will check whether there’s a more up to date version of Conduit.
+
+Also will check iff some connectors or processors could be updated.
+
+It requires having Conduit previously running.
```
lovromazgon, line 414: Why does this require a running instance of Conduit? 🤔

Also, IMO we could do without this command at first, I think it increases the scope.

raulb: @lovromazgon yeah, you're right. This sounds like a nice to have.

lovromazgon (approved):

PR #1983, opened 2024-11-19
feat: add ecdysis to Conduit CLI

### LINKED ISSUE #1981: Integrate ecdysis with existing CLI
As part of https://github.com/ConduitIO/conduit/issues/1911, [ecdysis](https://github.com/ConduitIO/ecdysis) was recently created, which extracted our learnings from working on a CLI at [meroxa/cli](https://github.com/meroxa/cli).

These learnings were publicly shared in [this blog post](https://meroxa.com/blog/how-we-built-our-meroxa-cli/).

Before we add new commands that meet the criteria of [this design document](https://github.com/ConduitIO/conduit/blob/main/docs/design-documents/20241024-conduit-cli.md), we need to incorporate Ecdysis into Conduit so we can continue building on top.

raulb:
### Description

Fixes #1981 

### Quick checks


cmd/conduit/internal/config.go
```diff
@@ -0,0 +1,82 @@
...
+		if strings.HasPrefix(key, fmt.Sprintf("%s_", ConduitPrefix)) {
+			// Strip the prefix and replace underscores with dots
+			strippedKey := strings.ToLower(strings.TrimPrefix(key, fmt.Sprintf("%s_", ConduitPrefix)))
+			strippedKey = strings.ReplaceAll(strippedKey, "_", ".")
+
+			// Set the value in Viper
+			v.Set(strippedKey, value)
+		}
+	}
+
```
hariso, line 75: Should Viper be doing this for us (replacing dots with underscores)?

cmd/conduit/root/root.go
```diff
@@ -0,0 +1,262 @@
...
+		return err
+	}
+
+	// 2. Load environment variables and update general config.
+	if err := internal.LoadConfigFromEnv(&c.cfg); err != nil {
+		return err
+	}
+
+	// 3. Update the general config from flags.
+	c.updateConfigFromFlags()
```
hariso, line 201: IIUC, these 3 steps can be done with a single Viper instance: https://gist.github.com/hariso/5ca078ba50b4916b4ca48673165dcc90, since Viper supports reading from multiple sources simultaneously. The precedence is the same as in Conduit: https://github.com/spf13/viper?tab=readme-ov-file#why-viper

cmd/conduit/root/root.go
```diff
@@ -0,0 +1,262 @@
...
+
+	// 4. Update flags from global configuration (this will be needed for conduit init)
+	c.updateFlagValuesFromConfig()
```
hariso, line 204: If we provide the flags to the Conduit `InitCommand`, then we wouldn't need to update the flag values.

cmd/conduit/root/root.go
```diff
@@ -0,0 +1,262 @@
...
+	flags := ecdysis.BuildFlags(&c.flags)
+
+	currentPath, err := os.Getwd()
+	if err != nil {
+		panic(cerrors.Errorf("failed to get current working directory: %w", err))
+	}
+	c.cfg = conduit.DefaultConfigWithBasePath(currentPath)
+
+	conduitConfigPath := filepath.Join(currentPath, "conduit.yaml")
+	flags.SetDefault("config.path", conduitConfigPath)
```
hariso, line 222: This might be a thing that we could try adding to ecdysis, completely mapping a struct to a flag set. We're already taking some info from the tags, we could probably also take the values and the field names from the config struct too.

cmd/conduit/root/root.go
```diff
@@ -0,0 +1,262 @@
...
+	// Conduit configuration file
+	ConduitConfigPath string `long:"config.path" usage:"global conduit configuration file" persistent:"true" default:"./conduit.yaml"`
+
+	// Version
+	Version bool `long:"version" short:"v" usage:"show current Conduit version" persistent:"true"`
+
+	// Root flags -------------------------------------------------------------
+
+	// Database configuration
+	DBType                     string `long:"db.type" usage:"database type; accepts badger,postgres,inmemory,sqlite"`
```
hariso, line 50: WDYT about moving these tags to Conduit's `Config` struct? It looks like that would make the `Config` struct the single source of truth. It should then be possible to build the flags by just traversing the `Config` fields. Also, when building the config.yaml file in the `InitCommand`, we would have everything in the config struct (name, usage, and values).

Btw, even if we find that this is possible and useful, IMO it can be done in a different PR.

raulb: https://github.com/ConduitIO/ecdysis/pull/1

cmd/conduit/root/root.go
raulb: I think you meant the configuration. I agree. Now that we have tags, we should be able to: https://github.com/ConduitIO/conduit/pull/1983/commits/b6abfb6908e058f1244db3ee1d42e16e60e17afa.

cmd/conduit/root/root.go
raulb: Trying it out (still need to fully test this) https://github.com/ConduitIO/conduit/pull/1983/commits/b6abfb6908e058f1244db3ee1d42e16e60e17afa

cmd/conduit/internal/config.go
raulb: https://github.com/ConduitIO/conduit/pull/1983/commits/b6abfb6908e058f1244db3ee1d42e16e60e17afa

cmd/conduit/root/pipelines/init.go
```diff
@@ -0,0 +1,253 @@
...
+}
+
+func (c *InitCommand) buildTemplatePipeline() (pipelineTemplate, error) {
+	srcSpec, err := c.getSourceSpec()
+	if err != nil {
+		return pipelineTemplate{}, cerrors.Errorf("failed getting source params: %w", err)
+	}
+
+	// provide a simplified version
+	if c.flags.Source == defaultSource {
```
hariso, line 189: I'm thinking, what if someone would like to configure a `generator` source himself?

cmd/conduit/root/root_test.go
```diff
@@ -0,0 +1,85 @@
...
+		for _, f := range flags {
+			if f.Long == ef.longName {
+				foundFlag = &f
+				break
+			}
+		}
+
+		is.True(foundFlag != nil)
+
+		if foundFlag != nil {
```
hariso, line 78: Tiny nitpick: we don't need this check, because the above line with is.True() will abort the test if it's not true.

cmd/conduit/root/root.go
```diff
@@ -0,0 +1,120 @@
...
+	return nil
+}
+
+func (c *RootCommand) ParseConfig() ecdysis.Config {
+	path := filepath.Dir(c.flags.ConduitCfgPath)
+	c.cfg = conduit.DefaultConfigWithBasePath(path)
+
+	return ecdysis.Config{
+		EnvPrefix:  "CONDUIT",
+		ParsedCfg:  &c.cfg,
```
hariso, line 64: This can also be a new instance right?

cmd/conduit/root/root_test.go
```diff
@@ -0,0 +1,78 @@
...
+	for _, ef := range expectedFlags {
+		var foundFlag *ecdysis.Flag
+		for _, f := range flags {
+			if f.Long == ef.longName {
+				foundFlag = &f
+				break
+			}
+		}
+
+		is.True(foundFlag != nil)
```
hariso, line 76: Nitpick: here we're checking if the flag is found, and that's great. It would also be good to check if the flag itself contains what we expect (not only the long name). We could do that if `expectedFlags` is a slice of `ecdysis.Flag`.

cmd/conduit/root/root.go
```diff
@@ -0,0 +1,120 @@
...
+	flags.SetDefault("pipelines.path", c.cfg.Pipelines.Path)
+	flags.SetDefault("pipelines.exit-on-degraded", c.cfg.Pipelines.ExitOnDegraded)
+	flags.SetDefault("pipelines.error-recovery.min-delay", c.cfg.Pipelines.ErrorRecovery.MinDelay)
+	flags.SetDefault("pipelines.error-recovery.max-delay", c.cfg.Pipelines.ErrorRecovery.MaxDelay)
+	flags.SetDefault("pipelines.error-recovery.backoff-factor", c.cfg.Pipelines.ErrorRecovery.BackoffFactor)
+	flags.SetDefault("pipelines.error-recovery.max-retries", c.cfg.Pipelines.ErrorRecovery.MaxRetries)
+	flags.SetDefault("pipelines.error-recovery.max-retries-window", c.cfg.Pipelines.ErrorRecovery.MaxRetriesWindow)
+	flags.SetDefault("schema-registry.type", c.cfg.SchemaRegistry.Type)
+	flags.SetDefault("schema-registry.confluent.connection-string", c.cfg.SchemaRegistry.Confluent.ConnectionString)
+	flags.SetDefault("preview.pipeline-arch-v2", c.cfg.Preview.PipelineArchV2)
```
hariso, line 104: As a future improvement, we could add something to ecdysis that sets default values from a struct. There would be less boilerplate, and we could make sure that the flag names match with tags.

cmd/conduit/root/root.go
raulb: @hariso correct, as long as it's a pointer, it should be fine.

cmd/conduit/root/root.go
```diff
@@ -0,0 +1,120 @@
...
+func (c *RootCommand) Flags() []ecdysis.Flag {
+	flags := ecdysis.BuildFlags(&c.flags)
+
+	currentPath, err := os.Getwd()
+	if err != nil {
+		panic(cerrors.Errorf("failed to get current working directory: %w", err))
+	}
+
+	c.cfg = conduit.DefaultConfigWithBasePath(currentPath)
+	flags.SetDefault("config.path", c.cfg.ConduitCfgPath)
```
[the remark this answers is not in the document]
raulb: @hariso also, because otherwise, the flag values will be indeed present (empty) and will take precedence over anything that was previously set via viper (default, config, env) 🤦

cmd/conduit/root/root.go
hariso: In that case, we should probably just use conduit.Config{} so it's more clear that it doesn't have to be the default. Or, we can use reflection in ecdysis to create a new config (using the DefaultCfg's type). That way, a user of the ecdysis library has to do less.

raulb: @hariso I would like the parsed configuration in c.cfg though since that's what it'll be used in `Execute` (and everywhere else).

I had changed this bit to this, maybe that's clearer?

```go
func (c *RootCommand) Config() ecdysis.Config {
	path := filepath.Dir(c.flags.ConduitCfgPath)

	return ecdysis.Config{
		EnvPrefix:     "CONDUIT",
		Parsed:        &c.cfg,
		Path:          c.flags.ConduitCfgPath,
		DefaultValues: conduit.DefaultConfigWithBasePath(path),
	}
}
```

cmd/conduit/root/root.go
raulb: @hariso https://github.com/ConduitIO/ecdysis/issues/4

cmd/conduit/root/root.go
hariso: Oh, you're right, it has to be reference to that field. Nevermind.:)

hariso (approved):

cmd/conduit/root/root_test.go
raulb: @hariso https://github.com/ConduitIO/conduit/pull/1983/commits/370a638ba55f7dd4d302875bb41e4ecc1fde4fd6

PR #1987, opened 2024-11-22
Fix: filtered records holding up pipeline with destination batching

### LINKED ISSUE #1986: Batching with delay and filtering is very slow
### Bug description

We noticed in a pipeline that has records that need to be filtered and when batching is enabled, the pipeline's throughput is slower than expected.

After we eliminated some possible causes for this, we found that the combination of having a record that needs to be filtered and batching is causing this behavior. Whenever a filtered record is found in a pipeline, Conduit simply waits. Only when the batching time is over, Conduit reads the next record.

### Steps to reproduce

The below pipeline config file reads a source file. Lines that have `dontwrite` in them need to be filtered. Example source file:

```text
write 1
write 2
write 3
dontwrite 1
write 4
write 5
write 6
write 7
write 8
dontwrite 2
```

Pipeline config file:

```yaml
version: "2.2"
pipelines:
  - id: pipeline1
    status: running
    name: pipeline1
    description: A pipeline with a template record formatter
    connectors:
      - id: source1
        type: source
        plugin: builtin:file
        name: file-source
        settings:
          "sdk.schema.extract.payload.enabled": "false"
          "sdk.schema.extract.key.enabled": "false"
          path: /tmp/file-source.txt
        processors:
          - id: add-filter-metadata
            plugin: "custom.javascript"
            settings:
              script: |
                function process(rec) {
                  let afterStr = String.fromCharCode.apply(String, rec.Payload.After);
                
                  let filter = afterStr.includes("dontwrite")  
                  
                  rec.Metadata["filter"] = filter.toString();
                  return rec;
                }
          - id: filter-records
            plugin: "filter"
            condition: '{{ eq .Metadata.filter "true" }}'
      - id: destination1
        type: destination
        plugin: file
        name: file-dst
        settings:
          path: /tmp/file-destination.txt
          sdk.batch.size: 10
          sdk.batch.delay: "5s"
```
With the above setup, Conduit will write the first 3 records in 5s, then the next 5 records in 5s, and so on.

### Version

v0.12.2

hariso:
### Description

In #1986 we discovered an issue where, if there are records to be filtered and batching with a time delay is used, Conduit is waiting for the batch delay as soon as it encounters a filtered record.

The underlying issue is the following: when a processor node in a pipeline identifies a filtered record, then it goes on to acknowledge it. However, the acker node will not acknowledge it until the previous records have been acknowledged.

If batching is enabled, and the batch is not full, then Conduit is waiting for the batch to fill up, to flush it and ack the records. This creates a cyclic dependency: the filtered record is waiting for the previous records to be acked (the batch flushed), and the previous records are waiting for the filtered record to be acked, so that new records can come in and form a full batch.

This issue is solved by **not** nacking the filtered records in the processor. Instead, we let them pass through the pipeline until they reach the final node, the destination record, which simply acks them.

Fixes #1986.

### Quick checks


raulb (approved):
Code looks good.

pkg/lifecycle/stream/destination.go
```diff
@@ -101,6 +101,15 @@ func (n *DestinationNode) Run(ctx context.Context) (err error) {
 		if err != nil || msg == nil {
 			return err
 		}
+		if msg.filtered {
+			n.logger.Info(ctx).Str(log.MessageIDField, msg.ID()).
```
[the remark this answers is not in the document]
maha-hajja: made some changes, one message is under the`debug` level, so we see that something is filtered there, the others are all under the `trace` level

pkg/lifecycle/stream/destination_test.go
```diff
@@ -133,3 +135,58 @@ func TestDestinationNode_ForceStop(t *testing.T) {
...
+		close(in)
+	}()
+
+	gotMsg, gotMsgBool, err := cchan.ChanOut[*Message](out).RecvTimeout(ctx, 100*time.Millisecond)
+	is.NoErr(err)       // expected node to close outgoing channel
+	is.True(gotMsgBool) // expected node to close outgoing channel
+	is.Equal(msg, gotMsg)
+
+	err = (*csync.WaitGroup)(&nodeStopped).WaitTimeout(ctx, 100*time.Millisecond)
+	is.NoErr(err) // timed out waiting for node to be done running}
```
maha-hajja, line 191:
```suggestion
	is.NoErr(err) // timed out waiting for node to be done running
```

pkg/lifecycle/stream/message.go
```diff
@@ -65,14 +65,16 @@ type Message struct {
 	acked  chan struct{}
 	nacked chan struct{}
 
+	filtered bool
```
maha-hajja, line 68: we should add a little comment to document this field

maha-hajja (approved):
nice work!

PR #2008, opened 2024-12-11
Add processors for generating embeddings, adding context to docs, submitting prompts

hariso:
### Description

Depends on: https://github.com/conduitio-labs/conduit-connector-weaviate/tree/haris/different-vectors.

### Quick checks


pkg/plugin/processor/builtin/impl/ai/openai/embedding.go
```diff
@@ -0,0 +1,162 @@
...
+	"github.com/conduitio/conduit/pkg/foundation/cerrors"
+	"github.com/conduitio/conduit/pkg/foundation/log"
+)
+
+const EmbeddingMetadataBase64 = "openai.embedding.base64"
+
+type embeddingProcConfig struct {
+	APIKey     string `json:"apiKey" validate:"required"`
+	Endpoint   string `json:"endpoint" default:"https://api.openai.com/v1"`
+	Model      string `json:"model" validate:"required,inclusion=gpt-4|gpt-4-turbo|gpt-3.5-turbo|text-davinci-003|text-davinci-002|text-curie-001|text-babbage-001|text-ada-001"`
```
lyuboxa, line 38: so these are not embeddings models?

hariso: Good catch, I was too quick to copy-paste all the models.:)

pkg/plugin/processor/builtin/impl/ai/openai/embedding.go
```diff
@@ -0,0 +1,162 @@
...
+		return out
+	}
+
+	p.logger.Trace(ctx).
+		Any("embedding_input", embeddingInputs).
+		Any("embedding_output", embeddings).
+		Msg("got embeddings")
+
+	for i, record := range records {
+		record.Metadata[EmbeddingMetadataBase64] = embeddings.Data[i].EmbeddingBase64
```
lyuboxa, line 141: These can be horribly large, my suggestion here is to:

1. add the embedding to `.Payload.After`
2. include the model name used for the embedding.

Particularly 2, since there are others openai compat systems which may allow for different models to be used.

hariso: You're right, especially about no. 2. As for 1, that would also mean that raw data records get transformed into structured data records. That might be unexpected for some destinations.

raulb:
Looks like this was an experiment. We'll implement this in a different way.

PR #2009, opened 2024-12-11
Prevent pipeline service rerunning already failed pipeline.

### LINKED ISSUE #1852: Failing to build a pipeline from a config. file results shows "processor already running" in logs
### Bug description

```
failed to start pipeline: could not build nodes for pipeline test-pipeline: could not build source nodes: could not build processor nodes for connector test-pipeline:test-connector: processor already running
```

### Steps to reproduce

1. Specify an older version of a built-in connector in a pipeline config file, e.g. `builtin:kafka@v0.7.0`
2. Run pipeline

### Version

v0.11.1

AdamHaffar:
### Description

 Prevents pipeline service from rerunning a pipeline that the provisioning service has already tried and failed to run by setting its status to `pipeline.StatusDegraded`.

### Quick checks


pkg/lifecycle/service.go
```diff
@@ -431,6 +431,8 @@ func (s *Service) buildRunnablePipeline(
 ) (*runnablePipeline, error) {
 	nodes, err := s.buildNodes(ctx, pl)
 	if err != nil {
+		delete(s.pipelines.List(ctx), pl.ID)
```
hariso, line 434: Is there a special reason for not using s.pipelines.Delete()?

AdamHaffar: newer version of conduit handles incorrectly provisioned pipelines and makes this line redundant so I have removed it 👍

pkg/lifecycle/service.go
```diff
@@ -431,6 +431,7 @@ func (s *Service) buildRunnablePipeline(
 ) (*runnablePipeline, error) {
 	nodes, err := s.buildNodes(ctx, pl)
 	if err != nil {
+		pl.SetStatus(pipeline.StatusDegraded)
```
lovromazgon, line 434: I don't understand why we would change the status, if we didn't even start the pipeline yet? 🤔 Is there an issue connected to this PR that describes what problem we're trying to solve?

AdamHaffar: was initially linked to issue #1852 but it was fixed from pipeline recovery epic. Will close this now.

Issue #2062, opened 2025-01-10
Bug: Not all configuration is correctly parsed


raulb:
### Bug description

> [!CAUTION]
> This issue is not part of any minor release, but needs to be fixed before 0.13.0

With the recent addition of ecdysis taking care of the configuration parsing, it looks like not all options are correctly parsed. Notice in the following example how `log.level` is configured correctly, but not `grpc.address`.

@lovromazgon also noticed that `config.path` is another that's not parsed correctly.

The bug fix might need to occur in https://github.com/ConduitIO/ecdysis, but I am tracking it here for better visibility at the moment.

It's important we fix this issue because the configuration shown via `conduit config` command is the one that will be used on `conduit run`

```
$ go run ./cmd/conduit/main.go config --grpc.address :8085 --log.level warn
config.path: /Users/rb/code/conduitio/conduit/conduit.yaml
db.type: badger
db.badger.path: /Users/rb/code/conduitio/conduit/conduit.db
db.postgres.table: conduit_kv_store
db.sqlite.path: /Users/rb/code/conduitio/conduit/conduit.db
db.sqlite.table: conduit_kv_store
api.enabled: true
http.address: :8080
grpc.address: :8084
log.level: warn
log.format: cli
connectors.path: /Users/rb/code/conduitio/conduit/connectors
processors.path: /Users/rb/code/conduitio/conduit/processors
pipelines.path: /Users/rb/code/conduitio/conduit/pipelines
pipelines.exit-on-degraded: false
pipelines.error-recovery.min-delay: 1s
pipelines.error-recovery.max-delay: 10m0s
pipelines.error-recovery.backoff-factor: 2
pipelines.error-recovery.max-retries: -1
pipelines.error-recovery.max-retries-window: 5m0s
schema-registry.type: builtin
preview.pipeline-arch-v2: false
```

### Steps to reproduce

1. Check out latest code from the main branch.
2. Run `conduit config` with some configuration options. 
3. Notice not all are respected.

### Version

v0.13.0-nightly.20250110

PR #2063, opened 2025-01-10
feat: add pipelines ls cmd (+client API)

### LINKED ISSUE #2019: CLI: conduit pipelines ls
This command will list the running pipelines.

#### Flags

| Name | Description | Required | Default Value |
|------|-------------|----------|---------------|
| grpc.address | address for serving the gRPC API | No | ":8084" |

#### `--help`

```bash
$ conduit pipelines ls
NAME            STATUS
my-pipeline     running
my-other        degraded      
```

-----

raulb:
### Description

- Fixes #2019
- Adds a gRPC client that will be used in the rest of the commands that require Conduit to be running.
- Adds a custom decorator `ExecuteWithClient` that parses the global flags, needed to correctly the client address.

### Quick checks


cmd/conduit/api/client.go
```diff
@@ -0,0 +1,55 @@
...
+	"google.golang.org/grpc"
+	"google.golang.org/grpc/credentials/insecure"
+	healthgrpc "google.golang.org/grpc/health/grpc_health_v1"
+)
+
+type Client struct {
+	conn            *grpc.ClientConn
+	pipelineService apiv1.PipelineServiceClient
+	HealthService   healthgrpc.HealthClient
+}
```
lovromazgon, line 31: Thinking out loud - do we need a client abstraction here? Seems like it will contain just lots of boilerplate code. Can we get away with just giving the grpc client to the commands? 🤔 Or maybe, for convenience, it could just combine all service clients and make them available as public struct fields (same as `HealthService`), then we don't have to write methods that forward calls to the specific service client.

cmd/conduit/root/pipelines/list.go
```diff
@@ -0,0 +1,118 @@
...
+
+func (c *ListCommand) Execute(ctx context.Context) error {
+	// TODO: Move this elsewhere since it'll be common for all commands that require having Conduit Running
+	// --------- START
+	conduitNotRunning := "Notice: To inspect the API, Conduit needs to be running" +
+		"\nPlease execute `conduit run`"
+
+	conduitGRPCAddr := c.RunCmd.GRPCAddress()
+
+	conduitClient, err := api.NewClient(ctx, conduitGRPCAddr)
```
lovromazgon, line 64: Not sure if possible, but we could think about injecting the client with a decorator through ecdysis. Similarly how we will inject a logger. Didn't we do something similar in the meroxa CLI?

cmd/conduit/root/root.go
```diff
@@ -76,7 +76,7 @@ func (c *RootCommand) SubCommands() []ecdysis.Command {
 		&config.ConfigCommand{RunCmd: runCmd},
 		&initialize.InitCommand{Cfg: &runCmd.Cfg},
 		&version.VersionCommand{},
-		&pipelines.PipelinesCommand{},
-		&run.RunCommand{},
+		&pipelines.PipelinesCommand{RunCmd: runCmd},
```
lovromazgon, line 79: I guess you're trying to inherit the flags from the run command? That won't work, since the pipelines command is not a subcommand of run. It would only work if `RootCommand` had the flags attached and they were marked as persistent (see the [example](https://github.com/ConduitIO/ecdysis/blob/4a9a7e14d5e156daec9e193cd6116343e8eb7d0e/example/main.go#L71-L72)).

cmd/conduit/root/run/run.go
```diff
@@ -42,6 +43,12 @@ type RunCommand struct {
 
 func (c *RunCommand) Execute(_ context.Context) error {
 	e := &conduit.Entrypoint{}
+
+	if !c.Cfg.API.Enabled {
+		fmt.Print("Warning: API is currently disabled. Most Conduit CLI commands won't work without the API enabled." +
+			"To enable it, run conduit with `--api.enabled=true` or set `CONDUIT_API_ENABLED=true` in your environment.")
+	}
```
hariso, line 50: Nice touch, I like it!

cmd/conduit/api/client.go
hariso: IMHO it plays a role how we manage to inject the client into commands. Having a single client that holds all services could be handy, because it might be easier to instantiate once and inject into all commands that need it.

cmd/conduit/root/pipelines/list.go
```diff
@@ -0,0 +1,118 @@
...
+	apiv1 "github.com/conduitio/conduit/proto/api/v1"
+	"github.com/conduitio/ecdysis"
+	healthgrpc "google.golang.org/grpc/health/grpc_health_v1"
+)
+
+var (
+	_ ecdysis.CommandWithExecute = (*ListCommand)(nil)
+	_ ecdysis.CommandWithAliases = (*ListCommand)(nil)
+	_ ecdysis.CommandWithDocs    = (*ListCommand)(nil)
+	// with API client ?
```
hariso, line 34: IMO that would be really nice and would make writing new commands easier.

cmd/conduit/root/pipelines/list.go
raulb: @lovromazgon yup. Been thinking about it. I think I'll do something similar https://github.com/meroxa/cli/blob/5cfa2489a37590377ff021a6c867724b31e4d149/cmd/meroxa/builder/builder.go#L67-L71. Had [a reminder](https://github.com/ConduitIO/conduit/pull/2063#discussion_r1911067810) for now, but I'll make it part of this.

cmd/conduit/root/root.go
raulb: @lovromazgon I was actually trying to read from the configured port. Haven't been able to test this properly because of this bug, so I might try to fix it to ensure this is working as expected https://github.com/ConduitIO/conduit/issues/2062.

cmd/conduit/api/client.go
raulb: @lovromazgon the idea I had was to abstract this for the CLI only in case we wanted to change specific versions on different endpoints (services) to keep these two independent. Maybe we could leave these for now here and refactor when we do really need to.

lovromazgon: I do see the value in having a single client that holds all services, but I fail to see the need to forward the calls through methods like `ListPipelines` in line 49. I'd just make `pipelineService` public and let the command call it directly - less abstraction, less boilerplate code.

cmd/conduit/root/root.go
lovromazgon: It is a bit awkward currently to share the same set of flags in unrelated commands (commands that don't share the same parent), or when commands do share the same parent, but the parent doesn't use the same flags. Maybe this can be addressed in ecdysis so it's easier to achieve, but I'm not sure how 🤔

pkg/conduit/config.go
```diff
@@ -39,47 +39,46 @@ const (
...
-	SQLite struct {
-		Path  string `long:"db.sqlite.path" usage:"path to sqlite3 DB"`
-		Table string `long:"db.sqlite.table" usage:"sqlite3 table in which to store data (will be created if it does not exist)"`
-	}
-}
+// Config holds all configurable values for Conduit.
+type Config struct {
+	ConduitCfgPath string `long:"config.path" usage:"global conduit configuration file" default:"./conduit.yaml"`
 
-type ConfigAPI struct {
```
hariso, line 61: Is there a special reason for reverting these changes? I don't mind either, I'm just curious if this helps something.

raulb: @hariso it doesn't. I just happened to scroll up and down a lot and realized that it didn't really make a difference. This way is easier to see the structure in my opinion.

cmd/conduit/cecdysis/decorators.go
```diff
@@ -0,0 +1,106 @@
...
+	ExecuteWithClient(context.Context, *api.Client) error
+}
+
+// CommandWithExecuteWithClientDecorator is a decorator that adds a Conduit API client to the command execution.
+type CommandWithExecuteWithClientDecorator struct{}
+
+func (CommandWithExecuteWithClientDecorator) Decorate(_ *ecdysis.Ecdysis, cmd *cobra.Command, c ecdysis.Command) error {
+	v, ok := c.(CommandWithExecuteWithClient)
+	if !ok {
+		return nil
```
maha-hajja, line 46: just a question, when would this case happen? and why not return an error?

pkg/conduit/config.go
```diff
@@ -39,47 +39,46 @@ const (
...
+		// When Driver is specified it takes precedence over other DB related
+		// fields.
+		Driver database.DB
+
+		Type   string `long:"db.type" usage:"database type; accepts badger,postgres,inmemory,sqlite"`
+		Badger struct {
+			Path string `long:"db.badger.path" usage:"path to badger DB"`
+		}
+		Postgres struct {
+			ConnectionString string `long:"db.postgres.connection-string" usage:"postgres connection string, may be a database URL or in PostgreSQL keyword/value format"`
```
maha-hajja, line 56: about `connection-string`, does camel case not work for a command? or was this format agreed on by the team?

maha-hajja (approved):
Nice work! the PR code is well written and structured :D

cmd/conduit/cecdysis/decorators.go
raulb: @maha-hajja since this is a decorator, when not implemented by a command, we wouldn't want to run the code and it wouldn't imply an error. This is an expected behavior.

raulb: These are the ecdysis decorators also for reference https://github.com/ConduitIO/ecdysis/blob/main/decorators.go

pkg/conduit/config.go
raulb: @maha-hajja `connection-string` was already a conduit flag and part of our configuration files https://conduit.io/docs/configuration#cli-flags. It's how all our flags with two words are normally named.

cmd/conduit/api/client.go
```diff
@@ -0,0 +1,66 @@
...
+	}
+
+	client := &Client{
+		conn:                  conn,
+		PipelineServiceClient: apiv1.NewPipelineServiceClient(conn),
+		HealthClient:          healthgrpc.NewHealthClient(conn),
+	}
+
+	if err := client.CheckHealth(ctx); err != nil {
+		client.Close()
```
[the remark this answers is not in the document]
raulb: @hariso the error is ultimately logged after it's returned [here](https://github.com/ConduitIO/conduit/blob/dce673d68531fbe76af6f8bff7fee587be3f4e50/cmd/conduit/api/client.go#L59)

cmd/conduit/cecdysis/decorators.go
```diff
@@ -0,0 +1,106 @@
...
+
+	cfg := ecdysis.Config{
+		EnvPrefix:     "CONDUIT",
+		Parsed:        &usrCfg,
+		Path:          path,
+		DefaultValues: defaultConfigValues,
+	}
+
+	// If it can't be parsed, we return the default value
+	err = ecdysis.ParseConfig(cfg, cmd)
```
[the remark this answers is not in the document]
raulb: Good call. https://github.com/ConduitIO/conduit/pull/2063/commits/c2e2eb277f38848ee331c27024ae9d1ea8713641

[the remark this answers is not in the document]
raulb:
```
We couldn't connect to Conduit at the configured address ":8084"
Please execute `conduit run` to start it.
To check the current configured `api.grpc.address`, run `conduit config`

Error details: rpc error: code = Unavailable desc = connection error: desc = "transport: Error while dialing: dial tcp [::1]:8084: connect: connection refused"
```

cmd/conduit/api/client.go
```diff
@@ -0,0 +1,66 @@
...
+		return nil, err
+	}
+
+	return client, nil
+}
+
+func (c *Client) CheckHealth(ctx context.Context) error {
+	healthResp, err := c.HealthClient.Check(ctx, &healthgrpc.HealthCheckRequest{})
+	if err != nil || healthResp.Status != healthgrpc.HealthCheckResponse_SERVING {
+		return fmt.Errorf("notice: to inspect the API, Conduit needs to be running\nPlease execute `conduit run`")
```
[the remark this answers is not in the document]
raulb: related https://github.com/ConduitIO/conduit/commit/c2e2eb277f38848ee331c27024ae9d1ea8713641.

cmd/conduit/root/pipelines/list.go
```diff
@@ -0,0 +1,87 @@
...
+		return fmt.Errorf("failed to list pipelines: %w", err)
+	}
+
+	displayPipelines(resp.Pipelines)
+
+	return nil
+}
+
+func displayPipelines(pipelines []*apiv1.Pipeline) {
+	if len(pipelines) != 0 {
```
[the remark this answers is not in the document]
raulb: https://github.com/ConduitIO/conduit/pull/2063/commits/e0f30dcd7776a98f3abb1c54206536a804e6e33e

pkg/conduit/config.go
maha-hajja: never noticed it for some reason! thanks for the reference

PR #2070, opened 2025-01-15
Fix CLI flag naming

hariso:
### Description

Fixes https://github.com/ConduitIO/conduit/issues/2062.

There were two issues:

1. With boolean flags, e.g. `api.enabled`, the "issue" was in the usage. Sometimes we use `--api.enabled true`, but pflag doesn't like that ([link](https://github.com/spf13/pflag?tab=readme-ov-file#command-line-flag-syntax), pflag is used by Viper that our CLI uses)
2. With other flags, the issue was with naming. When unmarshalling the input map (that contains flag names and values) into Conduit's configuration structs, certain fields were skipped because of name mismatches. mapstructure, used by viper, assumes default key names for Go struct fields, but we were using non-default in certain cases.

As part of above changes, `http.address` and `grpc.address` were renamed to `api.http.address` and `api.grpc.address` respectively. 

### Quick checks


pkg/conduit/config.go
```diff
@@ -76,10 +76,13 @@ type ConfigLog struct {
 
 // Config holds all configurable values for Conduit.
 type Config struct {
-	ConduitCfgPath string `long:"config.path" usage:"global conduit configuration file" default:"./conduit.yaml"`
-	DB             ConfigDB
-	API            ConfigAPI
-	Log            ConfigLog
+	ConduitCfg struct {
+		Path string `long:"config.path" usage:"global conduit configuration file" default:"./conduit.yaml"`
+	} `mapstructure:"config"`
```
hariso, line 81: Dots are not allowed in mapstructure/json tag values.

raulb (approved):
Thank you @hariso !!

PR #2081, opened 2025-01-17
feat(cli): add processor plugins ls

### LINKED ISSUE #2027: CLI: conduit processor-plugins ls
### Feature description

This command will list all the available processors.

It does require having Conduit previously running.

#### `--help`

```bash
$ conduit processor-plugins ls
NAME
avro.decode
avro.encode
base64.decode
```

-----

raulb:
### Description

Fixes #2027 

### Quick checks

### DIFF
--- a/cmd/conduit/root/processorplugins/list.go
+++ b/cmd/conduit/root/processorplugins/list.go
@@ -81,14 +81,14 @@ func displayProcessorPlugins(processorPlugins []*apiv1.ProcessorPluginSpecificat
 	table.Header = &simpletable.Header{
 		Cells: []*simpletable.Cell{
 			{Align: simpletable.AlignCenter, Text: "NAME"},
-			{Align: simpletable.AlignCenter, Text: "DESCRIPTION"},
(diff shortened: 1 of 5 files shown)


cmd/conduit/root/processorplugins/list.go
```diff
@@ -0,0 +1,98 @@
...
+
+	displayProcessorPlugins(resp.Plugins)
+
+	return nil
+}
+
+func displayProcessorPlugins(processorPlugins []*apiv1.ProcessorPluginSpecifications) {
+	if len(processorPlugins) == 0 {
+		return
+	}
```
hariso, line 77: Since this is the whole response that a user will (or won't) see, maybe it's good to write "no processor plugins found".

hariso (approved):

cmd/conduit/root/processorplugins/list.go
hariso: That makes sense to me. :+1:

Issue #1911, opened 2024-10-15
Epic: +CLI -UI


raulb:
As discussed in [this design document,](https://github.com/ConduitIO/conduit/discussions/1642) a Conduit CLI is what we believe should provide the developer experience our Conduit users need.

This issue will tackle all the necessary work to make a Conduit CLI a reality as part of [Conduit 0.13](https://github.com/ConduitIO/conduit/milestone/15).

> [!IMPORTANT]  
> While adding a new way to interact with Conduit, we'll **deprecate** the UI. There isn't feature parity at the moment and we strongly believe the CLI will be enough for now 

### -UI

- [x] https://github.com/ConduitIO/conduit/pull/1916
- [x] https://github.com/ConduitIO/conduit/pull/1917
- [x] Delete `UI` and `javascript` label [from conduit](https://github.com/ConduitIO/conduit/issues/labels)
- [x] conduit: remove UI code (https://github.com/ConduitIO/conduit/pull/2036)
- [x] conduit: remove [UI GitHub workflows](https://github.com/ConduitIO/conduit/tree/main/.github/workflows) (https://github.com/ConduitIO/conduit/pull/2036)
- [x] Archive, updating description of https://github.com/ConduitIO/mx-ui-components https://github.com/ConduitIO/mx-ui-components/pull/209

#### Does not require having Conduit running

- [x] https://github.com/ConduitIO/conduit/issues/1926
- [x] https://github.com/ConduitIO/ecdysis
- [x] https://github.com/ConduitIO/conduit/issues/1981
- [x] https://github.com/ConduitIO/conduit/issues/2017
- [x] https://github.com/ConduitIO/conduit/issues/2029
- [x] https://github.com/ConduitIO/conduit/issues/2018
- [x] https://github.com/ConduitIO/conduit/issues/2034
- [x] #2095

#### Requires having Conduit running

- [x] https://github.com/ConduitIO/conduit/issues/2019
- [x] https://github.com/ConduitIO/conduit/issues/2021
- [x] https://github.com/ConduitIO/conduit/issues/2025
- [x] https://github.com/ConduitIO/conduit/issues/2027
- [x] https://github.com/ConduitIO/conduit/issues/2023
- [x] https://github.com/ConduitIO/conduit/issues/2020
- [x] https://github.com/ConduitIO/conduit/issues/2022
- [x] https://github.com/ConduitIO/conduit/issues/2026
- [x] https://github.com/ConduitIO/conduit/issues/2024
[...]

PR #2082, opened 2025-01-17
feat(CLI): add connectors describe

### LINKED ISSUE #2022: CLI: conduit connectors describe ID
This command will describe the existing running connector.

#### Arguments

| Name | Description              | Required | Default Value |
|------|--------------------------|----------|---------------|
| id   | connector id to describe | Yes      |               |

#### Flags

| Name         | Description                      | Required | Default Value |
|--------------|----------------------------------|----------|---------------|
| grpc.address | address for serving the gRPC API | No       | ":8084"       |

#### `--help`

```bash
$ conduit connnectors describe ID [--grpc.address]

EXAMPLE:

$ conduit connectors describe my-source
NAME        URL             PIPELINE        PLUGIN
my-source   https://...     my-pipeline     postgres@v0.2.0
```

-----

raulb:
### Description

- Fixes #2022 
- Add some fixes to pipelines describe

### Quick checks


cmd/conduit/internal/print_utils.go
```diff
@@ -12,29 +12,28 @@
...
-func getPipelineStatus(pipeline *apiv1.Pipeline) string {
-	return prettyProtoEnum("STATUS_", pipeline.State.Status.String())
+func PrintStatusFromProtoString(protoStatus string) string {
+	return PrettyProtoEnum("STATUS_", protoStatus)
 }
 
-func prettyProtoEnum(prefix, protoEnum string) string {
+func PrettyProtoEnum(prefix, protoEnum string) string {
 	return strings.ToLower(
 		strings.ReplaceAll(protoEnum, prefix, ""),
```
lovromazgon, line 33:
```suggestion
		strings.TrimPrefix(protoEnum, prefix),
```

cmd/conduit/root/connectors/describe.go
```diff
@@ -0,0 +1,159 @@
...
+	args DescribeArgs
+}
+
+func (c *DescribeCommand) Usage() string { return "describe" }
+
+func (c *DescribeCommand) Docs() ecdysis.Docs {
+	return ecdysis.Docs{
+		Short: "Describe an existing connector",
+		Long: `This command requires Conduit to be already running since it will describe a connector registered
+by Conduit. You can list existing connectors with the 'conduit connectors ls' command.`,
```
lovromazgon, line 50: I'd use `list` in the doc string, since that's the full command name, `ls` is only an alias.
```suggestion
by Conduit. You can list existing connectors with the 'conduit connectors list' command.`,
```

cmd/conduit/root/pipelines/describe.go
```diff
@@ -36,17 +37,18 @@ var (
...
 
 func (c *DescribeCommand) Docs() ecdysis.Docs {
 	return ecdysis.Docs{
 		Short: "Describe an existing pipeline",
-		Long: `This command requires Conduit to be already running since it will list all pipelines registered 
-by Conduit. This will depend on the configured pipelines directory, which by default is /pipelines; however, it could 
-be configured via --pipelines.path at the time of running Conduit.`,
-		Example: "conduit pipelines describe\nconduit pipelines desc",
+		Long: `This command requires Conduit to be already running since it will describe a pipeline registered 
+by Conduit. You can list existing pipelines with the 'conduit pipelines ls' command.`,
```
lovromazgon, line 49:
```suggestion
by Conduit. You can list existing pipelines with the 'conduit pipelines list' command.`,
```

cmd/conduit/root/pipelines/describe.go
```diff
@@ -169,8 +171,8 @@ func printProcessors(b *strings.Builder, ids []string, indent int) {
 		return
 	}
 
-	fmt.Fprintf(b, "%sProcessors:\n", indentation(indent))
+	fmt.Fprintf(b, "%sProcessors:\n", internal.Indentation(indent))
 	for _, id := range ids {
-		fmt.Fprintf(b, "%s- %s\n", indentation(indent+1), id)
+		fmt.Fprintf(b, "%s- %s\n", internal.Indentation(indent+1), id)
 	}
```
lovromazgon, line 177: Here we only print IDs, while in the connectors describe we print the whole processor config and condition. Would be good to make this the same IMO.

raulb: @lovromazgon good point

lovromazgon (approved):
I'll approve, as the command in question here is looking good 👍

However, let me point out that in `pipeline describe` we now have the discrepancy between how a connector is shown vs. how a processor is shown. Arguably, connectors are even more important, but we only show their ID and plugin. How about we do the same thing for connectors and show the whole description of the connector? Essentially showing the whole pipeline config. WDYT?

The discrepancy I'm talking about:

Screenshot 2025-01-21 at 16 17 18

raulb:
> I'll approve, as the command in question here is looking good 👍
> 
> However, let me point out that in `pipeline describe` we now have the discrepancy between how a connector is shown vs. how a processor is shown. Arguably, connectors are even more important, but we only show their ID and plugin. How about we do the same thing for connectors and show the whole description of the connector? Essentially showing the whole pipeline config. WDYT?
> 
> The discrepancy I'm talking about:
> 
> Screenshot 2025-01-21 at 16 17 18

@lovromazgon You're right. I'll make that change

raulb:
@lovromazgon will make sure I address that as part of https://github.com/ConduitIO/conduit/issues/1911 separately.

PR #2086, opened 2025-01-21
describe connector plugins command

### LINKED ISSUE #2024: CLI: conduit connector-plugins describe PLUGIN
This command will describe the configuration for that connector plugin.

#### Arguments

| Name   | Description                  | Required | Default Value |
|--------|------------------------------|----------|---------------|
| PLUGIN | connector plugin to describe | Yes      |               |

#### `--help`

```bash
$ conduit connnector-plugins describe PLUGIN

EXAMPLE:

$ conduit connector-plugins describe conduit-connector-http@0.1.0
Name: ...
Description:
...
```

-----

maha-hajja:
### Description

describes a specific connector plugin.

example: `./conduit connector-plugin desc builtin:s3@v0.8.1`

image

Fixes #2024 

### Quick checks

### DIFF
--- a/cmd/conduit/internal/print_utils.go
+++ b/cmd/conduit/internal/print_utils.go
@@ -70,3 +70,32 @@ func DisplayProcessors(processors []*apiv1.Processor, indent int) {
 		fmt.Printf("%sUpdated At: %s\n", Indentation(indent+2), PrintTime(p.UpdatedAt))
 	}
 }
+
+// FormatLongString splits a string into multiple lines depending on the maxLineLength.
+func FormatLongString(paragraph string, maxLineLength int) string {
+	if len(paragraph) <= maxLineLength {
+		return paragraph
+	}
+
+	var result strings.Builder
+	var currentLine strings.Builder
+	words := strings.Fields(paragraph)
+	for _, word := range words {
+		// check if adding the next word would exceed the line length
+		if currentLine.Len()+len(word)+1 > maxLineLength {
+			result.WriteString(currentLine.String() + "\n")
+			currentLine.Reset()
+		}
+		if currentLine.Len() > 0 {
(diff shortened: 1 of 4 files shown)


cmd/conduit/internal/print_utils.go
```diff
@@ -70,3 +72,121 @@ func DisplayProcessors(processors []*apiv1.Processor, indent int) {
...
+			{Align: simpletable.AlignLeft, Text: item.name},
+			{Align: simpletable.AlignLeft, Text: formatType(item.param.GetType().String())},
+			{Align: simpletable.AlignLeft, Text: FormatLongString(item.param.Description, 100)},
+			{Align: simpletable.AlignLeft, Text: item.param.Default},
+			{Align: simpletable.AlignLeft, Text: formatValidations(item.param.Validations)},
+		}
+		table.Body.Cells = append(table.Body.Cells, r)
+	}
+
+	table.SetStyle(simpletable.StyleDefault)
```
raulb, line 160: I noticed this one sets a different style of a table. I don't dislike this one, but we should be consistent. Suggesting the ones we've used for now until we agreed on changing the rest.

```suggestion
	table.SetStyle(simpletable.StyleCompact)
```

raulb (approved):
I'm approving ahead of time, but I think before we ship the two PRs (this one and https://github.com/ConduitIO/conduit/pull/2088), we should:

1. Consolidate the look of all tables so they're consistent (fwiw, I'm ok with the default style since I agree that on describe, it reads better).
2. Hide those fields that are empty.

(Bonus is filing an issue for why author is always empty).

cmd/conduit/internal/print_utils.go
maha-hajja: change the rest to `StyleDefault `

Issue #2093, opened 2025-01-23
CLI: update processor type on `processors ls`


raulb:

raulb:
Originally, the issue tracked adding processor type on the ls command, but this in fact was already in place. The PR attached only update its output and adds a test.

raulb:
Fixed by https://github.com/ConduitIO/conduit/pull/2097

PR #2097, opened 2025-01-23
refactor: update type on processors ls

raulb:
### Description

- Updates #2093 

### Quick checks

### DIFF
--- a/cmd/conduit/internal/print_utils.go
+++ b/cmd/conduit/internal/print_utils.go
@@ -223,4 +223,3 @@ func ConnectorTypeToString(connectorType apiv1.Connector_Type) string {
 		return "unknown"
 	}
 }
-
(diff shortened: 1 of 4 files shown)


cmd/conduit/internal/print_utils.go
```diff
@@ -195,3 +195,17 @@ func formatValidations(v []*configv1.Validation) string {
 	}
 	return result.String()
 }
+
+// ProcessorTypeToString returns a human-readable string from a processor parent type
+func ProcessorTypeToString(processorParentType apiv1.Processor_Parent_Type) string {
```
lovromazgon, line 200: The word "type" is confusing to me here, how about we call this "parent" instead, like it is internally named? So that we would display `Parent: connector (conn-id-foo)`.

raulb: @lovromazgon that works

raulb: https://github.com/ConduitIO/conduit/pull/2097/commits/1d381aee2095cdca234e1275ef4d3acf4e6cd62d

lovromazgon (approved):

PR #2104, opened 2025-01-24
feat: use output decorator, adding tests

### LINKED ISSUE #2034: CLI: Implement easy capture of output (for tests)
Similarly to how we implemented it in meroxa's CLI: https://github.com/meroxa/cli/tree/4836e3d078f42ae19e48234d24300bac50e62734/log

Requirements are an easy way to overwrite the `stdout` on tests so we could compare the execution output with the expected one.

> [!NOTE]  
> It needs https://github.com/ConduitIO/ecdysis/pull/11

-----

raulb:
### Description

Fixes #2034 
### Quick checks

### DIFF
--- a/cmd/conduit/root/version/version_test.go
+++ b/cmd/conduit/root/version/version_test.go
@@ -1,0 +1,48 @@
+// Copyright © 2025 Meroxa, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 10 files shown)


cmd/conduit/root/version/version_test.go
```diff
@@ -0,0 +1,48 @@
...
+
+	buf := new(bytes.Buffer)
+
+	out := &ecdysis.DefaultOutput{}
+	out.Output(buf, nil)
+
+	cmd := &VersionCommand{}
+	cmd.Output(out)
+
+	expectedOutput := fmt.Sprintf("%s/%s", runtime.GOOS, runtime.GOARCH)
```
maha-hajja, line 40: is there a reason why we're not using `conduit.Version` here too?

maha-hajja (approved):

cmd/conduit/root/version/version_test.go
raulb: @maha-hajja you're right. Just updated it https://github.com/ConduitIO/conduit/pull/2104/commits/316c03470e072942fb0616745ad52be0476d620a, thank you

PR #2111, opened 2025-01-28
chore: add tests to ExecuteWithClient for pipelines cmds

### LINKED ISSUE #2095: CLI: Improve test coverage
Add tests for the `Execute` methods.

raulb:
### Description

Fixes #2095

- [x] pipelines ls
- [x] pipelines describe
- [x] connectors ls
- [x] connectors describe
- [x] processors ls
- [x] processors describe
- [x] processorplugins ls
- [x] processorplugins describe
- [x] connectorplugins ls
- [x] connectorplugins describe

### Quick checks


cmd/conduit/root/pipelines/list_test.go
```diff
@@ -0,0 +1,107 @@
...
+
+	cmd := &ListCommand{}
+	cmd.Output(out)
+
+	err := cmd.ExecuteWithClient(ctx, client)
+	is.NoErr(err)
+
+	output := buf.String()
+	is.True(len(output) > 0)
+	is.True(strings.Contains(output, "ID"))
```
maha-hajja, line 66: what do you think about creating the output table as we do under Execute, and comparing the whole string?

raulb: @maha-hajja I can try something like that. In the past, these string comparisons ended up being tricky due to different on spaces, `\t`, `\n`, etc... but I can give it a try.

maha-hajja (approved):
lgtm

cmd/conduit/internal/testutils/mock_helpers.go
```diff
@@ -0,0 +1,111 @@
...
+			Config: &apiv1.Processor_Config{
+				Settings: settings,
+			},
+		},
+	}, nil).Times(1)
+}
+
+// ConnectorService --------------------------------------------
+
+func MockGetListConnectors(mockService *mock.MockConnectorService, pipelineID string, connectors []*apiv1.Connector) {
```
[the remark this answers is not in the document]
raulb: @maha-hajja thanks for the review! I actually would like to leave the `Get` in there for future mocks where the method is different. It's interesting that the grpc method is `ListConnectors`, but for other services is still uses `Get`  when they are not lists. 🤷 . I just updated `MockListPipelines` to `MockGetListPipelines` so they all are consistent.

cmd/conduit/internal/testutils/mock_helpers.go
```diff
@@ -0,0 +1,144 @@
...
+import (
+	"time"
+
+	"github.com/conduitio/conduit/cmd/conduit/api/mock"
+	apiv1 "github.com/conduitio/conduit/proto/api/v1"
+	"go.uber.org/mock/gomock"
+	"google.golang.org/protobuf/types/known/timestamppb"
+)
+
+func GetDateTime() *timestamppb.Timestamp {
```
hariso, line 26: Maybe indicate in the method name this is the current time?

raulb: @hariso It's not the current time! :P

cmd/conduit/root/connectorplugins/list.go
```diff
@@ -31,14 +31,20 @@ var (
...
 
 type ListCommand struct {
-	flags ListFlags
+	flags  ListFlags
+	output ecdysis.Output
+}
+
+func (c *ListCommand) Output(output ecdysis.Output) {
+	c.output = output
 }
```
hariso, line 48: Would it make sense to have an `OutputCommand` struct that's embedded, so that the ListCommand can embed it and get the ecdysus.Output field and the Output() method?

cmd/conduit/root/connectorplugins/list_test.go
```diff
@@ -54,3 +62,138 @@ func TestListCommandFlags(t *testing.T) {
...
+	cmd := &ListCommand{
+		flags: ListFlags{
+			Name: "builtin",
+		},
+	}
+	cmd.Output(out)
+
+	ctx := context.Background()
+	ctrl := gomock.NewController(t)
+	defer ctrl.Finish()
```
hariso, line 82: Out of curiosity, why is this needed? I've seen and used the gomock controller a few times, but never saw defer ctrl.Finish()?

cmd/conduit/root/connectors/list_test.go
```diff
@@ -54,3 +62,136 @@ func TestConnectorsListCommandFlags(t *testing.T) {
...
+	cmd := &ListCommand{}
+	cmd.Output(out)
+
+	err := cmd.ExecuteWithClient(ctx, client)
+	is.NoErr(err)
+
+	output := buf.String()
+	is.True(len(output) > 0)
+
+	strings.Contains(output, ""+
```
hariso, line 114: Should we actually test if output is equal to the expected string?

cmd/conduit/root/pipelines/describe_test.go
```diff
@@ -54,3 +64,127 @@ func TestDescribeExecutionCorrectArgs(t *testing.T) {
...
+		}, []string{"proc3"})
+
+	testutils.MockGetDLQ(mockPipelineService, cmd.args.PipelineID, "dlq-plugin")
+
+	client := &api.Client{
+		PipelineServiceClient:  mockPipelineService,
+		ProcessorServiceClient: mockProcessorService,
+		ConnectorServiceClient: mockConnectorService,
+	}
+
```
hariso, line 138: Maybe it makes sense to have a helper method that creates this client and the needed mock services?

cmd/conduit/root/processorplugins/processor_plugins.go
```diff
@@ -26,7 +26,9 @@ var (
 
 type ProcessorPluginsCommand struct{}
 
-func (c *ProcessorPluginsCommand) Aliases() []string { return []string{"processor-plugin"} }
+func (c *ProcessorPluginsCommand) Aliases() []string {
+	return []string{"processor-plugin", "processorsplugins", "processorplugins", "processorplugin"}
+}
```
hariso, line 32: :1st_place_medal:

hariso (approved):
Great work Raúl!

cmd/conduit/root/connectorplugins/list_test.go
raulb: I actually assumed this was needed to ensure the call was made and correctly cleaned up regardless of a panic, etc. But I just checked the source code and noticed this is done automatically so there's no need to do it. I'll remove them all.

```go
// NewController returns a new Controller. It is the preferred way to create a Controller.
//
// Passing [*testing.T] registers cleanup function to automatically call [Controller.Finish]
// when the test and all its subtests complete.
func NewController(t TestReporter, opts ...ControllerOption) *Controller {
	h, ok := t.(TestHelper)
	if !ok {
		h = &nopTestHelper{t}
	}
	ctrl := &Controller{
		T:             h,
		expectedCalls: newCallSet(),
	}
	for _, opt := range opts {
		opt.apply(ctrl)
	}
	if c, ok := isCleanuper(ctrl.T); ok {
		c.Cleanup(func() {
			ctrl.T.Helper()
			ctrl.finish(true, nil)
		})
	}

	return ctrl
}
```

cmd/conduit/root/connectors/list_test.go
raulb: @hariso tried that, but then I had to worried about possible subtle differences of a single space 🤦

cmd/conduit/root/connectorplugins/list.go
raulb: @hariso hmmm, not entirely sure. I think it'd change how the other decorators behave, and if anything, I'd tackle it in a separate PR to not change its scope.

cmd/conduit/root/connectorplugins/list_test.go
raulb: https://github.com/ConduitIO/conduit/pull/2111/commits/da52457e000baf156381bde513172c47e5e8fea0

cmd/conduit/root/connectors/list_test.go
raulb: Figured out what's the actual difference, so I think I'll change them all. Thanks for the suggestion of taking another look at this.

raulb: https://github.com/ConduitIO/conduit/pull/2111/commits/e091a228748c6284fbaea1cf34ef5d59b5607887

cmd/conduit/root/pipelines/describe_test.go
raulb: @hariso that's a good idea. I think I'll ship it like this, and maybe re-iterate on this on future PRs. Thanks for the suggestion!

PR #2114, opened 2025-01-31
fix(CLI): not found errors 

### LINKED ISSUE #2112: CLI: Do not print out help if the response is a 404
Currently, if a CLI command is executed correctly, but its response is a 404, the error is returned, and automatically the help is printed. 

**Example:**

Here, I'm trying to describe a non-existent pipeline, and notice how help is automatically printed out.

```bash
$ conduit pipelines describe foo
Error: failed to get pipeline: rpc error: code = NotFound desc = failed to get pipeline by ID: pipeline instance not found (ID: foo): pipeline instance not found
Usage:
  conduit pipelines describe [flags]

Aliases:
  describe, desc

Examples:
conduit pipelines describe pipeline-with-dlq
conduit pipelines desc multiple-source-with-processor

Flags:
  -h, --help   help for describe

Global Flags:
      --api.grpc.address string   address where Conduit is running
      --config.path string        path to the configuration file

failed to get pipeline: rpc error: code = NotFound desc = failed to get pipeline by ID: pipeline instance not found (ID: foo): pipeline instance not found
exit status 1
```

**Expected output**

```
$ conduit pipelines describe foo
pipeline instance not found (ID: foo): pipeline instance not found
```

## What to do

We'll need to not show help if the error is expected. Only show help when there is unexpected behavior (wrong flags, wrong args, etc.).

raulb:
### Description

Fixes #2112 

### Quick checks


cmd/conduit/cecdysis/decorators.go
```diff
@@ -57,24 +60,44 @@ func (CommandWithExecuteWithClientDecorator) Decorate(_ *ecdysis.Ecdysis, cmd *c
...
+
+		// st.Message() is already an entire representation of the error
+		// need to grab the desc
+		descIndex := strings.Index(errMsg, "desc =")
+		if descIndex != -1 {
+			errMsg = errMsg[descIndex+len("desc = "):]
+		}
+	}
+	_, _ = fmt.Fprintf(os.Stderr, "%v\n", errMsg)
+	return nil
```
lovromazgon, line 97: Wouldn't this mean that the exit code will be 0, i.e. no error? 🤔

raulb: @lovromazgon that's correct. This is the only way to ensure usage is not shown. The error bubbles up to cobra and if there's an error and we're not silencing usage, this will be shown. I actually just thought of something that I'm going to tru.

lovromazgon: I'm thinking if we should rather set `SilenceUsage` to `true`, maybe even in ecdysis, so that it's the default (see https://github.com/spf13/cobra/issues/340#issuecomment-243811964). Apparently the usage would still get displayed if `--help` is added or when the subcommand is missing, which is what we're looking for, right? Incorrect CLI usage results in usage, otherwise we silence the usage and display the returned error.

raulb: Need to investigate further, but the immediate consequence is that usage is not shown on other validations we do make (flags, args)... maybe that's still ok

```
$ conduit pipelines desc 
Error: requires a pipeline ID
exit status 1
```

raulb: I think I'm ok with this https://github.com/ConduitIO/conduit/pull/2114/commits/c0d321d05295f8df6d0d3b9e8caf39ae7a6e5f87. Much simpler.

lovromazgon: I see, I also tried this:

```
$ ./conduit foo
Error: unknown command "foo" for "conduit"
Run 'conduit --help' for usage.
```

Not quite what I expected, but I think it's fine 🤷 It's a clear output and a good compromise, given the simplicity of the code change.

lovromazgon (approved):

PR #2117, opened 2025-02-03
fix init bug & docs

maha-hajja:
### Description

found a bug while testing `conduit init`, it creates the folders under the wrong directory.
Also, fixed some docs 

### Quick checks

### DIFF
--- a/cmd/conduit/root/initialize/init.go
+++ b/cmd/conduit/root/initialize/init.go
@@ -61,10 +61,9 @@ func (c *InitCommand) Docs() ecdysis.Docs {
 func (c *InitCommand) createDirs() error {
 	// These could be used based on the root flags if those were global
 	dirs := []string{"processors", "connectors", "pipelines"}
-	conduitPath := filepath.Dir(c.flags.Path)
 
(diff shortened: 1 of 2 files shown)


cmd/conduit/root/initialize/init.go
```diff
@@ -61,10 +61,9 @@ func (c *InitCommand) Docs() ecdysis.Docs {
 func (c *InitCommand) createDirs() error {
 	// These could be used based on the root flags if those were global
 	dirs := []string{"processors", "connectors", "pipelines"}
-	conduitPath := filepath.Dir(c.flags.Path)
```
hariso, line 64: c.flags.Path should be the path to conduit.yaml right @raulb ? That's why we're taking the directory from that path.

raulb: @hariso actually, no. This change is correct. We have `config.path` which is global and indicates the full path of the configuration file (including `conduit.yaml`). This one in `init` indicates the root directory where the whole conduit structure would be created.

raulb (approved):

PR #2180, opened 2025-02-27
Haris/custom conduit with cli

### LINKED ISSUE #2150: Provide example of custom Conduit with the CLI included
### Feature description

Currently, our example/recommendation to implement a custom Conduit instance (e.g. with built-in connectors) is this:

```go
package main

import (
	http "github.com/conduitio-labs/conduit-connector-http"
	"github.com/conduitio/conduit/pkg/conduit"
)

func main() {
	// Get the default configuration, including all built-in connectors
	cfg := conduit.DefaultConfig()

	// Add the HTTP connector to list of built-in connectors
	cfg.ConnectorPlugins["http"] = http.Connector

	conduit.Serve(cfg)
}
```

(Taken from https://conduit.io/docs/using/connectors/additional-built-in-plugins/)

However, with the introduction of CLI, the example above won't include the CLI. 

The goal of this issue is to provide a way to build a custom Conduit with the CLI included.

hariso:
### Description

Fixes #2150.

### Quick checks


raulb (commented):
Why is needed to specify a configuration rather than relying on ecdysis to parse it?

cmd/conduit/root/root.go
```diff
@@ -85,7 +86,7 @@ func (c *RootCommand) Docs() ecdysis.Docs {
 }
 
 func (c *RootCommand) SubCommands() []ecdysis.Command {
-	runCmd := &run.RunCommand{}
+	runCmd := &run.RunCommand{Cfg: c.DefaultCfg}
```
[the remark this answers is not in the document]
raulb: @hariso Nevermind. Looking at https://github.com/hariso/custom-conduit/blob/main/main.go, I understand now why this is needed.

cmd/conduit/root/root.go
```diff
@@ -50,8 +50,9 @@ type RootFlags struct {
 }
 
 type RootCommand struct {
-	flags  RootFlags
-	output ecdysis.Output
+	DefaultCfg conduit.Config
```
raulb, line 53: I personally think this would look better if the field were named simply `Cfg`. This way, it doesn't necessarily imply that the configuration used here is the default (it'll most likely be), but technically speaking, it could be any other configuration.

```suggestion
	Cfg conduit.Config
```

cmd/conduit/root/root.go
raulb, line 89:
```suggestion
	runCmd := &run.RunCommand{Cfg: c.Cfg}
```

cmd/conduit/cli/cli.go
```diff
@@ -0,0 +1,41 @@
...
+	"github.com/conduitio/conduit/cmd/conduit/cecdysis"
+	"github.com/conduitio/conduit/cmd/conduit/root"
+	"github.com/conduitio/conduit/pkg/conduit"
+	"github.com/conduitio/ecdysis"
+)
+
+func Run(cfg conduit.Config) {
+	e := ecdysis.New(ecdysis.WithDecorators(cecdysis.CommandWithExecuteWithClientDecorator{}))
+
+	cmd := e.MustBuildCobraCommand(&root.RootCommand{DefaultCfg: cfg})
```
raulb, line 29:
```suggestion
	cmd := e.MustBuildCobraCommand(&root.RootCommand{Cfg: cfg})
```

raulb (approved):
Looks great. Only left one naming suggestion.

cmd/conduit/root/root.go
hariso: I like the suggestion, I'll commit it.:)

Issue #2145, opened 2025-02-09
OpenAI Embeddings processor


simonl2002:
### Feature description

https://platform.openai.com/docs/guides/embeddings
https://platform.openai.com/docs/api-reference/embeddings

alarbada:
On it.

PR #2182, opened 2025-02-28
Add openai embeddings processor

alarbada:
### Description

Adds openai embeddings processor as a builtin plugin.

Fixes https://github.com/ConduitIO/conduit/issues/2145.

### Quick checks


pkg/plugin/processor/builtin/impl/openai/embeddings/embeddings.go
```diff
@@ -0,0 +1,164 @@
...
+type openaiCall interface {
+	Call(ctx context.Context, input string) ([]float32, error)
+}
+
+type embeddingsProcessorConfig struct {
+	openaiwrap.Config
+
+	// Field is the reference to the field to process. Defaults to ".Payload.After".
+	Field string `json:"field" default:".Payload.After"`
+	// APIKey is the OpenAI API key. Required.
```
raulb, line 49: Removing the `required` bit from here to leave validations indicate that. This would require generating the json again.

```suggestion
	// APIKey is the OpenAI API key.
```

pkg/plugin/processor/builtin/impl/openai/embeddings/embeddings.go
```diff
@@ -0,0 +1,164 @@
...
+	APIKey string `json:"api_key" validate:"required"`
+	// Model is the OpenAI embeddings model to use (e.g., text-embedding-3-small). Required.
```
raulb, line 51: Removing the `required` bit from here to leave validations indicate that. This would require generating the json again.

```suggestion
	// Model is the OpenAI embeddings model to use (e.g., text-embedding-3-small).
```

pkg/plugin/processor/builtin/impl/openai/embeddings/embeddings.go
```diff
@@ -0,0 +1,164 @@
...
+	Model string `json:"model" validate:"required"`
+	// Dimensions is the number of dimensions the resulting output embeddings should have.
+	Dimensions int `json:"dimensions"`
+	// EncodingFormat is the format to return the embeddings in. Can be "float" or "base64".
+	EncodingFormat string `json:"encoding_format"`
+	// User is the user identifier for OpenAI API.
+	User string `json:"user"`
```
raulb, line 58: Per https://platform.openai.com/docs/api-reference/users/object, I wonder if maybe we could be more explicit and call this `UserID`. What do you think?

```suggestion
	User string `json:"user"`
```

pkg/plugin/processor/builtin/impl/openai/embeddings/embeddings.go
```diff
@@ -0,0 +1,164 @@
...
+		sdk.Logger(ctx).Warn().Msg("openai API call was overriden with a custom implementation")
+	}
+
+	return nil
+}
+
+func (p *embeddingsProcessor) Specification() (sdk.Specification, error) {
+	return sdk.Specification{
+		Name:        "openai-embeddings",
+		Summary:     "generate embeddings for records using openai models",
```
raulb, line 94:
```suggestion
		Summary:     "Generate embeddings for records using OpenAI models.",
```

pkg/plugin/processor/builtin/impl/openai/embeddings/embeddings.go
```diff
@@ -0,0 +1,164 @@
...
+		Description: "embeddings is a conduit processor that will generate vector embeddings for a record using OpenAI's embeddings API",
```
raulb, line 95:
```suggestion
		Description: "Embeddings is a Conduit processor that will generate vector embeddings for a record using OpenAI's embeddings API.",
```

pkg/plugin/processor/builtin/impl/openai/embeddings/embeddings.go
```diff
@@ -0,0 +1,164 @@
...
+		EncodingFormat: openai.EmbeddingEncodingFormat(o.config.EncodingFormat),
+		Dimensions:     o.config.Dimensions,
+		User:           o.config.User,
+	})
+	if err != nil {
+		return nil, err
+	}
+
+	if len(resp.Data) > 0 {
+		embeddings = resp.Data[0].Embedding
```
raulb, line 160: Could `resp.Data` be nil in any situation?

pkg/plugin/processor/builtin/impl/openai/embeddings/embeddings_examples_test.go
```diff
@@ -0,0 +1,90 @@
+// Copyright © 2024 Meroxa, Inc.
```
raulb, line 1:
```suggestion
// Copyright © 2025 Meroxa, Inc.
```

pkg/plugin/processor/builtin/impl/openai/openai.go
```diff
@@ -0,0 +1,195 @@
...
+
+// Config contains common configuration options for OpenAI processors
+type Config struct {
+	// MaxRetries is the maximum number of retries for API calls. Defaults to 3.
+	MaxRetries int `json:"max_retries" default:"3"`
+	// InitialBackoff is the initial backoff duration in milliseconds. Defaults to 1000ms (1s).
+	InitialBackoff int `json:"initial_backoff" default:"1000"`
+	// MaxBackoff is the maximum backoff duration in milliseconds. Defaults to 30000ms (30s).
+	MaxBackoff int `json:"max_backoff" default:"30000"`
+	// BackoffFactor is the factor by which the backoff increases. Defaults to 2.
```
raulb, line 37:
```suggestion
	// BackoffFactor is the factor by which the backoff increases. Defaults to 2.0
```

raulb (changes requested):
@alarbada Only left some small cosmetic things. Everything else looks great. Apologies for the slow review here.

pkg/plugin/processor/builtin/impl/openai/embeddings/embeddings.go
alarbada: I'd rather not, I want to model exactly how the request object is from openai.

Here's the request object property:
https://platform.openai.com/docs/api-reference/embeddings/create#embeddings-create-user

pkg/plugin/processor/builtin/impl/openai/embeddings/embeddings.go
alarbada: I don't think so, I wrote this because it felt a bit panicky to just write it with the raw index access. In theory it should never happen, in all my experience using OpenAI APIs the data object has always one json object inside.

We can remove it if it feels better.

pkg/plugin/processor/builtin/impl/openai/embeddings/embeddings.go
raulb: @alarbada Makes sense.

pkg/plugin/processor/builtin/impl/openai/embeddings/embeddings.go
raulb: @alarbada that's fine. Was wondering if we needed to guard against `resp.Data[0]` being nil.

raulb (approved):
:shipit:

Issue #1875, opened 2024-10-02
Feature: Ability to update pipelines provisioned by config files


nickchomey:
### Feature description

It is much easier/faster to define a pipeline file config files than to do so via the Web UI (too many clicks, and various bugs like buttons sporadically not working) or HTTP API (would require setting up scripts to run all the http requests and their varied payloads etc...). However, file configurations are immutable, so you have to restart the Conduit server in order to change them.

It would be great if we could update the config file and the changes would get reflected immediately. 

I dont know what limitations there are with this compared to the HTTP API, but perhaps [fsnotify](https://github.com/fsnotify/fsnotify) could be used to monitor the `pipelines` and maybe even the `connectors` directories?

lovromazgon:
This feature would be great to have, but we currently have higher priority items on our roadmap. Although, if you wanted to add this feature we are happy to assist you with a review!

lovromazgon:
The way we treat this issue when Conduit starts is to do a diff between the pipeline Conduit finds in its store (e.g. badger DB) and the pipeline in the config file. The assumption is that the config file is the source of truth, so if an entity (source, destination, processor, pipeline) can't be found in the file we delete it, if a new entity is found we create it, if an existing one is changed we update it. We match entities using the IDs, so let's say changing the ID of a source in the pipeline config file will result in the existing source being deleted and a new one being created instead, consequently the pipeline will start from scratch because the position state will be removed together with the source (as explained [here](https://conduit.io/docs/pipeline-configuration-files/provisioning#updating-a-provisioned-pipeline)).

That's the same behavior I'd expect to see if we implement hot-reloads of pipeline config files. In my opinion, Conduit shouldn't bother with a diff or user guided reconciliation, that sounds like something that can be handled separately by establishing proper deploy procedures, using git and separate deploy environments to ensure a config works correctly before it hits production. We need to see Conduit as the low-level tool that it is, focused on moving data. So I'm more for detecting file changes and treating them as the new source of truth, Conduit then stops the existing pipeline, applies the changes and restarts the pipeline.

PR #2236, opened 2025-04-04
Allow for hot-reloading yaml config pipelines

nickchomey:
### Description

This PR adds functionality to allow for upserting (update if exists, otherwise create) and hot-reloading (graceful shutdown of running pipeline) of yaml-config pipelines - no need to shut down the entire server anymore. This will enable the use of a file watcher to solve #1875 

It also accepts a yaml string, rather than just a yaml file, which will allow for me to use yaml configs in the NATS api that I am working on (and you could add a similar endpoint for the grpc and http apis if you like), without any files at all.

I don't think this has any side effects on Conduit, and as far as I can tell it wont cause any of its own issues. However, I assume there's plenty of ways in which this could be improved - be it refactoring, allowing other existing functionality to leverage this, etc... 

It does not add any api endpoints, schema etc for using this - i think that's beyond my pay grade ;) 

Any feedback and guidance to improve this would be quite appreciated!

### Quick checks

### DIFF
--- a/pkg/provisioning/service.go
+++ b/pkg/provisioning/service.go
@@ -67,8 +67,8 @@ func NewService(
 	}
 }
 
-// parseYamlInput parses a YAML input (either a file path or a YAML string) and returns the first pipeline configuration.
-func (s *Service) parseYamlInput(ctx context.Context, yamlInput string) (config.Pipeline, string, error) {
+// parseYamlInput parses a YAML input (either a file path or a YAML string) and returns all pipeline configurations.
+func (s *Service) parseYamlInput(ctx context.Context, yamlInput string) ([]config.Pipeline, string, error) {
 	// First, determine if the input is a file path or a YAML string
 	isFilePath := false
 	if !strings.Contains(yamlInput, "\n") && len(yamlInput) < 1024 {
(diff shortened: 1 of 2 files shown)


hariso:
@nickchomey can you explain briefly how you actually connect NATS and Conduit, so that NATS is delivering the pipeline configs to Conduit? 

The team talked about this yesterday, and it appears to us that the goal is actually to be able to "refresh" a pipeline configuration or add a new one, without having to restart the Conduit service. If that's true, then we find that useful for sure and it would be a really useful feature in Conduit.

However, watching a file might not be the best way, because it may trigger unnecessary updates (e.g. an editor might be auto-saving while you're working on a file, and there might be other reasons too). Automated updates can backfire pretty easily, which is why we think that the actual reload should be triggered by a user. Currently, that trigger is a service restart. We propose to have an API endpoint that a user can use to reload a pipeline configure file(s). 

In other words the flow would be:

1. A pipeline configuration file is added/updated/removed.
2. User calls `/v1/pipelines/reload` (just an example) that reloads either all or just targeted pipelines.

I can open a discussion where we can discuss the details. What do you think about this? Would it be helpful for you?

Issue #2290, opened 2025-05-19
Feature: Allow processors to split records


lovromazgon:
### Feature description

A processor can currently only return 1 output record for 1 input record (or 0 if it's filtering it). This issue tracks the work that enables processors to split input records into N records.

raulb:
@lovromazgon we can close this one, right?

lovromazgon:
Oh, right, this was implemented in https://github.com/ConduitIO/conduit/pull/2292.

PR #2292, opened 2025-05-19
[arch-v2] Multirecord

lovromazgon:
### Description

This PR refactors the `funnel` package a bit and adds support for mutlirecords (i.e. processors splitting records). It also adds a bunch of tests.

Some things to note:
- The `Ack`, `Nack`, `Filter` and `Retry` functions on `funnel.Batch` were made a bit smarter and take into account filtered records. In other words, if a batch contains 5 records, but 3 of them are filtered, the processor will receive only 2 records. When we then receive back 2 records and mark the new status via one of those functions, we don't need to worry about adjusting the index based on where the filtered records are, as that will happen automatically. There is a caveat - we need to mark filtered records from the end of the batch towards the start, as filtering a record changes the indices for the next call (see the `ProcessorTask` and how it marks records in reverse).
- The DLQ had a small bug where it wasn't possible to disable the DLQ with a `0` window size, it was fixed.
- If a processor returns a `MultiRecord` slice with 0 records it's the same as filtering the record.
- If a processor returns a `MultiRecord` slice with 1 record, it's the same as just operating on that single record (no split).

Depends on https://github.com/ConduitIO/conduit-processor-sdk/pull/137
Related to #2290 

### Quick checks

- [ ] I have made sure that the PR is of reasonable size and can be easily reviewed. (oh well, it's a big change)


go.mod
```diff
@@ -16,8 +16,8 @@ require (
 	github.com/conduitio/conduit-connector-postgres v0.12.2
 	github.com/conduitio/conduit-connector-protocol v0.9.4-0.20250522154900-3708d4f3a481
 	github.com/conduitio/conduit-connector-s3 v0.9.2
-	github.com/conduitio/conduit-connector-sdk v0.13.5-0.20250522155326-e7f71d4ed767
-	github.com/conduitio/conduit-processor-sdk v0.4.4
+	github.com/conduitio/conduit-connector-sdk v0.13.4
+	github.com/conduitio/conduit-processor-sdk v0.4.5-0.20250519135306-105cf0cc0974
```
lovromazgon, line 20: Don't forget to update these values before merging!

Issue #2253, opened 2025-04-15
Conduit shouldn't recursively traverse the pipelines directory


hariso:
### Feature description

This came up in Discord https://discord.com/channels/828680256877363200/1360147516703637666/1360591403469770902.

rauanmayemir:
If pipelines directory is mounted from ConfigMaps in Kubernetes environment with each configmap key mapped to a file, the parsed pipeline won't be started due to `ERR provisioning failed error="2 pipelines with ID \"pipeline_name\" will be skipped: duplicated pipeline ID"`.

The reason for that is in the way kubernetes mounts configmaps:
![Image](https://github.com/user-attachments/assets/caeae9c4-8304-42ff-8e48-149d3e4ad6c1)

So when conduit traverses the directory it reads the mounted yaml file and then also resolves the hidden symlink causing duplicate error.

lovromazgon:
Sidenote: we could reuse the code for loading connector plugins, that part doesn't traverse sub-directories.

lyuboxa:
> Sidenote: we could reuse the code for loading connector plugins, that part doesn't traverse sub-directories.

This and also when walking the dir omit any files that don't have the right extension and only load up `.yml` and `.yaml`

rauanmayemir:
@hariso I've just tried this and I don't believe it's fixed. For a mounted configmap with my pipelines, conduit just ignored everything and deleted all pipelines. I had to revert back to copying yaml's to a separate dir.

hariso:
@rauanmayemir Thanks for reporting back! I just checked the code and your screenshot above. It looks like the code checks if a file is "regular", i.e., it skips symbolic links. I'll fix that.

hariso:
@rauanmayemir the fix is now in the `main` branch.

PR #2298, opened 2025-05-21
 Read sym-linked pipeline configuration files 

hariso:
### Description

Adds support for reading symbolic links in the pipelines directory. Reported in https://github.com/ConduitIO/conduit/issues/2253#issuecomment-2897877322.

### Quick checks


pkg/provisioning/service.go
```diff
@@ -170,39 +170,79 @@ func (s *Service) Delete(ctx context.Context, id string) error {
...
+		return "", cerrors.Errorf("could not read symlink: %w", err)
+	}
+
+	// If symlink path is relative, make it absolute
+	if !filepath.IsAbs(resolvedPath) {
+		resolvedPath = filepath.Join(basePath, resolvedPath)
+	}
+
+	// Check if the resolved path exists and is a regular file
+	_, err = os.Stat(resolvedPath)
```
hariso, line 224: We could go one step further, and then check if the resolved path is a symbolic itself, but that feels like over-engineering.:)

pkg/provisioning/service.go
```diff
@@ -170,39 +170,79 @@ func (s *Service) Delete(ctx context.Context, id string) error {
 // files that end with .yml or .yaml. It only reads files in the top-level
 // directory and skips non-regular files.
 func (s *Service) getYamlFiles(path string) ([]string, error) {
+	ctx := context.Background()
```
lovromazgon, line 173: We have a context in the caller, I'd just add it to the function parameters and get it from there.

pkg/provisioning/service.go
```diff
@@ -170,39 +170,79 @@ func (s *Service) Delete(ctx context.Context, id string) error {
...
+	info, err := os.Stat(path)
+	if err != nil {
+		return false
+	}
+	if !info.Mode().IsRegular() {
+		return false
+	}
+
+	// Check if it's a YAML file
+	ext := filepath.Ext(path)
```
lovromazgon, line 242:
```suggestion
	ext := strings.ToLower(filepath.Ext(path))
```

lovromazgon (approved):
Re over-engineering: I feel like this already on the verge 🙃  We have the same logic for [loading standalone connectors](https://github.com/ConduitIO/conduit/blob/8b2ef3a219ad830024655311420d3e88b90b53e7/pkg/plugin/connector/standalone/registry.go#L86) that accepts symlinks and doesn't traverse deeper with a fraction of the code. That said, let's leave it as it is now, but if we find another bug, let's agree to mimic the code we have there rather than adding more fixes.

PR #2326, opened 2025-06-04
add schema registry basic and bearer authentication

### LINKED ISSUE #2275: Feature: add schema registry authentication settings
### Feature description

I've just realized that schema registry doesn't support auth settings.

It would be great to have extra options like:
```
--schema-registry.confluent.connection-string "redpanda-endpoint:8081"
--schema-registry.type confluent
--schema-registry.authentication.type basic
--schema-registry.authentication.basic.username "username"
--schema-registry.authentication.basic.password "password"
```

The reason why I suggest authentication parameters to be 'outside' confluent is because both redpanda and apicurio implement confluent protocol, but add authentication on top of it.

maha-hajja:
### Description

Support Basic and Bearer authentication for schema registry.

Fixes #2275 

### Quick checks


pkg/conduit/config.go
```diff
@@ -219,6 +226,19 @@ func (c Config) validateSchemaRegistryConfig() error {
 		if c.SchemaRegistry.Confluent.ConnectionString == "" {
 			return requiredConfigFieldErr("schema-registry.confluent.connection-string")
 		}
+		if c.SchemaRegistry.Confluent.Authentication.Type == "basic" {
```
hariso, line 229: Nitpick: it would be good to fail if the type is something other than basic or bearer.

hariso (approved):

PR #2330, opened 2025-06-10
 Support reading files in pipelines.path 

hariso:
### Description

This fixes a regression where the `pipelines.path` parameter doesn't support file paths anymore.

### Quick checks

### DIFF
--- a/pkg/provisioning/service.go
+++ b/pkg/provisioning/service.go
@@ -179,6 +179,45 @@ func (s *Service) getYamlFiles(ctx context.Context, path string) ([]string, erro
 		Str("pipelines_path", path).
 		Msg("loading pipeline configuration files")
 
+	info, err := os.Stat(path)
+	if err != nil {
+		return nil, cerrors.Errorf("failed to stat pipelines path %q: %w", path, err)
+	}
+	if !info.IsDir() {
+		return s.getYamlFile(path)
+	}
+
+	return s.getYamlFilesFromDir(ctx, path)
+}
+
+func (s *Service) getYamlFile(path string) ([]string, error) {
+	info, err := os.Lstat(path)
+	if err != nil {
+		return nil, cerrors.Errorf("failed to stat pipelines path %q: %w", path, err)
+	}
+
+	dir := filepath.Dir(path)
+	filename := filepath.Base(path)
+	filePath := path
+
+	// If it's a symlink, resolve it
+	if info.Mode()&os.ModeSymlink != 0 {
+		resolvedPath, err := s.resolveSymLink(dir, filename)
+		if err != nil {
+			return nil, cerrors.Errorf("could not resolve symlink %q: %w", path, err)
+		}
+
+		filePath = resolvedPath
+	}
+
+	if s.isYamlFile(filePath) {
+		return []string{filePath}, nil
+	}
+
+	return nil, cerrors.Errorf("%v is not a yaml file", filePath)
+}
+
+func (s *Service) getYamlFilesFromDir(ctx context.Context, path string) ([]string, error) {
 	dirEntries, err := os.ReadDir(path)
 	if err != nil {
 		s.logger.Warn(ctx).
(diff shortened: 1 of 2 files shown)


pkg/provisioning/service.go
```diff
@@ -179,6 +179,45 @@ func (s *Service) getYamlFiles(ctx context.Context, path string) ([]string, erro
...
+		return nil, cerrors.Errorf("failed to stat pipelines path %q: %w", path, err)
+	}
+	if !info.IsDir() {
+		return s.getYamlFile(path)
+	}
+
+	return s.getYamlFilesFromDir(ctx, path)
+}
+
+func (s *Service) getYamlFile(path string) ([]string, error) {
```
lovromazgon, line 193: Do this:
```sh
mkdir -p pipelines-example/subfolder
echo "version: 2.2" > pipelines-example/subfolder/pipeline
ln -s subfolder/pipeline pipelines-example/pipeline.yml
```

Then run conduit using `./conduit run --pipelines.path=pipelines-example` and you'll see it works. ✅ 
Then run `./conduit run --pipelines.path=pipelines-example/pipeline.yml` and you'll see this error ❌:
```
ERR provisioning failed error="failed to read pipelines folder \"pipelines-example/pipeline.yml\": pipelines-example/subfolder/pipeline is not a yaml file" stack=[{"file":"/Users/lovro/dev/conduitio/conduit/pkg/provisioning/service.go","func":"github.com/conduitio/conduit/pkg/provisioning.(*Service).Init","line":78},{"file":"/Users/lovro/dev/conduitio/conduit/pkg/provisioning/service.go","func":"github.com/conduitio/conduit/pkg/provisioning.(*Service).getYamlFile","line":217}]
```

I think this all stems from the fact that we resolve symlinks manually. Can we simplify this and only work with the path the user gives us, without resolving the symlink manually? Afaik `os` will resolve it for us when we read the file. Why did we start manually resolving symlinks in the first place?

Also I'll point back to my [comment](https://github.com/ConduitIO/conduit/pull/2298#pullrequestreview-2861167853) in the old PR:
> We have the same logic for [loading standalone connectors](https://github.com/ConduitIO/conduit/blob/8b2ef3a219ad830024655311420d3e88b90b53e7/pkg/plugin/connector/standalone/registry.go#L86) that accepts symlinks and doesn't traverse deeper with a fraction of the code. That said, let's leave it as it is now, but if we find another bug, let's agree to mimic the code we have there rather than adding more fixes.

hariso: > Why did we start manually resolving symlinks in the first place?

It's for validation purposes: when we resolve a symlink, we can check if it's something we accept as a pipeline file or not (a different type of a file, a directory, etc.)

> We have the same logic for [loading standalone connectors](https://github.com/ConduitIO/conduit/blob/8b2ef3a219ad830024655311420d3e88b90b53e7/pkg/plugin/connector/standalone/registry.go#L86) that accepts symlinks and doesn't traverse deeper with a fraction of the code. That said, let's leave it as it is now, but if we find another bug, let's agree to mimic the code we have there rather than adding more fixes.

Yup, I remembered that one, but I didn't go back to it since this part is about handling a path that points to a file. We only accept a directory as a connectors' path, whereas a pipeline path can be a directory, but also a file.

pkg/provisioning/service.go
```diff
@@ -208,27 +213,6 @@ func (s *Service) getYamlFiles(ctx context.Context, path string) ([]string, erro
...
-	// Check if the resolved path exists and is a regular file
-	_, err = os.Stat(resolvedPath)
-	if err != nil {
-		return "", cerrors.Errorf("could not stat symlink: %w", err)
-	}
-
-	return resolvedPath, nil
-}
-
 func (s *Service) isYamlFile(path string) bool {
```
lovromazgon, line 216: How about we pass the whole `dirEntry` in here and use `dirEntry.IsDir()` to determine if it's a file or a dir?

lovromazgon (approved):
LGTM, left one nitpick, but that's it 👍

PR #2332, opened 2025-06-11
Clone processor

lovromazgon:
### Description

This adds a `clone` processor which allows you to clone records in the pipeline.

Note that it also updates the postgres connector to the latest version in main, which includes support for composite keys.

### Quick checks


go.mod
```diff
@@ -13,7 +13,7 @@ require (
 	github.com/conduitio/conduit-connector-generator v0.10.4
 	github.com/conduitio/conduit-connector-kafka v0.12.3
 	github.com/conduitio/conduit-connector-log v0.7.3
-	github.com/conduitio/conduit-connector-postgres v0.13.0
+	github.com/conduitio/conduit-connector-postgres v0.13.1-0.20250611110928-9da30f9cd4b4
```
raulb, line 16: No needed, right?

pkg/plugin/processor/builtin/impl/clone.go
```diff
@@ -0,0 +1,89 @@
...
+		Name:    "clone",
+		Summary: "Clone records.",
+		Description: `Clone all records N times. All cloned records will have
+the same data as the original record, except for the metadata field
+` + "`clone.index`" + `, which will be set to the index of the clone (0-based).
+
+**Important:** Add a [condition](https://conduit.io/docs/using/processors/conditions)
+to this processor if you only want to clone some records.
+
+**Important:** This processor currently only works using the pipeline architecture
```
raulb, line 55: Nitpick (considering hopefully soon all of these will be v2 ready): I would probably start collecting what processors are under this v2 flag so we keep track of these easily. e.g.:

```golang
v2Processors := []string{"split", "clone"}
```

or something like this to easily access via processor name.

```golang
v2ProcessorsMap := map[string]bool{
     "split": true,
     "clone": true,
}
```

This way we could add this warning (only in one place) and add it to the specification if needed.

pkg/plugin/processor/builtin/impl/clone.go
```diff
@@ -0,0 +1,89 @@
...
+
+	config cloneConfig
+}
+
+func NewCloneProcessor(log.CtxLogger) *CloneProcessor { return &CloneProcessor{} }
+
+type cloneConfig struct {
+	// The number of records after cloning (e.g. if count is 3, the processor
+	// will output 3 records for every input record).
+	Count int `json:"count" validate:"required,gt=1"`
```
raulb, line 41: (If `Count` changes)

```suggestion
	Count int `json:"count" validate:"required,gt=0"`
```

pkg/plugin/processor/builtin/impl/clone.go
```diff
@@ -0,0 +1,89 @@
...
+type CloneProcessor struct {
+	sdk.UnimplementedProcessor
+
+	config cloneConfig
+}
+
+func NewCloneProcessor(log.CtxLogger) *CloneProcessor { return &CloneProcessor{} }
+
+type cloneConfig struct {
+	// The number of records after cloning (e.g. if count is 3, the processor
```
raulb, line 39: I think the count should be the number of times we're cloning (while preserving the original record we already had). In other words, `count = 1` would be two records in total after that.

I realize that we would also need to update where this is used to make sure we maintain its logic.

```suggestion
	// The number of times the record is cloned (e.g. if count is 2, the processor
```

lovromazgon: Oh right, we talked about that. I'll change it 👍 The one thing that's maybe a bit strange is that the original record will also be changed, as the metadata field `clone.index` will be added to it, even though it isn't a clone. But I don't think it's too bad, will try to document it clearly.

PR #2508, opened 2026-07-04
perf(metrics): estimate record size without json.Marshal (#2268)

### LINKED ISSUE #2268: [arch-v2] Investigate faster ways to obtain record size for metrics purposes
### Feature description

After profiling a few Conduit instances that were running with the new architecture enabled, we discovered that JSON serialization took a lot of time. Tracking metrics about record size took a significant part in that.

The reason for that appears to be that, when obtaining metrics for a record, we serialize the record into a JSON object first ([here](https://github.com/ConduitIO/conduit/blob/c1770bdda2fd221de24a937e47c7673e28a99872/pkg/lifecycle-poc/funnel/source.go#L103) and [here](https://github.com/ConduitIO/conduit/blob/main/pkg/lifecycle-poc/funnel/destination.go#L131)).

Running our benchmarks with the source metrics disabled showed an improvement of ~25%. Disabling metrics in both, the source and destination tasks in the new architecture, results in a ~50% increased throughput. For example:

| pipeline                   | metrics                           | throughput |
|----------------------------|-----------------------------------|------------|
| generator-chaos-structured | enabled                           | 40684.56   |
| generator-chaos-structured | disabled source metrics           | 48041      |
| generator-chaos-structured | disabled source and dest. metrics | 63490.6        |
| postgres-to-log-cdc        | enabled                           | 84632      |
| postgres-to-log-cdc        | disabled source metrics           | 108289     |
| postgres-to-log-cdc        | disabled source and dest. metrics | 121403.43     |

We should find a faster way to obtain a record's size for metrics. Until then, it probably makes sense for size metrics to be disabled.

devarismeroxa:
Closes #2268.

## Problem
`RecordBytesHistogram.SizeOf` measured each record's size for the bytes histogram by calling `Bytes()` on the key and payload. For `StructuredData`, `Bytes()` runs a full **`json.Marshal`** — allocating and encoding the entire structure on **every record**, purely to produce an observability number. Profiling in #2268 found this dominates metrics cost; disabling the size metric yielded up to **~50% more throughput** on `postgres-to-log-cdc`.

The root cause is in shared code (`pkg/foundation/metrics/metrics.go`), so it slows **both** the default engine (`pkg/lifecycle`) and the preview arch-v2 engine (`pkg/lifecycle-poc`).

## Fix
Replace the marshal with a traversal that estimates the JSON-encoded size without allocating the encoded output:
- **RawData** — exact byte length (unchanged, already cheap).
- **StructuredData** — walk the value summing key/value sizes; floats and ints format into a **stack buffer** via `strconv.Append*`, so the whole path is **allocation-free**.
- Any other `Data` implementation falls back to the exact (slower) `Bytes()` length.

The metric value is now an approximation (ignores string escaping, uses the shortest float form) — acceptable for a size histogram, and verified within **10%** of the real `json.Marshal` size on a representative CDC record.

## Benchmarks
Structured record (structured key + 8-field payload), `go test -bench SizeOf -benchmem`:

```
SizeOf (new)       212 ns/op      0 B/op    0 allocs/op
SizeOf via Bytes  1077 ns/op    865 B/op    6 allocs/op
```

**~5× faster, zero allocations.** Both benchmarks are committed so the win is reproducible in-repo.

## Tests
- Estimate is within 10% of the real JSON size on a mixed CDC record (strings, floats, bool, nested object, array, null).
- RawData measured exactly; nil record → 0; a simple object estimated exactly.
- Both engines' callers build and the metrics suite passes; `golangci-lint run` clean.

## Risk tier
Tier 2. No data-integrity invariant is in play — this is an internal metrics computation, not the wire format or any serialized state. The only user-visible effect is that the `conduit_connector_bytes` / `conduit_dlq_bytes` histograms now report an approximate (rather than exact-JSON) size for structured records.

### DIFF
--- a/pkg/foundation/metrics/metrics.go
+++ b/pkg/foundation/metrics/metrics.go
@@ -15,6 +15,8 @@
 package metrics
 
 import (
+	"fmt"
+	"strconv"
 	"time"
 
 	"github.com/conduitio/conduit-commons/opencdc"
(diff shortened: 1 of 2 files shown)


devarismeroxa:
Independent fresh-context review (per author-never-approves-own-PR) initially returned **request-changes** — two legitimate accuracy issues, now fixed:

- **Float formatting** — the first pass used `'g'`, but the opencdc encoder (goccy/go-json) uses `'f'` for |f| in [1e-6, 1e21) and `'e'` otherwise, so epoch-ms timestamps and large IDs (decoded as float64, ubiquitous in CDC) were under-counted 15–33%. Fixed to match; notably goccy does **not** trim the exponent leading zero (`1e-07`, not stdlib's `1e-7`), so strconv's raw `'e'` output matches directly.
- **String escaping** — was ignored, under-counting URLs with `&` (HTML-escaped to `&`) and escaped strings by up to 50%. Now counted exactly in a single allocation-free byte scan; `[]byte` fixed to base64 length.

Rather than loosen the "within 10%" claim, the estimate is now **byte-exact** against `StructuredData.Bytes()`, and the tests assert exact equality across the flagged payloads (URLs with `&`, escaped strings, boundary/epoch-ms floats, unicode, nesting) plus an `AllocsPerRun==0` regression guard.

Perf after the exactness work: **305 ns/op, 0 allocs** vs the old `Bytes()` path's **1078 ns/op, 865 B/op, 6 allocs** — ~3.5× faster, allocation-free, and now exact. Re-review in progress.

devarismeroxa:
Re-review returned **approve**. Both recommended nits are now taken and pinned with tests: backspace/form-feed are counted as their 6-byte `\u00XX` forms (goccy doesn't use the `\b`/`\f` shortcuts), and object keys are now escaped via `stringJSONLen` too. Estimate is byte-exact against the opencdc encoder for all JSON-decoded data; the only documented residual is U+2028/U+2029 (bounded, vanishingly rare, irrelevant to a size histogram). Merging once CI is green.

PR #2509, opened 2026-07-04
fix(provisioning): isolate invalid documents in multi-document YAML (#2255)

### LINKED ISSUE #2255: Bug: Provisioning of multiple pipelines in a yaml file fails if one of them has an invalid version
### Bug description

I was exploring how the multiple pipelines per yaml config worked and was trying to make it fail on one of them. 

When one has an invalid version, the whole provisioning fails with an error like `ERR provisioning failed error="could not parse file \"/home/nick/conduittest/pipelines/test.yaml\": parsing error: unrecognized version 4"`

Conversely, if you omit the connectors section from a pipeline, the others will provision and start and you just get this error `ERR provisioning failed error="pipeline \"yaml-test-pipeline-3\", error while provisioning: could not start the pipeline \"yaml-test-pipeline-3\": could not build nodes for pipeline yaml-test-pipeline-3: can't build pipeline without any source connectors"` for the bad pipeline. 

### Steps to reproduce

You could try using a config like this

```
---
version: 2.2
pipelines:
  - id: yaml-test-pipeline-1
    description: First pipeline created via YAML API
    connectors:
      - id: test-source-1
        type: source
        plugin: builtin:file
        settings:
          path: ./example.in
      - id: test-destination-1
        type: destination
        plugin: builtin:file
        settings:
          path: ./example-yaml-1.out
    processors: []
---
version: 2.2
pipelines:
  - id: yaml-test-pipeline-2
    description: Second pipeline created via YAML API
    connectors:
      - id: test-source-2
        type: source
        plugin: builtin:file
        settings:
          path: ./example.in
      - id: test-destination-2
        type: destination
        plugin: builtin:file
        settings:
          path: ./example-yaml-2.out
    processors: []
---
version: 4
pipelines:
  - id: yaml-test-pipeline-3
    description: Malformed pipeline with invalid version
    connectors:
      - id: test-source-3
        type: source
        plugin: builtin:file
        settings:
          path: ./example.in
      - id: test-destination-3
        type: destination
        plugin: builtin:file
        settings:
          path: ./example-yaml-3.out
    processors: []
```

### Version

0.13.4

devarismeroxa:
Closes #2255.

## Problem
A pipeline config file can contain several YAML documents (`---` separated), each with its own `version`. If **one** document had an unrecognized version (e.g. `version: 4`) — or any decoding error — `ParseConfigurations` returned an error for the **whole file**, discarding every valid pipeline in it. One typo took down unrelated pipelines.

## Fix
Isolate failures to the offending document:
- **`ParseConfigurations`** aggregates per-document errors (`cerrors.Join`) and keeps parsing. On a version-parse failure the version decoder has already consumed the document, so the configuration decoder is advanced past it (`Decode(&any)`) to keep the two decoders in sync. If they can't be resynced, parsing stops rather than risk misattributing documents to the wrong config.
- **`Parse` / `parsePipelineConfigFile` / `Service.Init`** propagate the successfully-parsed pipelines *alongside* the error, so valid documents are still provisioned while the bad ones are reported — matching the service's existing per-pipeline error-aggregation pattern.

## Behavior change
`Parser.Parse`/`ParseConfigurations` now return partial results with a non-nil error instead of `nil` + error. Existing callers/tests that check `err != nil` are unaffected (the error is still non-nil, now aggregated); the unrecognized-*minor*-version cases were already warnings (fallback), not errors, so they're unchanged.

## Tests
- `TestParser_V2_MultiDocument_IsolatesInvalidVersion` — a `[valid, version-4, valid]` stream parses the two valid documents (in order) and reports `unrecognized version 4`.
- `TestParser_V2_MultiDocument_AllValid` — guards the two-decoder sync change: a normal multi-document file still parses fully.
- Full `pkg/provisioning/...` suite passes with `-race`; `golangci-lint` clean.

## Risk tier
Tier 2 (provisioning/config path; not the runtime record path). The main subtlety is the decoder resynchronization, covered by the two tests above.

### DIFF
--- a/pkg/provisioning/config/yaml/parser.go
+++ b/pkg/provisioning/config/yaml/parser.go
@@ -112,6 +112,7 @@ func (p *Parser) ParseConfigurations(ctx context.Context, reader io.Reader) (Con
 				}
 				// the decoders can no longer be kept in sync, stop parsing the
 				// rest of the file rather than risk misattributing documents
+				p.logger.Warn(ctx).Err(skipErr).Msg("could not recover parser after an invalid document; any remaining documents in this file were not processed")
 				errs = append(errs, cerrors.Errorf("could not skip invalid document: %w", skipErr))
 				break
 			}
--- a/pkg/provisioning/config/yaml/parser.go
+++ b/pkg/provisioning/config/yaml/parser.go
@@ -71,11 +71,10 @@ func NewParser(logger log.CtxLogger) *Parser {
 
 func (p *Parser) Parse(ctx context.Context, reader io.Reader) ([]config.Pipeline, error) {
 	configs, err := p.ParseConfigurations(ctx, reader)
-	if err != nil {
-		return nil, err
-	}
-
-	return configs.ToConfig(), nil
+	// Return whatever parsed successfully alongside any per-document errors, so a
+	// single bad document doesn't discard the valid pipelines in the same file
+	// (see #2255). Callers that require strictness can check the error.
+	return configs.ToConfig(), err
 }
 
 func (p *Parser) ParseConfigurations(ctx context.Context, reader io.Reader) (Configurations, error) {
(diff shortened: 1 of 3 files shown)


devarismeroxa:
Independent fresh-context review (per author-never-approves-own-PR) **traced the conduitio/yaml decoder internals and empirically verified the two-decoder synchronization is correct** for every edge case — bad document first/middle/last, consecutive bad documents, non-recoverable decode errors, and syntax errors — confirming no valid pipeline can be dropped or misattributed. The implementation was sound.

The reviewer's request-changes was a **test-coverage gap**: the config-decode-error isolation path and the harder sync positions weren't pinned by tests (CLAUDE.md: "every bug fix ships with the test that would have caught it"). Addressed:
- Replaced the two initial tests with a table-driven suite covering every position × failure mode: unrecognized version (first/middle/last/consecutive), non-recoverable decode error (first/middle/consecutive), and a syntax error (stops parsing but keeps prior documents) — asserting valid documents parse in order in each case. All 9 sub-cases pass.
- Added an operator-facing warning when the parser can't resynchronize after a malformed document (review nit), so skipped-remainder isn't silent.

`go test -race ./pkg/provisioning/...` and `golangci-lint run ./pkg/provisioning/...` clean. Re-confirming with the reviewer.

PR #2510, opened 2026-07-04
fix(provisioning): don't delete embedder-provisioned pipelines on restart (#1274)

### LINKED ISSUE #1274: Bug: Pipelines created programmatically are deleted on restart
### Bug description

When using Conduit's provisioning API in an embedded way to provision pipeline, the provisioned pipelines 
are removed from runtime upon restart. My understanding is that on startup pipelines which are not available locally
as files, will be deleted.

### Steps to reproduce

1. Provision pipeline through the provisioning pkg (using `Import()`)
2. Wait for pipeline to start
3. Stop conduit process
4. Start conduit process
5. Provisioned pipeline will be deleted.

### Version

latest

devarismeroxa:
Closes #1274.

## Problem
`createPipelineAction` hardcoded `pipeline.ProvisionTypeConfig`, so **every** pipeline created through `Import` — including programmatic calls by applications **embedding** Conduit — was tagged as config-file-provisioned. On restart, `Init`'s `deleteOldPipelines` deletes any `ProvisionTypeConfig` pipeline not present in the config files, so embedder-provisioned pipelines were silently removed. (The HTTP/gRPC API path was fine — it tags `ProvisionTypeAPI` directly.)

## Fix
Thread the provision type through the import path:
- `createPipelineAction` gains a `provisionedBy` field, used in `Do()` instead of the hardcoded constant; `deletePipelineAction` (a type alias of it) reuses it on rollback.
- `actionsBuilder` carries the pipeline provision type; `newActionsBuilder` **defaults to `ProvisionTypeConfig`** so all existing behavior is preserved.
- The **public `Import`** now tags `ProvisionTypeAPI` (programmatic pipelines survive restart); the **config-file path** (`provisionPipeline`) calls the internal `importPipeline` with `ProvisionTypeConfig` so reconciliation still works.

Only the *pipeline* provision type matters: there's no independent connector/processor reconciliation sweep, and deleting a pipeline cascades to its connectors — so tagging the pipeline is sufficient (minimal blast radius).

## Tests
- `TestService_Import_ProvisionType` — table-driven: the public `Import` tags `API`, the config-file path tags `Config`. **Verified to fail against the previous hardcoded behavior** (the API case) and pass with the fix.
- Existing action-level tests updated for the new field; full `pkg/provisioning/...` suite passes; `golangci-lint` clean.

## Risk tier
Tier 2 (provisioning path, not the runtime record path). Behavior-preserving for config-file provisioning (still `Config`); the only change is that programmatic `Import` callers now correctly get `API`.

### DIFF
--- a/pkg/provisioning/import.go
+++ b/pkg/provisioning/import.go
@@ -25,14 +25,26 @@ import (
 	"github.com/google/go-cmp/cmp/cmpopts"
 )
 
-// Import takes a pipeline config and imports it into Conduit.
+// Import takes a pipeline config and imports it into Conduit. Pipelines imported
+// this way are tagged with ProvisionTypeAPI, so they are treated as
+// programmatically provisioned and are NOT removed by the config-file
+// reconciliation on restart (see #1274). The config-file provisioning path uses
+// importPipeline with ProvisionTypeConfig instead.
 func (s *Service) Import(ctx context.Context, newConfig config.Pipeline) error {
+	return s.importPipeline(ctx, newConfig, pipeline.ProvisionTypeAPI)
+}
+
+// importPipeline imports a pipeline config, tagging any newly created pipeline
+// with the given provision type.
+func (s *Service) importPipeline(ctx context.Context, newConfig config.Pipeline, provisionedBy pipeline.ProvisionType) error {
 	oldConfig, err := s.Export(ctx, newConfig.ID)
 	if err != nil && !cerrors.Is(err, pipeline.ErrInstanceNotFound) {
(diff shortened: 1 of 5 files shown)


devarismeroxa:
Re-review returned **approve**. B1 (delete-branch provision type) fixed and pinned by a builder-level test that fails without it; added `TestService_deleteOldPipelines_SkipsNonConfig` driving the actual sweep (the reported symptom's mechanism); N1 deferred to #2511 with softened wording. Update path confirmed to preserve `ProvisionedBy`. CI green, `-race` clean, lint 0 issues. Merging.

PR #2512, opened 2026-07-04
fix(provisioning): honor status:stopped for a previously-running pipeline (#2061)

### LINKED ISSUE #2061: Bug: Pipeline status `stopped` in a configuration file still runs the pipeline
### Bug description

This pipeline with the status `stopped` shouldn't run:

```yaml
version: "2.2"
pipelines:
  - id: example-pipeline
    status: stopped
    name: "test"
    connectors:
      - id: example-source
        type: source
        plugin: "generator"
        settings:
          # Generate field 'airline' of type string
          # Type: string
          # Optional
          format.options.airline: 'string'
          # Generate field 'scheduledDeparture' of type 'time'
          # Type: string
          # Optional
          format.options.scheduledDeparture: 'time'
          # The format of the generated payload data (raw, structured, file).
          # Type: string
          # Optional
          format.type: 'structured'
          # The maximum rate in records per second, at which records are
          # generated (0 means no rate limit).
          # Type: float
          # Optional
          rate: '1'
      - id: example-destination
        type: destination
        plugin: "file"
        settings:
          # Path is the file path used by the connector to read/write records.
          # Type: string
          # Optional
          path: './destination.txt'
```

This used to work so it seems there's been a regression.

### Steps to reproduce

1. Use the configuration pipeline that's above.
2. Run conduit `$ conduit run`
3. Notice this pipeline runs (it shouldn't).

### Version

v0.13.0-nightly.20250109

devarismeroxa:
Closes #2061.

## Problem
When Conduit loads a pipeline persisted as `StatusRunning`, it flips it to `StatusSystemStopped` — the marker the lifecycle service uses to **auto-resume** pipelines that were running before a restart. `provisionPipeline` handled a config `status: running` by starting the pipeline, but had **no handling for `status: stopped`**: it correctly skipped `Start`, yet left the pipeline as `StatusSystemStopped`. The subsequent `lifecycle.Init` then resumed it anyway — so a pipeline the operator marked `stopped` kept running and **writing to its destination** after a restart.

Flow:
1. `pipeline.Service.Init`: `StatusRunning` → `StatusSystemStopped`.
2. `provisionPipeline`: `cfg.Status == stopped` → no `Start`, but the marker is left as `SystemStopped`.
3. `lifecycle.Service.Init`: resumes **all** `SystemStopped` pipelines → runs it despite `status: stopped`.

## Fix
Handle the stopped case in `provisionPipeline`: when the config says `stopped` and the pipeline is `StatusSystemStopped`, convert it to `StatusUserStopped` so `lifecycle.Init` leaves it stopped. Done after the import transaction commits, alongside the existing start-on-running logic. Added `UpdateStatus` to the provisioning `PipelineService` interface (already implemented by `pipeline.Service`) and regenerated the mock.

Only the `SystemStopped` marker is converted — new/already-user-stopped pipelines are untouched, and `running` behavior is unchanged.

## Tests
- `TestService_provisionPipeline_StatusStopped` — a `SystemStopped` pipeline with a `stopped` config is converted to `UserStopped` and **never started** (gomock fails if `Start` is called). **Verified to fail without the fix** (missing `UpdateStatus` call).
- Full `pkg/provisioning/...` suite passes with `-race`; `golangci-lint` clean; generated mock committed.

## Risk tier
**Tier 1** — the bug caused unintended writes to a destination on the data path. The change is narrow (a status-marker conversion during provisioning, after commit, before lifecycle resume) and behavior-preserving for `running` and already-stopped pipelines. Requesting human maintainer sign-off per CLAUDE.md.

## Failure-mode analysis
- Crash between commit and `UpdateStatus`: the pipeline stays `SystemStopped` and would be resumed on the *next* restart — same as today's behavior, no worse; the next successful provision converts it. No data loss (at-least-once is unaffected; this only governs whether a stopped pipeline auto-resumes).
- `UpdateStatus` failure: returns an error from `provisionPipeline`, surfaced by `Init`'s per-pipeline error aggregation; the pipeline is not silently started.
- Does not touch ack/position/checkpoint logic or serialized formats.

### DIFF
--- a/pkg/provisioning/import_test.go
+++ b/pkg/provisioning/import_test.go
@@ -421,6 +421,38 @@ func TestService_deleteOldPipelines_SkipsNonConfig(t *testing.T) {
 	is.Equal(len(deleted), 0)
 }
 
+// TestService_provisionPipeline_StatusStopped is the regression test for #2061: a
+// pipeline whose config says `status: stopped` must not be resumed on restart. A
+// pipeline that was running before the restart is marked StatusSystemStopped (the
+// lifecycle auto-resume marker); provisioning must convert it to StatusUserStopped
+// so the lifecycle service leaves it stopped, and must never call Start.
+func TestService_provisionPipeline_StatusStopped(t *testing.T) {
+	is := is.New(t)
+	ctx := context.Background()
+	ctrl := gomock.NewController(t)
+	srv, pipSrv, _, _, _, _ := newTestService(ctrl, log.Nop())
+
+	// existing pipeline, was running before the restart -> now SystemStopped, with
+	// a config identical to the incoming one so the import produces no actions.
+	instance := &pipeline.Instance{
+		ID:            "p1",
+		Config:        pipeline.Config{Name: "p1"},
+		DLQ:           pipeline.DefaultDLQ,
+		ProvisionedBy: pipeline.ProvisionTypeConfig,
+	}
+	instance.SetStatus(pipeline.StatusSystemStopped)
+	cfg := config.Pipeline{ID: "p1", Status: config.StatusStopped}
+
+	pipSrv.EXPECT().Get(gomock.Any(), "p1").Return(instance, nil).AnyTimes()
+	// the fix: convert the SystemStopped marker to UserStopped so lifecycle.Init
+	// does not resume it. gomock has no Start expectation on lifecycleService, so
+	// the test fails if the pipeline is (wrongly) started.
+	pipSrv.EXPECT().UpdateStatus(gomock.Any(), "p1", pipeline.StatusUserStopped, "")
+
+	err := srv.provisionPipeline(ctx, cfg)
+	is.NoErr(err)
+}
+
 func TestActionsBuilder_PreparePipelineActions_Delete(t *testing.T) {
 	is := is.New(t)
 	logger := log.Nop()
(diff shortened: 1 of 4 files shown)


devarismeroxa:
Independent Tier-1 review returned **approve, no blocking issues**. It verified the four load-bearing properties rather than trusting them:
- **No startup race** — `runtime.go` runs `pipeline.Init` → `provisioning.Init` → `lifecycle.Init` sequentially; the `UpdateStatus` write lands before lifecycle reads status.
- **`SystemStopped`-only conversion is necessary and sufficient** — `lifecycle.Init` auto-resumes *only* `SystemStopped`, and `pipeline.Init` flips *only* `Running`→`SystemStopped`, so that's the sole state from which a stopped-config pipeline can be resumed. The guard can't over-reach (no genuinely-running pipeline exists by then).
- **Running / new-stopped behavior unchanged**; **no data-integrity invariant (1–7) touched** (status-label + resume only; the fix *strengthens* correctness by stopping unintended destination writes).
- Regression test confirmed to fail without the fix.

Addressed the review nits: expanded the test to pin that only `SystemStopped` is converted (UserStopped/Degraded left as-is, none started); filed the cosmetic status-label edge (Degraded/Recovering + stopped config) as #2513. Redundant `Get` confirmed harmless (in-memory, no staleness).

CI green. Given this is **Tier 1**, holding for explicit maintainer sign-off before merge per CLAUDE.md.

devarispbrown (approved):

devarispbrown (approved):

devarispbrown (approved):

PR #2562, opened 2026-07-06
fix(lifecycle-poc): tear down source on the fatal-error path (resource leak)

### LINKED ISSUE #2559: lifecycle-poc: source connector not torn down on fatal-error path (resource leak)
**Preview arch (PipelineArchV2) resource leak, surfaced while un-skipping the #1659 poc test (#2558).**

On a fatal source-read error, the poc funnel arch never tears down the source connector/plugin, unlike the default arch. Trace (all verified):
- `runPipeline` error/cleanup path calls `rp.w.Close()` (`pkg/lifecycle-poc/service.go:586`), not `rp.w.Stop()` (line 228, the graceful path).
- `Worker.Close` (`funnel/worker.go:208`) closes each task via `task.Close()` + the DLQ — but `SourceTask.Close` (`funnel/source.go:79`) is an explicit no-op (`// source is torn down in the worker on stop`).
- `Source.Teardown` is only called from `Worker.Stop` (`worker.go:187`), reached solely on the graceful-stop path.
- Contrast: the default arch's `SourceNode.Run` (`pkg/lifecycle/stream/source.go:~120`) tears the source down unconditionally via `defer`, regardless of exit path.

So on any fatal source error, the source plugin subprocess/connection leaks until Conduit exits. Standalone/WASM connectors make this a real subprocess/FD leak.

**Fix direction:** make source teardown unconditional on the poc error path — either `SourceTask.Close` actually tears the source down (and `Worker.Stop` stops double-tearing), or `Worker.Close` tears down the source. Mirror the default arch's `defer`-based guarantee. Needs a test asserting Teardown is called on the fatal-error path (the un-skipped #2558 test currently uses an optional-teardown mock precisely because the call is missing — tighten it back to required once fixed).

Found by: #2558 (`SourcePluginWithOptionalTeardown` + code comments mark the gap). Preview arch — fix before PipelineArchV2 becomes default.

devarismeroxa:
Closes #2559 — a resource leak in the preview arch (PipelineArchV2), to fix before it becomes default.

## The bug
On a **fatal source-read error**, the funnel Worker degrades the pipeline directly and never calls `Worker.Stop`. `Worker.Close` closed the tasks + DLQ but **not the source** (`SourceTask.Close` is a no-op), so the source connector/plugin was never torn down → its subprocess/connection leaks. The default arch tears the source down unconditionally via `defer`; the poc did not.

Traced (all four links verified): `runPipeline` error path → `w.Close()` (not `w.Stop()`); `Worker.Close` → `task.Close()` + DLQ; `SourceTask.Close` no-op; `Source.Teardown` only reachable from `Worker.Stop`.

## The fix
Teardown in `Stop` is **load-bearing** — it interrupts a blocked source `Read` on graceful stop — so it can't just move to `Close`. Instead, make teardown **idempotent** (`sync.Once` → `Worker.tearDownSource`) and call it from **both** `Stop` (graceful) and `Close` (every path, including the fatal error). The `Once` also correctly serializes the `Stop`/`Close` goroutine race so teardown happens exactly once.

## Regression test
`TestServiceLifecycle_PipelineError` (lifecycle-poc, un-skipped in #2558) now **requires** the source `Teardown` — reverting the #2558 `SourcePluginWithOptionalTeardown` workaround. **Verified: fails without the Close-path teardown, passes with it (30× -race).** Graceful-path tests still see exactly one teardown (no double-teardown from the `Once`). Removed the now-unused mock helper.

## Adversarial self-review
- **Stop/Close race** → `sync.Once` serializes; exactly one teardown either ordering.
- **Graceful Read-interruption preserved** — `Stop` still tears down first.
- **Error propagation** — first caller into `Once.Do` gets the teardown error; no double-report.
- No data-integrity invariant touched (this is resource cleanup, not record/ack/position handling).

Tier 2 (preview-arch resource cleanup; not the data path). Full poc suite green under `-race -shuffle`.

### DIFF
--- a/pkg/lifecycle-poc/funnel/source.go
+++ b/pkg/lifecycle-poc/funnel/source.go
@@ -77,7 +77,8 @@ func (t *SourceTask) Open(ctx context.Context) error {
 }
 
 func (t *SourceTask) Close(context.Context) error {
-	// source is torn down in the worker on stop
+	// The source is torn down by the worker (Worker.tearDownSource, called from
+	// both Stop and Close), not here.
 	return nil
 }
 
--- a/pkg/lifecycle-poc/funnel/worker.go
+++ b/pkg/lifecycle-poc/funnel/worker.go
@@ -90,24 +90,34 @@ type Worker struct {
 	processingLock chan struct{}
 	// stop stores the information if a graceful stop was triggered.
 	stop atomic.Bool
-	// teardownOnce guarantees the source is torn down exactly once, no matter
-	// which shutdown path (graceful Stop or Close) reaches it first.
-	teardownOnce sync.Once
+	// teardownMu guards sourceTornDown and serializes teardown so the source is
+	// torn down at most once across the Stop/Close race.
+	teardownMu sync.Mutex
+	// sourceTornDown is set only after Source.Teardown SUCCEEDS, so a failed
+	// teardown is retried by a later call rather than masked as done.
+	sourceTornDown bool
 
 	logger log.CtxLogger
 }
 
-// tearDownSource tears the source down exactly once. Teardown serves two
+// tearDownSource releases the source connector's resources. It serves two
 // purposes: on a graceful stop it interrupts a blocked source Read, and on every
-// path it releases the source connector's resources. It is called from both Stop
-// (graceful) and Close (all paths, including a fatal error where Stop is never
-// called), so it must be idempotent.
+// path it frees the source. It is called from both Stop (graceful) and Close (all
+// paths, including a fatal error where Stop is never called), so it is idempotent
+// — but only a *successful* teardown is remembered: if Source.Teardown fails, the
+// source is left un-torn-down so the next caller retries, rather than masking the
+// failure and leaking the source.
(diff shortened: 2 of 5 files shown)


devarismeroxa:
**Fresh-context review round (architect-reviewer): APPROVE-WITH-NITS → nit fixed.**

The review verified the leak/root-cause against the code, confirmed no deadlock (Close doesn't take the processing lock), correct context usage (Close uses `Background()` so teardown isn't skipped on an already-canceled tomb ctx), preserved graceful Read-interruption, and independently reproduced the fails-without-fix regression. It also confirmed the force-stop path is now correctly torn down (filed #2563 for its pre-existing lack of test coverage).

It found one real gap my self-review missed: **`sync.Once` masks a *failed* `Teardown` as success** on the next touch, silently reintroducing the leak (gated on Teardown failing once). Fixed in b7946a1 — replaced `sync.Once` with a mutex + `sourceTornDown` flag set only after Teardown *succeeds*, so a failed teardown is retried. Added two `funnel` unit tests (retry-after-failure — verified to fail against the mask-on-failure variant — and once-on-success).

PR #2572, opened 2026-07-08
feat(cli): shared foundations for v0.17 CLI-as-product commands

devarismeroxa:
## Summary

Ships the shared foundations that `pipelines validate`, `doctor`, and
`connector|processor new` (v0.17 "CLI as product") will all build on, built
first so the three commands can't drift onto private output formats — per
[`docs/design-documents/20260707-cli-output-conventions.md`](../blob/docs/v0.17-first-wave-design/docs/design-documents/20260707-cli-output-conventions.md)
and
[`docs/architecture-decision-records/20260707-check-engine-shared-by-doctor-and-scaffold.md`](../blob/docs/v0.17-first-wave-design/docs/architecture-decision-records/20260707-check-engine-shared-by-doctor-and-scaffold.md)
(both on branch `docs/v0.17-first-wave-design`, #2571 — not yet merged; this
PR implements to that spec but doesn't depend on merge order).

**No `validate`/`doctor`/`connector new` commands land in this PR.** Foundations
only, so the three commands can build against a stable, already-reviewed
surface instead of each inventing their own envelope/exit-code/glyph logic.

### 1. `pkg/conduit/check` — the neutral check engine (per the ADR)

No cobra/ecdysis imports, no `os.Exit`.

- `Status` (`pass|warn|fail`), `CheckResult{Name, Status, Message, Suggestion, Code, ConfigPath, Category}` (field order matches the ADR), `Check` interface, `Report{Checks, Summary}`.
- `Run(ctx, checks []Check, opts ...Option) Report` — per-check `recover()` turns a panic into a `Fail` (`code: internal.error`) instead of crashing the process; per-check context timeout (a `TimeoutCheck` override, else the batch default via `WithTimeout`, else `DefaultTimeout`).
- `Report.ExitCode() int` — aggregates every `Fail` result to ONE exit code: synthesizes a `*conduiterr.ConduitError` per failing result via `conduiterr.LookupCode`, classifies each independently with `exitcode.ExitCode`, and returns the **MAX** bucket (env 3 > validation 2 > runtime 1). This is new aggregation logic, not `cerrors.Join` + classify-once (which would silently prefer whichever failure happened to be joined first). Also floors at `Runtime` instead of `OK` if a `Fail`'s `Code` ever resolves to the `Canceled`/`OK` gRPC category — a `Fail` can never silently report success, even from a misbehaving caller.
- Generic, reusable check constructors doctor and scaffold preflight will both compose: `BinaryOnPath(name, minVersion)`, `DirWritable(path, configKey)`, `AddrBindable(addr, configKey)`.

### 2. `cmd/conduit/internal/ui` — shared human-rendering helper (conventions §2)

- `Renderer`: owns TTY/`--no-color`/`NO_COLOR` detection + glyph/color selection (Unicode+ANSI on a color TTY, ASCII `[OK]`/`[WARN]`/`[FAIL]` otherwise).
- `RenderFinding(w, glyph, code, configPath, message, suggestion)`: the canonical located-finding layout, byte-verified against the conventions doc's own example.

### 3. `cmd/conduit/cecdysis` — offline result decorator + shared `--json` envelope

- `CommandWithResult` / `CommandWithResultDecorator`: the offline sibling of `CommandWithExecuteWithClientResultDecorator` (no API client dial), for commands like `doctor` and `pipelines validate` whose own checks are the thing being run.
- Marshals the shared `Result{command, ok, summary, result, error}` envelope, lowerCamelCase, matching existing protojson output.
- Sets `SilenceErrors`/`SilenceUsage` structurally (every command using this decorator gets it, not opt-in) so cobra never double-prints — including for an error from an *earlier* decorator in the chain, not just from `ExecuteWithResult` itself (see self-review below).
- Registered in `cmd/conduit/cli/cli.go` alongside the existing client decorators.

## Adversarial self-review

Found two real gaps before opening this PR, both now covered by regression tests:

1. **`Report.ExitCode()` could return `OK` despite a `Fail` being present.** If a `Fail` result's `Code` happened to resolve to the `Canceled`/`OK` gRPC category (something no well-behaved registered `conduiterr.Code` should do, but the method shouldn't trust callers on), the aggregation would silently report success. Fixed to floor at `Runtime` whenever any `Fail` is present. See `TestReport_ExitCode_FailNeverReportsOK`.
2. **`CommandWithResultDecorator` only rendered `ExecuteWithResult`'s own error.** An error from an *earlier* decorator in the chain (e.g. a future confirm/prompt step on a scaffold command, or `ecdysis.CommandWithExecuteDecorator` if a command implements both) would exit nonzero with **no output at all**, under both `--json` and human mode, since `SilenceErrors` suppresses cobra's own printing and nothing else was rendering it. Fixed to render that path identically to a hard `ExecuteWithResult` failure. See `TestCommandWithResultDecorator_PriorDecoratorFailure_StillRendered` and its `--json` counterpart.

The MAX-not-first aggregation is tested with both orderings of a
`Fail@Unavailable` (env, 3) + `Fail@InvalidArgument` (validation, 2) pair, and
with all six permutations of a three-bucket `Fail` set (runtime/validation/env)
— always exits 3/the max, never order-dependent.

## Exported surface for the next agents (validate/doctor/scaffold)

```go
package check // pkg/conduit/check

type Status string
const (StatusPass Status = "pass"; StatusWarn Status = "warn"; StatusFail Status = "fail")

const (CategoryToolchain = "toolchain"; CategoryFilesystem = "filesystem"; CategoryNetwork = "network")

type CheckResult struct {
    Name, Message, Suggestion, Code, ConfigPath, Category string
    Status Status
}

type Check interface {
    Name() string
    Run(ctx context.Context) CheckResult
}
type TimeoutCheck interface { Check; Timeout() time.Duration }

type Summary struct{ Total, Passed, Warned, Failed int }
type Report struct {
    Checks  []CheckResult
    Summary Summary
}
func (r Report) ExitCode() int

const DefaultTimeout = 10 * time.Second
type Option func(*runConfig)
func WithTimeout(d time.Duration) Option
func Run(ctx context.Context, checks []Check, opts ...Option) Report

func BinaryOnPath(name, minVersion string) Check
func DirWritable(path, configKey string) Check
func AddrBindable(addr, configKey string) Check
```

```go
package cecdysis // cmd/conduit/cecdysis

type Result struct {
    Command string
    OK      bool
    Summary any
    Result  any
    Error   *ResultError
}
type ResultError struct { Code, Message, Suggestion, ConfigPath string; Fix any }

type Outcome struct { OK bool; Summary, Result any }
type CommandWithResult interface {
    ecdysis.Command
    ResultCommand() string
    ExecuteWithResult(ctx context.Context) (Outcome, error)
    Render(outcome Outcome) string
}
type CommandWithResultDecorator struct{}
```

```go
package ui // cmd/conduit/internal/ui

type Status int
const (StatusPass Status = iota; StatusWarn; StatusFail)
func NewRenderer(w io.Writer, noColor bool) *Renderer
func (r *Renderer) Glyph(status Status) string
func (r *Renderer) Color() bool
func (r *Renderer) Glyphs() bool
func RenderFinding(w io.Writer, glyph, code, configPath, message, suggestion string)
```

## Verification

- `go build ./...` — clean
- `go test ./pkg/conduit/check/... ./cmd/conduit/internal/ui/... ./cmd/conduit/cecdysis/... -race -count=1` — green
- `go test ./... -race -count=1` — green (full repo, unaffected)
- `golangci-lint run` — 0 issues, both scoped to the new packages and repo-wide
- `go mod tidy` — promotes `github.com/mattn/go-isatty` from indirect to direct (already vendored; used for TTY detection in `ui.Renderer`). No other dependency changes.

## Test plan

- [x] `go build ./...`
- [x] `go test ./pkg/conduit/check/... ./cmd/conduit/internal/ui/... ./cmd/conduit/cecdysis/... -race -count=1`
- [x] `go test ./... -race -count=1`
- [x] `golangci-lint run` (repo-wide)
- [ ] Human review of the `Report.ExitCode()` aggregation semantics and the `CommandWithResult` envelope shape, since this is the surface `validate`/`doctor`/scaffold will all build against

Tier 2. No breaking changes (new packages + one additive decorator registration).

https://claude.ai/code/session_01Tapg9dLWXKMZJoL65R24vd

### DIFF
--- a/cmd/conduit/cecdysis/decorators.go
+++ b/cmd/conduit/cecdysis/decorators.go
@@ -16,6 +16,7 @@ package cecdysis
 
 import (
 	"context"
+	"fmt"
 	"path/filepath"
 
 	"github.com/conduitio/conduit/cmd/conduit/api"
@@ -136,11 +137,18 @@ func (CommandWithExecuteWithClientResultDecorator) Decorate(_ *ecdysis.Ecdysis, 
 			if err != nil {
 				return cerrors.Errorf("could not marshal result to JSON: %w", err)
 			}
-			cmd.Println(string(b))
+			// cmd.Println/cmd.Print resolve through cobra's OutOrStderr, not
+			// OutOrStdout (see cobra's own doc on Print: "fallback to
+			// Stderr if not set") — and no production call site ever calls
+			// cmd.SetOut, so using them here put --json output on stderr
+			// instead of stdout. Write to OutOrStdout explicitly instead;
+			// see cecdysis.CommandWithResultDecorator for the same fix
+			// applied to this decorator's offline sibling.
+			fmt.Fprintln(cmd.OutOrStdout(), string(b))
 			return nil
 		}
 
-		cmd.Print(v.Render(result))
+		fmt.Fprint(cmd.OutOrStdout(), v.Render(result))
 		return nil
 	}
 
--- a/cmd/conduit/cecdysis/result.go
+++ b/cmd/conduit/cecdysis/result.go
@@ -16,6 +16,7 @@ package cecdysis
 
 import (
 	"context"
+	"fmt"
 
 	"github.com/conduitio/conduit/cmd/conduit/internal/ui"
 	"github.com/conduitio/conduit/pkg/foundation/cerrors"
@@ -159,11 +160,17 @@ func (CommandWithResultDecorator) Decorate(_ *ecdysis.Ecdysis, cmd *cobra.Comman
 			if merr != nil {
 				return cerrors.Errorf("could not marshal result to JSON: %w", merr)
 			}
-			cmd.Println(string(b))
+			// cmd.Println/cmd.Print resolve through cobra's OutOrStderr, not
+			// OutOrStdout (cobra's own doc: "fallback to Stderr if not
+			// set") — since no production call site ever calls cmd.SetOut,
+			// using them here would silently put the "single JSON object to
+			// stdout" envelope on stderr instead. Write to OutOrStdout
+			// explicitly everywhere in this file instead.
+			fmt.Fprintln(cmd.OutOrStdout(), string(b))
 			return nil
 		}
 
-		cmd.Print(v.Render(outcome))
+		fmt.Fprint(cmd.OutOrStdout(), v.Render(outcome))
 		return nil
 	}
 
@@ -185,11 +192,16 @@ func renderResultError(cmd *cobra.Command, command string, err error, asJSON boo
 		b, merr := marshalJSON(res)
 		if merr != nil {
 			// Should not happen (see doc above); fall back to something
-			// rather than emitting nothing on stdout under --json.
+			// rather than emitting nothing at all under --json. This one
+			// case intentionally goes to stderr (via the correctly-behaved
+			// PrintErrln, which does resolve through ErrOrStderr) since
+			// there is no well-formed JSON object to put on stdout here.
 			cmd.PrintErrln("Error:", err)
 			return
 		}
-		cmd.Println(string(b))
+		// See the OutOrStdout note in Decorate: cmd.Println would put this
+		// on stderr instead of stdout in production.
+		fmt.Fprintln(cmd.OutOrStdout(), string(b))
 		return
 	}
 
--- a/cmd/conduit/cecdysis/result.go
+++ b/cmd/conduit/cecdysis/result.go
@@ -1,0 +1,221 @@
+// Copyright © 2026 Meroxa, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package cecdysis
+
+import (
+	"context"
+
+	"github.com/conduitio/conduit/cmd/conduit/internal/ui"
+	"github.com/conduitio/conduit/pkg/foundation/cerrors"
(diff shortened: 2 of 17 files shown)


devarismeroxa:
## Second-pass adversarial review (fresh session, independent of the author)

Per the repo's AI-authored-code process, ran a second review pass in a fresh session with no context from the authoring session, specifically hunting for bugs in the `Report.ExitCode()` MAX-not-first aggregation and the `CommandWithResultDecorator` output/error paths. It found three real issues, all fixed in the follow-up commit with regression tests that fail without the fix and pass with it:

1. **Panic-isolation gap**: `runOne` called `c.Name()` *before* registering the deferred `recover()`, so a panicking `Name()` (not just a panicking `Run()`) escaped uncaught and crashed the whole `Run()` call — the exact failure mode the isolation exists to prevent, and worse than a panicking `Run` because it aborted every subsequent check too. Fixed; see `TestRun_PanicIsolated_FromName`.
2. **`--json` output landing on stderr in production**: `CommandWithResultDecorator` (and the pre-existing sibling `CommandWithExecuteWithClientResultDecorator` it mirrors) used `cmd.Println`/`cmd.Print`, which resolve through cobra's `OutOrStderr`, not `OutOrStdout` ("fallback to Stderr if not set," per cobra's own doc) — and no production call site ever calls `cmd.SetOut`. Every existing test masked this by calling `cmd.SetOut(&buf)`. Fixed both decorators to write via `fmt.Fprint(ln)(cmd.OutOrStdout(), ...)` explicitly, and added a regression test that swaps the real process-level `os.Stdout`/`os.Stderr` instead of using `SetOut`, so this class of bug is actually caught going forward: `TestCommandWithResultDecorator_JSON_ActuallyGoesToRealStdout`.
3. **Test-hygiene nit**: a test called `conduiterr.Register()` inside a function body instead of a package-level var initializer, per that function's own documented invariant. Moved to match the existing convention in `pkg/conduit/exitcode`'s test suite.

The `ExitCode()` MAX-not-first aggregation itself was independently re-verified and confirmed sound (both orderings of a 2-bucket case, all 6 permutations of a 3-bucket case).

Full diff between commits `e9686ba` (original) and `425c193` (post-review fixes) is 6 files, +152/-25. `go build ./...`, `go test ./... -race -count=1`, and `golangci-lint run ./...` all green after the fixes.

PR #2592, opened 2026-07-08
feat(cli): add conduit pipeline lint and dry-run

devarismeroxa:
## Summary

Implements `conduit pipeline lint` and `conduit pipeline dry-run` per
[docs/design-documents/20260708-cli-pipeline-inspect-lint-dryrun.md](docs/design-documents/20260708-cli-pipeline-inspect-lint-dryrun.md)
(the `inspect` verb from that doc is out of scope for this PR — it's the
online verb, this PR is the two offline ones). Builds on the merged
`cmd/conduit/internal/validate` engine from #2574: one engine, now three
verbs.

- **`lint <path>`** (offline): validate + advisory parser warnings
  (deprecated/renamed/unknown fields, version fallback), each with
  line/column. Warnings exit 0 unless `--strict`, which promotes them to
  exit 2. An error always exits 2, `--strict` or not, and both render
  together.
- **`dry-run <path>`** (offline): validate + the **enriched** graph (final
  `pipelineID:connectorID` IDs, injected DLQ defaults, worker counts) +
  `--resolve-plugins` (default on) resolving `builtin:`-prefixed refs
  against the compiled-in builtin registry. Unknown builtin →
  `connector.plugin_not_found`/`processor.plugin_not_found`, exit 2.
  `standalone:`/unprefixed refs are not statically verifiable (an
  unprefixed ref tries standalone before builtin at runtime — see
  `pkg/plugin/connector/service.go`'s `newDispenser`) and are never a false
  failure — marked `(unverified)` instead.
- Both registered under `PipelinesCommand`; `validate`/`lint`/`dry-run`
  `--help` cross-reference each other per the design doc's review-outcome
  UX fix.
- README's pipeline-config section rewritten to cover all three verbs.

### The warnings-exposure refactor (`pkg/provisioning/config/yaml`)

`lint`/`dry-run` need the parser's collected-then-discarded warnings.
Exported the previously-unexported `warning`/`warnings` type as
`Warning`/`Warnings`, and added `Parser.ParseWithWarnings` as a **new,
separate entry point** that returns warnings instead of only logging them.
`Parser.Parse`/`ParseConfigurations` — `conduit run`'s path, via
`pkg/provisioning.Service.parsePipelineConfigFile` — are untouched: same
signature, same behavior, still only logs. This is proven, not asserted:
`TestParser_WarningsExposure_DoesNotChangeRunBehavior` runs both paths over
the same fixture with two independent loggers and asserts the run path's
logged JSON output is byte-for-byte identical to the pre-refactor baseline
(`TestParser_V1_Warnings`), while the new path returns the same 6 warnings
instead of logging anything itself.

## Adversarial self-review

Re-read the diff hunting for bugs before opening this PR (see individual
commits for the two rounds):

1. **Caught and fixed**: `DryRunCommand.Render` branched purely on
   `FileReport.OK` (errors-only) to decide whether to print a file's
   findings. Since `dry-run` always includes parser warnings (like `lint`),
   a warning-only file was "OK" at that level, so its warning silently
   never appeared in human output (while still present in `--json` — an
   OK/`--json` asymmetry, not a data-loss bug). Fixed by giving `dry-run`
   the same three-way pass/warn/fail classification `lint` already had
   (shared `fileLintStatus`/`findingLocation` helpers). Regression test
   (`TestDryRunCommand_WarningOnlyFile_HumanShowsWarning`) verified failing
   against the pre-fix `Render` and passing after — see the second commit.
2. **Verified, not just asserted**: the run-unchanged guarantee (top risk
   per the task). Ran the full `pkg/provisioning` suite (exercises the real
   `conduit run` path) plus the dedicated regression test above — both
   green.
3. **Checked**: no data-path/invariant risk — this is CLI/config-parsing
   only, no record flow, no ack/position/checkpoint logic. Tier 2, not
   Tier 1.
4. **Checked**: resource cleanup — `validateFile`'s `defer f.Close()` is
   unchanged; no new goroutines, no concurrency.
5. **Considered and scoped out**: DLQ plugin refs are not resolved by
   `--resolve-plugins` (only connector/processor refs) — matches the design
   doc's explicit scope; noted here rather than silently expanding scope.
6. One unrelated, pre-existing test flake observed in the full-repo run
   (`pkg/plugin/processor/standalone`'s WASM specification tests,
   `context deadline exceeded` under `-race` + full parallelism) —
   confirmed passing in isolation, confirmed unrelated (package untouched
   by this PR).

## Test plan

- [x] `go build ./...`
- [x] `go test ./cmd/conduit/... ./pkg/provisioning/... -race -count=1` — green
- [x] `golangci-lint run ./cmd/conduit/... ./pkg/provisioning/...` — 0 issues
- [x] AC-1/2/3 (lint): warning w/ line+column exit 0; `--strict` exit 2;
      error+warning both rendered, error dominates, exit 2
- [x] AC-4 (run-unchanged): dedicated regression test, described above
- [x] AC-5/6 (dry-run): enriched IDs/DLQ defaults shown; unknown builtin
      exit 2; standalone advisory exit 0 (not a false fail)
- [x] AC-7 (offline): engine-level `TestRun{Lint,DryRun}_Offline_NoDial` +
      CLI-level timing checks against an unroutable address
- [x] Manually exercised every case end-to-end via a built binary (human +
      `--json` output, `--help` cross-refs) before writing the automated
      tests

https://claude.ai/code/session_01Tapg9dLWXKMZJoL65R24vd

### DIFF
--- a/README.md
+++ b/README.md
@@ -330,10 +330,13 @@ share.
 
 ## Validating pipeline configs
 
-`conduit pipelines validate <path>` (alias: `conduit pipeline validate`) statically checks one
-pipeline config file, or every `.yml`/`.yaml` file in a directory (not recursed), without starting
-Conduit or contacting a running instance — it's fully offline, so it's safe to run in CI or before
-`conduit run` with no server anywhere nearby.
+`conduit pipelines validate|lint|dry-run <path>` (alias: `conduit pipeline ...`) share one offline
+engine that statically checks one pipeline config file, or every `.yml`/`.yaml` file in a directory
+(not recursed), without starting Conduit or contacting a running instance — all three are fully
+offline, so they're safe to run in CI or before `conduit run` with no server anywhere nearby.
+
+**`validate`** — parse, enrich, and check every field. Every problem in every file is reported — a
+bad file never stops the rest from being checked.
 
 ```console
 $ conduit pipelines validate pipelines/orders.yaml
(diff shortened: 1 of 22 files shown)


devarismeroxa:
Closing as a duplicate. `conduit pipelines lint` and `dry-run` already landed on main via #2584 and #2585 (implemented inline while this branch's agent was recovering from an API stall), so this PR conflicts with the merged engine and can't land as-is.

The effort isn't lost: this implementation resolved **processor** plugin refs in dry-run (the merged version only resolves connector plugins) and had a stronger byte-for-byte AC-4 (run-unchanged) test. Both are being carried forward as a focused enhancement PR onto the merged engine. Thanks for the thorough work — and apologies for the worktree removal mid-run that forced the redo.

PR #2601, opened 2026-07-13
docs: design — live in-place hot-reload (processor + metadata) + conduit run --dev (§4)

devarismeroxa:
## What

Design doc for v0.17 execution-plan **§4 (hot-reload)** + **§3 (dev verb)**. Roadmap: Phase 1 (Developer Experience Core). **Design only — no implementation yet.** Opening for **Tier-1 design sign-off** before code, mirroring how #2588's design landed as #2596 first.

## The scope decision (made in review)

In-place-class changes must apply **without restarting** the pipeline — the real §4, not the always-restart shortcut (rejected). Invariant-safety line:

- **Live-swappable in place:** processor config + pipeline Name/Description — no position/ack/connection state, so a swap at a clean record boundary can't skip a position, lose an ack, or drop a record.
- **Restart-class (existing `ApplyPlanLive`):** source/dest connector settings, DLQ (live destination w/ ack semantics), all topology. Live swap of these is invariant-1/2/3 territory → deferred to Phase 2 with the chaos harness.

## Core new work

A live, at-a-record-boundary **`ProcessorNode` swap** in `pkg/lifecycle/stream`, built on the existing `ControlMessage`/`InjectControlMessage` idiom (already used to stop sources in-band). Open-new-before-teardown-old so a bad edit never drops the pipeline. `ApplyPlanLiveInPlace` decides: all-swappable diff → in place; anything restart-class → delegate to `ApplyPlanLive`. Dev-mode (`run --dev` + `pipelines dev` alias) is a consumer.

## Grounding + rigor

Every mechanism claim verified against the tree: `ApplyPlanLive`/`StopAndWait`, `ProcessorNode.Run`, `nodeBase.Receive`, `connectorService.Update` (store-only), `buildNodes`, and #2236 (superseded). Includes:
- Failure-mode table (bad-Open keeps old, idle-pipeline swap, mid-swap crash, ensure-running recovery, atomic-save, file-delete).
- Invariant (1–7) analysis of the live swap.
- **Honest new-crash-surface accounting:** store commit is the single source of truth; node swaps are in-memory/non-durable → crash always resumes consistent. Gated by a **targeted mid-swap fault test** (full chaos suite remains a Phase 2 gate, not claimed).

## Adversarial self-review (done inline; the async review agents stalled on infra)

[...]

PR #2613, opened 2026-07-13
feat: live in-place hot-reload engine (processor swap + apply decision) — PR1 of §4 (Tier 1)

devarismeroxa:
## What

PR1 (engine) of the live in-place hot-reload arc (§4). Makes a processor-config or pipeline name/description change apply to a **running** pipeline in place — no stop/restart, no availability blip, no position replay, no record loss. Design: `docs/design-documents/20260712-pipeline-dev-hot-reload.md` (merged #2601). PR2 (the `conduit run --dev` surface) follows.

**Risk tier: 1 (data path).** Requesting DeVaris sign-off + failure-mode analysis (below) before merge.

## The five increments
1. **Swap primitive** (`pkg/lifecycle/stream`) — `ProcessorNode.Reconfigure`: swaps a processor at a record boundary, applied by the run-loop goroutine itself (so `n.Processor` is never mutated concurrently with `Process`), open-before-teardown, idle-interruptible.
2. **Classification** (`pkg/provisioning`) — `Change.LiveSwappable` + `Diff.LiveEligible()`: processor + name/description → swappable; connector/DLQ/topology → restart.
3. **Lifecycle wiring** — `ReconfigureProcessor` reaches the live node; parallel/absent → `ErrProcessorNotLiveReconfigurable` → restart fallback; poc arch refuses in parity.
4. **Apply decision** — `ApplyPlanLive` applies live-eligible diffs in place; parallel processor → restart fallback; open-failure → rollback + error, never restart.
5. **Tests** — swap boundary, open-fail-keeps-old, idle, ctx-cancel; classification (every action type); apply in-place/fallback/rollback; **batching contract test** (1-in-1-out tripwire); **property test** (200 records × many swaps: no drop/reorder); **mid-swap fault test**.

## Failure-mode analysis (Tier 1)

- **Could it drop/reorder records?** The swap happens at a record boundary in the run-loop goroutine (no record mid-`Process`); backpressure pauses the source during the swap; open-before-teardown keeps the old processor on a bad edit. Proven by the property test (exactly-once, in-order across many swaps), the contract test (1-in-1-out), and the fault test. **Invariants 1/3/4 hold.**
- **Could a failed apply diverge store vs live?** On a post-commit swap failure, `rollbackInPlace` restores the old config and re-swaps back; on a parallel-processor fallback, the restart rebuilds from the committed config. The pipeline keeps running its old processors throughout (open-before-teardown).
- **Crash mid-swap?** The store commit is the single source of truth; node swaps are in-memory/non-durable, so restart resumes from the committed config (invariant 5). The mid-swap fault test covers ctx-cancellation; the full SIGKILL chaos test is **honestly deferred to Phase 2** per the process-maturity table (design doc states this).
- **What would show a regression?** A record-throughput drop or DLQ spike on a pipeline that took an in-place apply; pipeline status flapping.
- **Rollback:** additive to `ApplyPlanLive`; the operator gate (`allowRestartOnRunning`) defaults off, so the in-place path is only reachable when explicitly authorized. Reverting the PR returns `ApplyPlanLive` to pure drain-and-restart (the shipped behavior). A stuck pipeline can always be stopped + started (full rebuild from stored config).

## Decisions to confirm at sign-off
- **Operator gate (Option A, conservative):** an in-place change to a running pipeline still requires `allowRestartOnRunning`. The in-place path is the non-disruptive *execution* of an authorized apply, not a gate bypass. A future refinement could relax it for provably-in-place diffs.
- **Rollback reverse-import** is exercised by the fault path but not fully asserted by the stateless provisioning mock (noted in the test) — worth a chaos/integration follow-up in Phase 2.

## Adversarial self-review
Re-read for: the swap boundary race (safe — single-goroutine application), open-before-teardown resource cleanup (failed new processor torn down), the ErrProcessorNotLiveReconfigurable fallback (restart is a safe superset), and the rollback path (best-effort, pipeline keeps running). All existing `ApplyPlanLive` restart tests still pass (connector/DLQ/topology diffs aren't live-eligible → restart path unchanged).

### DIFF
--- a/pkg/conduit/runtime.go
+++ b/pkg/conduit/runtime.go
@@ -120,6 +120,10 @@ type lifecycleService interface {
 	// lifecycle-poc(pkg/lifecycle-poc).Service.StopAndWait for why the
 	// Preview.PipelineArchV2 implementation always refuses.
 	StopAndWait(ctx context.Context, pipelineID string) error
+	// ReconfigureProcessor keeps this interface a superset of
+	// provisioning.LifecycleService (see StopAndWait above) — it is used by the
+	// live in-place apply path. See lifecycle.Service.ReconfigureProcessor.
+	ReconfigureProcessor(ctx context.Context, pipelineID, processorID string) error
 	Init(ctx context.Context) error
 }
 
--- a/pkg/lifecycle-poc/service.go
+++ b/pkg/lifecycle-poc/service.go
@@ -353,6 +353,18 @@ func (s *Service) StopAndWait(context.Context, string) error {
 	return ce
 }
 
+// ReconfigureProcessor refuses under the experimental Preview.PipelineArchV2
+// lifecycle service, mirroring StopAndWait: live in-place apply is not supported
+// under this arch. Returning the same coded error means a live apply is cleanly
+// refused rather than silently downgraded.
+func (s *Service) ReconfigureProcessor(context.Context, string, string) error {
+	ce := conduiterr.New(CodeStopAndWaitUnsupported,
+		"live in-place apply (processor reconfigure on a running pipeline) is not supported under "+
+			"the experimental Preview.PipelineArchV2 lifecycle service")
+	ce.Suggestion = "disable Preview.PipelineArchV2, or stop the pipeline manually before applying changes"
+	return ce
+}
+
 // buildRunnablePipeline will build and connect all tasks configured in the pipeline.
 func (s *Service) buildRunnablePipeline(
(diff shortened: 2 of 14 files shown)


devarismeroxa:
Tier-1 sign-off from DeVaris (maintainer). Independent fresh-context review complete; its MAJOR-1 (Workers misclassification) blocker and MINOR-1/2/3 are fixed in 731300a. All CI green. Merging.

PR #2616, opened 2026-07-13
feat(cli): conduit run --dev hot-reload watcher (PR2 of §4)

devarismeroxa:
## What
`conduit run --dev` — the developer surface over the live in-place hot-reload **engine** (PR1, #2613). Watches the pipelines directory and applies config changes to running pipelines on save: a **processor / name / description** edit applies **in place** (no restart, positions continuous); a **connector / DLQ / topology** edit **drain-restarts**, each labeled. Completes execution-plan **§4** (+ §3). Design: `docs/design-documents/20260712-pipeline-dev-hot-reload.md` (merged #2601). Risk tier 2 — a watcher over the already-Tier-1-reviewed engine.

## Key points
- **Watcher lives in the runtime** (`pkg/conduit/dev`), started after `ProvisionService.Init`, tied to the serve context (Ctrl-C cancels it — invariant 7). `--dev` sets `Config.Dev`.
- **`--dev` is the operator authorization** for the applies its watcher drives (interactive, local, watched) — it does **not** enable the process-level API gate; remote agents stay independently gated.
- **Ensure-running:** a new file, or a pipeline left stopped by a prior failed apply, is started (unless the config declares it stopped).
- **Failure modes:** parse/validate error keeps the running pipeline untouched; deleted watched file leaves the pipeline running + logs (never auto-deletes); atomic-save (rename) tolerated by watching the directory.
- `--json` events; `conduit pipelines dev [dir]` thin alias. `fsnotify` promoted to a direct dep (already transitively present).

## Tests
Unit (event→file→pipeline mapping, debounce/coalesce, parse-error tolerance, ensure-running variants, file-delete keep-running) **plus an integration test** with real fsnotify + a generator→processor→file pipeline: edit the processor → applied in place; edit a source → labeled restart; syntax error → pipeline keeps running. Build/test (`-race`)/lint green; operator runbook added; no llms drift.

## Process note
Authored by a subagent; finalized (runbook + full verification) by the main session after it stalled at the finish line. **Pending independent review before merge.**

### DIFF
--- a/cmd/conduit/root/pipelines/apply.go
+++ b/cmd/conduit/root/pipelines/apply.go
@@ -111,7 +111,7 @@ func (c *ApplyCommand) Args(args []string) error {
 func (c *ApplyCommand) Config() ecdysis.Config {
 	path := filepath.Dir(c.flags.ConduitCfg.Path)
 	return ecdysis.Config{
-		EnvPrefix:     "CONDUIT",
+		EnvPrefix:     envPrefix,
 		Parsed:        &c.flags.Config,
 		Path:          c.flags.ConduitCfg.Path,
 		DefaultValues: conduit.DefaultConfigWithBasePath(path),
--- a/cmd/conduit/root/pipelines/deploy.go
+++ b/cmd/conduit/root/pipelines/deploy.go
@@ -124,7 +124,7 @@ func (c *DeployCommand) Args(args []string) error {
 func (c *DeployCommand) Config() ecdysis.Config {
 	path := filepath.Dir(c.flags.ConduitCfg.Path)
 	return ecdysis.Config{
-		EnvPrefix:     "CONDUIT",
+		EnvPrefix:     envPrefix,
 		Parsed:        &c.flags.Config,
 		Path:          c.flags.ConduitCfg.Path,
 		DefaultValues: conduit.DefaultConfigWithBasePath(path),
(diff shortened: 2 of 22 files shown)


devarismeroxa:
## Independent review + hardening

Fresh-context review of this PR (rebased on the merged engine PR #2619) confirmed **both prior blockers are genuinely fixed** and found no merge-blocker:
- **In-place works** — the integration test asserts a real `mode:in_place` with zero restart against the real engine + real fsnotify, no failure-tolerant branch.
- **Shutdown race fixed** — `debouncer.run` waits out the in-flight apply on cancel (buffered `done`, balanced receive, no deadlock); `consume`'s `wg` guarantees no apply outlives `Watcher.Run`. Verified race-free.
- Confirmed `--dev` does **not** set the process-level `--api.allow-live-restart-apply` gate; `modeFromApplied` fallback semantics correct against the real `ApplyMode` zero value.

Addressed the review's hardening findings in `ff74831`:
- **MEDIUM — panic containment:** `applyFile` runs rapidly-edited user YAML on a plain goroutine; an unrecovered panic would crash the server and take down every live pipeline. It now recovers and reports the panic as an error event. Regression test `TestApplyFile_RecoversFromPanic`.
- **LOW/MEDIUM — shutdown teardown:** `consume` now derives a child context cancelled before `wg.Wait`, so per-path debouncers are torn down no matter why it returns (parent-ctx cancel OR channel close); a closed `fsErrors` returns instead of busy-spinning. Regression test `TestConsume_ChannelClose_TearsDownAndReturns` (hangs without the fix).
- **LOW — shutdown hygiene:** the debounce timer no longer starts a fresh apply once ctx is cancelled.

Deferred (non-blocking, noted for follow-up): unbounded debouncer growth per distinct filename over a very long session; `--dev.json` naming (scoped streaming surface) vs the repo-standard `--json`; `renderHuman` not printing `ConfigPath`.

PR #2656, opened 2026-07-22
feat(registry): embed production trust anchors (Gate 3, Tier-1)

devarismeroxa:
## What

Populates the connector registry's compiled-in trust anchors — the final code change that makes `conduit connectors install` verify real indexes instead of failing closed. Gate 3 of the bootstrap ceremony.

- `anchors.go` — `go:embed`s `trustanchors/` and, at `init`, parses the root + freshness SPKI ed25519 PEMs into `index.TrustAnchors` (keyed by `index.KeyID`; multiple concatenated blocks = a rotation window). Populates `defaultTrustAnchors`, used by install/audit/bundle.
- `trustanchors/{root,freshness}.pub.pem` — the **public** keys minted by the `registry-signing` Environment's `root-keygen` run (29950357295). Private keys never left that Environment.
- keyIds pinned in `TestEmbeddedTrustAnchorsParse`:
  - root `sha256:d657c27…009795`, freshness `sha256:50bfd2c…9cf783`

## Risk tier: Tier-1 (trust/data-path) — needs sign-off

This flips every real `conduit connectors install` from fail-closed to trust-these-keys. Before this PR: `defaultTrustAnchors` empty → every real index refuses with `registry.trust_anchor_expired`. After: indexes signed by these keys verify.

## Design decisions for your sign-off

1. **Degrade to fail-closed, never panic.** If the embedded PEMs are absent or unparseable, `init` leaves `defaultTrustAnchors` empty → `trust_anchor_expired`. It never panics (so `conduit run` and unrelated commands are unaffected) and never fails open.
2. **No distinct "corrupt build" CLI error.** A missing/corrupt embed and a genuinely-expired anchor both surface `trust_anchor_expired`. I chose not to distinguish them in the CLI: both fail closed, the distinction adds control-flow risk in the trust path, and `TestEmbeddedTrustAnchorsParse` already blocks shipping a broken/empty embed in CI. Happy to add the louder operator message if you'd prefer it.
3. **keyIds pinned in a test.** A swap of the embedded keys fails CI — a key change is deliberate (rotation) and must update the pinned constants in the same reviewed PR, never a silent substitution.

## Failure-mode analysis

- *Embedded key corrupted/truncated in a future release* → `TestEmbeddedTrustAnchorsParse` fails CI (parse error or keyId mismatch); if it somehow shipped, installs fail closed, never open.
- *Attacker swaps the PEM* → CI keyId-pin fails; and they'd still need the matching private key (in the Environment) to sign an index the swapped public key accepts.
- *Rollback* → revert this commit → `defaultTrustAnchors` empty → universal fail-closed. Safe, non-destructive.
- *Rotation* → add the incoming key as a second PEM block (both trusted during the window), update the pinned keyId, then remove the outgoing key in a later PR.

## Verification

- `TestEmbeddedTrustAnchorsParse` passes asserting the exact production keyIds; 7 parser/loader tests (valid, rotation-window, non-ed25519, wrong-block-type, empty/garbage, both-roles, missing-role) pass.
- Existing `TestInstall_UnsignedIndexFailsClosedAtTrustAnchor` still passes — the pre-existing fail-closed path is undisturbed.
- `go build ./...`, `go test -race ./cmd/conduit/root/connectors/`, golangci-lint (0 issues), `make generate` drift-clean.

## Adversarial self-review

Re-read for the trust-path traps: parser is strict (any non-PUBLIC-KEY block, non-ed25519 key, or zero blocks is an error — no partial-parse trust); keyId derivation is the shared `index.KeyID` (cross-verified against the workflow's independent openssl computation); no fail-open path exists (empty anchors ⇒ `trust_anchor_expired`); embed reads only the two `.pem` paths, ignoring `doc.md`.

## Roadmap

Connector registry MVP bootstrap, Gate 3. Remaining: Gate 4 (L3-sign the 6 seed connectors), Gate 5 (freshness-sign the index), Gate 6 (E2E verify → cut v0.18).


devarismeroxa:
Update: implemented the **louder operator message** (per maintainer request) — reversing design decision #2 above.

A build whose embedded anchors fail to load now refuses up front with a distinct **`registry.trust_anchors_unavailable`** (`codes.Internal`, "reinstall a release build of conduit") instead of the generic `trust_anchor_expired`. `install`/`audit`/`bundle` call `guardTrustAnchors()` before constructing the verifier.

- Fires **only** when the embed is broken (`errAnchorLoad != nil`) — nil in every normal release, so the existing fail-closed path (`TestInstall_UnsignedIndexFailsClosedAtTrustAnchor`) is untouched.
- Can only fail **closed** — never open, never panics unrelated commands.
- New error code registered; `llms.txt`/`llms-full.txt` regenerated (76 → 77 codes).
- Tests: `TestGuardTrustAnchors` (unit) + `TestInstall_BrokenAnchorEmbed_ReportsTrustAnchorsUnavailable` (end-to-end CLI).

PR #2673, opened 2026-07-23
feat(embed): B2 pipelines-in-code builder

devarismeroxa:
## Summary

- Adds a fluent `PipelineBuilder` (`NewPipeline` → `.WithConnector`/`.WithProcessor`/`.WithDLQ` → `.Build()`) to the root `github.com/conduitio/conduit` package, plus `NewSourceConnector`/`NewDestinationConnector`/`NewProcessor`/`NewDLQ` for nested entities. This is v0.19 Workstream 2 ("libconduit"), B2, following B1 (`conduit.New`/`Engine`/`Handle`/`Import`, merged in #2667).
- `Build()` produces exactly the `config.Pipeline` (`PipelineConfig`) the YAML provisioner parses — verified by an AC-6 round-trip test — so `Engine.Import(built)` works identically to importing parsed YAML.
- `Build()` validates via `provisioning/config.Enrich`+`Validate`, the same sequence `conduit pipelines validate` runs, so a Build-time error catches missing id/plugin/type, duplicate connector/processor ids, etc. — but returns the *raw, unenriched* struct, keeping a hand-built `PipelineConfig` indistinguishable from a parsed one.

## Round-trip test coverage (AC-6)

- `TestPipelineBuilder_RoundTrip`: a 432-case deterministic combinatorial matrix (pipeline status/name/description × 4 DLQ variants, including the `*int` nil-vs-explicit-zero distinction × 0-2 connectors × 0-2 processors per scope) asserting `build(x) == parse(marshal(fromConfig(build(x))))`. No `pgregory.net/rapid`/`gopter` dependency added — none exists in this module yet and CLAUDE.md requires justifying new deps; a fixed matrix gives reproducible sub-test names for free and hits ≥100 cases without one.
- `TestPipelineBuilder_RoundTrip_Fixtures`: 3 cases reproducing the *existing* `pkg/provisioning/config/yaml/v2/testdata/pipelines1-success.yml` pipelines field-for-field via the builder (the other AC-6 direction: YAML → parse → builder reproduces it) — reusing real testdata instead of new throwaway fixtures, since Workstream 3's template gallery hasn't landed yet.
- `builder_test.go`: unit tests for Build() validation (missing id/plugin/type, duplicate connector/processor ids → coded `ConduitError`, never a panic), the DLQ window-field nil-vs-explicit-zero contract, settings merge semantics, and the documented nested-builder snapshot-at-attach-time behavior.

## Bugs found and fixed along the way

The round-trip property test caught two real, previously-uncovered bugs in `pkg/provisioning/config/yaml/v2` (`FromConfig` had no production caller before this workstream, so nothing exercised it against `ToConfig`'s output before now):

1. `fromProcessorsConfig` silently dropped `Processor.Condition` when converting `config.Processor` → the v2 wire model.
2. `Connector/Processor/DLQ.ToConfig()` didn't normalize an empty `Settings` map to `nil` the way the sibling slice fields already do (`connectorsToConfig`/`processorsToConfig`), so a nil `Settings` map round-tripping through `FromConfig`+`yaml.Marshal` came back as a non-nil empty map (yaml.v3 marshals a nil map as `{}`, not an omitted key, absent an `omitempty` tag this wire format has never carried).

Both are fixed with regression tests (`TestConfiguration_FromConfig`'s `Condition` field, `TestToConfig_EmptySettingsNormalizedToNil`) verified to fail without the fix and pass with it (reverted the fix locally, confirmed both tests red, restored).

## Adversarial self-review (AI-authored, per CLAUDE.md)

- **Error paths:** `Build()` never panics — all failures flow through `provisioning/config.Validate`'s existing coded `*conduiterr.ConduitError`s (no new error codes minted, so no `allcodes.go` change needed).
- **Concurrency:** builders hold no shared mutable state across goroutines; documented not-safe-for-concurrent-use, matching the rest of the package's builder-pattern conventions.
- **Resource cleanup:** N/A — no resources acquired by this code path.
- **Nil-nested-builder gap (flagged, not fixed):** `WithConnector(nil)`/`WithProcessor(nil)`/`WithDLQ(nil)` panic on the nil-pointer dereference inside `.build()` rather than returning a coded error. B1 set a precedent of guarding `Handle.Stop` against a nil receiver; I considered mirroring that here but the design doc's B2 section doesn't call out this case (only duplicate IDs and the DLQ `*int` distinction are named edge cases), the realistic misuse surface is near-zero since nested builders are always constructed inline as the `With...` call's argument, and adding a "poisoned builder" error-propagation path purely for this added meaningful complexity for a case not in the accepted edge-case table (CLAUDE.md's no-speculative-generality bar). Documented explicitly in `builder.go` instead of silently leaving it unhandled — flagging for review rather than asserting it away.
- **Data-integrity invariants:** none touched — no data-path code, no new serialized format (`PipelineConfig` is the existing alias from B1); this PR only reaches the pipeline-config schema layer.

## Gates run (not faked)

- `make generate && git diff --exit-code` — clean (no drift beyond this PR's intentional changes).
- `golangci-lint run` — 0 issues.
- `go build ./...` — clean.
- `go test -race ./ ./pkg/provisioning/config/...` — all green.
- `go test ./cmd/conduit/internal/llmsgen/...` (`TestAllCodesComplete`) — green (no new error codes added, so no `allcodes.go` change needed).
- `go test ./...` (full repo, no `-race`, for breadth) — all green, confirming the `pkg/provisioning/config/yaml/v2` fix has no other callers affected.

Per CLAUDE.md's Process Maturity table: this PR does **not** claim chaos/upgrade/soak coverage (not live until Phase 2) or a coverage-floor/benchi gate (Phase 1) — only what's live now (lint, race, unit+integration, conventional commits) is asserted as run.

## Risk tier

**Tier 2** (embedding surface / config-schema-adjacent feature), per the design doc (`docs/design-documents/20260722-embed-libconduit-v1.md`) and implementation plan. No data-path, ack/position/checkpoint, or `Runtime.Run` code touched.

## Test plan

- [x] `make generate && git diff --exit-code`
- [x] `golangci-lint run`
- [x] `go build ./...`
- [x] `go test -race ./ ./pkg/provisioning/config/...`
- [x] `go test ./cmd/conduit/internal/llmsgen/...`
- [x] `go test ./...` (full repo)
- [x] Regression tests verified red-without-fix / green-with-fix

Roadmap: v0.19 Workstream 2 ("libconduit" — Embedded v1, Phase 1), B2 slice of `docs/design-documents/20260722-embed-libconduit-v1.md`.


devarismeroxa:
Pushed the three DX must-fixes from the libconduit DX audit (commit deb58c6):

1. **`Engine.ImportPipeline(ctx, *PipelineBuilder) error`** — Build-then-Import in one call. `doc.go`'s package example and `example_test.go`'s `Example_helloPipeline` now use it, dropping the `provisioning/config` import from the common-case quickstart entirely.
2. **Nil-arg panic fixed.** `WithConnector(nil)`/`WithProcessor(nil)`/`WithDLQ(nil)` (pipeline scope) and `ConnectorBuilder.WithProcessor(nil)` (connector scope) now record a `conduiterr.CodeInvalidArgument` error (no new code minted — same code `Run`/`Stop` misuse already uses) instead of panicking; it surfaces, joined with any other accumulated/validation errors, from the next `Build()` call. `Build()` on a nil `*PipelineBuilder` is also now a coded error rather than a nil-pointer panic. New tests cover all of these plus a multi-nil-arg case proving the errors join rather than only the first one surfacing.
3. **Snapshot-at-attach semantics are now explicit in `PipelineBuilder`'s godoc**, with a worked example showing a `*ConnectorBuilder` reused across two pipelines with a mutation in between, and what each pipeline ends up holding. No behavior changed — this was already correct, just undocumented.

Also added `TestImportPipeline_RoundTripsThroughRunningEngine` (success path through a real running `Engine`, mirroring the existing `TestImport_RoundTripsThroughRunningEngine`) and `TestImportPipeline_PropagatesBuildError` (a `Build()` failure never reaches `Import`/the provisioning service).

**Heads-up on `conduit.go`:** I'm aware a parallel B1-hardening branch (lazy DB open) also touches this file — expect a merge conflict there. Not resolving preemptively since the maintainer is sequencing the two; flagging so it's not a surprise at merge time.

Gates re-run and green: `make generate && git diff --exit-code`, `golangci-lint run` (0 issues), `go build ./...`, `go test -race ./ ./pkg/provisioning/config/...`, `go test ./cmd/conduit/internal/llmsgen/...` (`TestAllCodesComplete`), full `go test ./...`.

Still not merging — leaving that to the maintainer per the standing instruction.

devarismeroxa:
## Fix: reference-field snapshot-isolation gap (review finding)

Rebased onto current `main` (picks up the merged sev-0 `Source.Ack` ordering fix, #2680 — disjoint files, no conflicts).

**Bug (confirmed real):** the doc on `PipelineBuilder` promises attaching a nested builder "does not keep a live reference" to it. True for scalar fields (`Name`, `Plugin`, ...), false for reference-typed fields. `ConnectorBuilder.build()` / `ProcessorBuilder.build()` / `DLQBuilder.build()` returned their `provisioningconfig.*` struct by value — which copies a `Settings` map or `Processors` slice *header*, not its contents. So calling `WithSetting`/`WithSettings`/`WithProcessor` on a builder *after* it had already been attached (and the parent `Build()` called) silently mutated the already-built `PipelineConfig` too. Verified with the exact repro from review (`cb.WithSetting` after attach mutates `p1.Connectors[0].Settings` in place).

**Fix:** each `build()` now deep-copies its `Settings` map and (for `ConnectorBuilder`) its `Processors` slice, recursing into each nested `Processor`'s own `Settings` map, so an attached snapshot is fully isolated from any later mutation of the source builder. Added `cloneSettings`/`cloneProcessors` helpers in `builder.go`; nil-safe (preserves nil-vs-empty-map semantics used elsewhere in this package).

**New regression tests** (`builder_test.go`), each verified red-without-fix / green-with-fix via `git stash`:
- `TestConnectorBuilder_SnapshotIsolation_WithSetting` / `_WithSettings` — the exact review repro.
- `TestConnectorBuilder_SnapshotIsolation_NestedProcessorSettings` — a `*ProcessorBuilder` nested inside a `*ConnectorBuilder`'s `Processors`, mutated after the connector itself was attached to a pipeline (the "slice-element aliasing" case).
- `TestProcessorBuilder_SnapshotIsolation_WithSetting` — pipeline-scoped processor.
- `TestDLQBuilder_SnapshotIsolation_WithSetting` — DLQ.

Note on scope: I checked whether *appending* a new element to `Processors`/`Connectors` after attach (as opposed to mutating an existing element's `Settings` map) could itself corrupt an already-built snapshot. It can't — Go copies the slice header (ptr/len/cap) by value on struct assignment, and `append` never writes below the frozen `len` of an earlier holder. Confirmed empirically before writing tests, so I didn't add a test asserting that (it would pass with or without the fix and wouldn't be a meaningful regression check). The real aliasing risk is entirely in the reference-typed fields (maps, and nested structs holding maps), which the deep copy now closes.

**Also flagging per review:** the two `pkg/provisioning/config/yaml/v2` fixes already in this PR (`fromProcessorsConfig` dropping `Condition`, and empty-`Settings` not normalizing to `nil` in `ToConfig`) touch the **production YAML parse path** — the same code every `conduit run` / `conduit pipelines validate` exercises, not just this builder's round-trip tests. They're correct and covered by `TestConfiguration_FromConfig` / `TestToConfig_EmptySettingsNormalizedToNil`, but calling out the shared-path blast radius explicitly so it's not a surprise in review.

**Gates run:**
- `make generate && git diff --exit-code` — clean, no drift beyond this fix's intentional changes.
- `golangci-lint run ./...` — 0 issues on `builder.go`/`builder_test.go` (pre-existing issues elsewhere in the repo, e.g. vendored `internal/diff`, are untouched by this change).
- `go build ./...` — clean.
- `go test -race ./ ./pkg/provisioning/config/...` — all green.
- `go test ./cmd/conduit/internal/llmsgen/... -run TestAllCodesComplete` — green.

Not merging — leaving for maintainer review per Tier 2 process.

PR #2678, opened 2026-07-23
feat(cli): pipelines init --template (vendored gallery)

devarismeroxa:
## Summary

- Adds `conduit pipelines init --template <name>` (vendored, `go:embed`'d gallery: `generator-log`,
  `generator-file`, `postgres-s3`, `postgres-cdc-kafka`) and `--template list --json` (enumeration,
  Family A envelope), per v0.19 Workstream 3.
- Each template is a literal, embedded YAML fixture (not re-templated per invocation) with a README
  (config reference, delivery-semantics note, runnable example).
- `--template` is mutually exclusive with `--source`/`--destination` (coded refusal), reuses #2666's
  `--force`/`--dry-run` conventions, refuses cleanly on an unknown template name or on a
  version-pinned settings mismatch against the connector actually built into this binary.
- 3 new registered error codes (`pipelines.init_unknown_template`,
  `pipelines.init_template_flags_exclusive`, `pipelines.init_template_version_mismatch`).

## Testing

- Unit: catalog validation (reserved name, non-builtin connector, duplicate name, empty YAML),
  version-mismatch checker (real catalog regression guard + synthetic stale-fixture case), flag
  wiring (unknown template, mutual exclusivity incl. the `list` sentinel, force/dry-run reuse,
  custom pipeline name), `--template list --json` envelope conformance.
- E2E, actually executed and passing in this sandbox: `generator-log`, `generator-file` — real
  `pkg/conduit.Runtime`, asserting on destination-side content (log lines / file lines), not just
  YAML parseability.
- E2E, infra-dependent, **NOT executable in this sandbox (no Docker daemon)**: `postgres-s3` (against
  MinIO) and `postgres-cdc-kafka` (against Kafka), behind a dedicated `templates_e2e` build tag and
  `test/compose-templates.yaml`, wired as a new CI matrix job (`template-gallery-e2e`). These compile
  and vet cleanly and were reviewed carefully, but their first real execution will be in CI. Flagging
  per CLAUDE.md rather than claiming a gate that wasn't actually run.
- Gates run: `go build ./...`, `golangci-lint run ./...` (0 issues), `go vet ./...`,
  `go test -race ./cmd/conduit/...` (green), `make generate && git status` (clean, llms.txt/
  llms-full.txt regenerated and idempotent), `go test ./cmd/conduit/cli/...` (schema-golden test
  green — no fixture registration needed since `pipelines init` already implements
  `cecdysis.CommandWithResult`; added `TestTemplateList_JSON_EnvelopeShape` as the payload fixture),
  `go test ./cmd/conduit/internal/llmsgen/...` (green — new codes/command auto-covered via the
  existing `cmd/conduit/root/pipelines` blank-import in `allcodes.go`), full-repo markdownlint
  (0 errors, pinned CI version).

## Adversarial self-review

A background code-review pass found two real issues, both fixed before opening this PR:

- **Blocker (fixed):** the workstream's planning doc (`v019-execution-plan-v2.md` "Workstream 3"
  elaboration) was never committed to this repo — code comments across
  `template_gallery.go`/`init.go`/tests cited its section numbers (e.g. "templates.md §6 AC-3")
  but no such file existed anywhere a reviewer could find it, violating CLAUDE.md's "design doc
  before non-trivial work, in `docs/design-documents/`" gate. Fixed by committing it as
  `docs/design-documents/20260723-templates-gallery.md` (content unchanged from the planning
  version except the status line and a post-implementation note at the end of §12) and
  repointing every in-code citation at the real path.
- **Should-fix (fixed):** the `postgres-s3` and `postgres-cdc-kafka` READMEs both pointed at
  `template_gallery_e2e_test.go` for their E2E coverage — wrong file; their tests actually live in
  `template_gallery_e2e_integration_test.go` (the infra-dependent one). Corrected both.

No correctness, data-integrity, security, or test-quality findings at or above reporting
threshold — mutual-exclusivity logic, the `"list"` sentinel reservation, the wildcard parameter
matcher, and the delivery-semantics claims in each README were all checked against the actual
YAML/code and found accurate.

## Open questions

- `docs/design-documents/20260723-templates-gallery.md` frames itself (per its origin as a
  cross-repo planning note) as needing DeVaris sign-off *before* implementation; this session
  implemented it directly per the assigning task, so that sign-off is still outstanding — flagging
  explicitly rather than treating this PR's existence as implicit approval. In particular, open
  question 1 in that doc's §12 (command-surface decision: `pipelines init --template` vs.
  top-level `conduit init --template`) is a real deviation from the original plan text and
  deserves an explicit look.
- `conduit.io` docs-site source isn't part of this repo (lives elsewhere per the repo map) — not
  updated in this PR; flagging as a known follow-up rather than silently skipping it.
- The MinIO E2E test relies on the s3 connector's client (external, vendored,
  `conduit-connector-s3` — no path-style override hook) defaulting to virtual-hosted-style S3
  addressing, i.e. `*.localhost` resolving to loopback. True on GitHub Actions' Ubuntu runners
  (RFC 6761 + systemd-resolved), but unverified by me directly.

Roadmap: v0.19 Workstream 3 (Templates gallery).
Risk tier: Tier 2 (Features).

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -62,6 +62,50 @@ jobs:
       - name: Build linux/386
         run: GOOS=linux GOARCH=386 go build ./...
 
+  template-gallery-e2e:
+    # End-to-end CI for the two infra-dependent vendored pipeline templates
+    # (v0.19 Workstream 3,
+    # docs/design-documents/20260723-templates-gallery.md §6 AC-3: "infra spun up, pipeline
+    # run, records asserted landed at the destination — not just YAML
+    # parseability"). The two infra-FREE templates (generator-log,
+    # generator-file) need no separate job: their end-to-end tests
+    # (template_gallery_e2e_test.go) carry no build tag, so they already run
+    # as part of the `test` job's `make test-integration` step above, on
+    # every PR, with zero extra CI wiring.
+    #
+    # A matrix entry per template (rather than one job covering both) so a
+    # postgres-s3 regression and a postgres-cdc-kafka regression show up as
+    # distinct, independently-failing checks — matching
+    # docs/design-documents/20260723-templates-gallery.md §6
+    # AC-3's "one CI job per template (or a matrix job)".
+    name: template gallery e2e (${{ matrix.template }})
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - template: postgres-s3
+            run: TestTemplateGalleryE2E_Integration_PostgresS3
+          - template: postgres-cdc-kafka
+            run: TestTemplateGalleryE2E_Integration_PostgresCDCKafka
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Set up Go
+        uses: actions/setup-go@v5
+        with:
+          go-version-file: 'go.mod'
+
+      - name: Bring up template gallery infra (Postgres+MinIO+Kafka)
+        run: docker compose -f test/compose-templates.yaml up --quiet-pull -d --wait
+
+      - name: Run ${{ matrix.template }} end-to-end test
+        run: go test -v -race --tags=templates_e2e -run '${{ matrix.run }}' ./cmd/conduit/root/pipelines/...
+
+      - name: Tear down template gallery infra
+        if: always()
+        run: docker compose -f test/compose-templates.yaml down
+
   docker-build:
     # Build the release Docker image. The image is otherwise only built by the
     # release workflow, so a Dockerfile issue (e.g. a base image whose Go version
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -45,6 +45,41 @@ and that all the tests still run successfully.
 We would like to ask you to use the provided Git hooks (by running `git config core.hooksPath githooks`),
 which automatically run the tests and the linter when pushing code.
 
+## Adding a vendored pipeline template
+
+`conduit pipelines init --template <name>` scaffolds a runnable pipeline from a small,
+permanently-maintained, embedded gallery (`cmd/conduit/root/pipelines/templates/`,
+`cmd/conduit/root/pipelines/template_gallery.go`) — this is the vendoring model the project
+intends to keep long-term (cargo-new / rustup-init style), not a stopgap ahead of a future
+connector/template registry. Adding a template is a normal Tier-2 PR: one reviewer approval,
+green CI, docs in the same PR — it is **not** the registry's signed-publish flow (there is no
+`conduit templates publish`, and this contribution path has nothing to do with registry trust or
+signing).
+
+To propose a new template:
+
+1. **Use only built-in connectors.** Every source/destination the template configures must be a
+   key in `builtin.DefaultBuiltinConnectors` (`pkg/plugin/connector/builtin/registry.go`) — a
+   template requiring a manual connector download reintroduces exactly the cliff this gallery
+   exists to avoid, and `TestValidateGalleryCatalog_RejectsNonBuiltinConnector`-style checks will
+   fail the build if it doesn't.
+2. **Add a directory** under `cmd/conduit/root/pipelines/templates/<name>/` with a `pipeline.yaml`
+   (the literal, fully-rendered pipeline config — this is embedded via `go:embed` and is the exact
+   bytes `init` writes to disk) and a `README.md` covering: a config reference table, a
+   delivery-semantics note (state only what the underlying connectors actually guarantee — see
+   CLAUDE.md's Invariant 3, at-least-once is the floor), and a runnable example.
+3. **Register it** in `galleryCatalogSpec()` (`template_gallery.go`) with a one-line
+   `Description` and `DeliverySemantics` string (surfaced by `--template list --json`). The name
+   must never be `list` — that value is reserved to mean "enumerate the catalog".
+4. **Add an end-to-end test**, not just a parse check: spin up whatever infra the template needs
+   (reuse `test/compose-templates.yaml`'s pattern if it needs its own containers), scaffold the
+   template for real, run it against the real engine, and assert records actually land at the
+   destination.
+5. Reviewer checklist (Tier 2): built-in connectors only (1), README sections present and
+   accurate against what the new CI job actually asserts (2), catalog entry registered (3), and a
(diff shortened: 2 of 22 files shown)


devarismeroxa:
## Root cause (confirmed)

Both `postgres-s3` and `postgres-cdc-kafka` templates
(`cmd/conduit/root/pipelines/templates/*/pipeline.yaml`) repeat the placeholder
connection string in an explanatory comment directly above the actual `url:`
setting:

```yaml
          # Postgres connection string, e.g.
          # postgres://user:password@localhost:5432/dbname?sslmode=disable
          url: postgres://user:password@localhost:5432/dbname?sslmode=disable
```

`template_gallery.go` embeds these files with `go:embed` and writes the
**literal bytes** to disk on scaffold (comments included) — confirmed by
that file's own doc comment. So the E2E test's
`strings.Replace(edited, "postgres://user:password@...", templatesE2EPostgresURL, 1)`
matched the **first** occurrence — the comment — and left the real `url:`
line, which is exactly the field `conduit-connector-postgres`'s source
config reads (`source/config.go`'s `URL string \`json:"url"\``), untouched.
The connector kept dialing the placeholder host/port/user/db, so zero rows
ever reached MinIO/Kafka. The `is.True(edited != original)` guard still
passed because the comment text did change — it just wasn't proof the
runtime-effective setting changed.

This is not a "different settings shape" issue (the connector's key genuinely
is `url`, matching the template) — it's a duplicate-literal-string /
first-occurrence bug in the test's patching logic.

## Fix

Added `patchPostgresSourceURL` in
`cmd/conduit/root/pipelines/template_gallery_e2e_integration_test.go`:

- Anchors the replacement on the `url: ` key prefix (`url: postgres://...`),
  which is unique in the file (the comment line has a `# ` prefix, not
  `url: `), so it targets only the real setting.
- Asserts exactly one such settings line exists before patching (fails loudly
  if the template YAML shape ever changes).
- Asserts, after patching, that the effective `url: ` line now holds the real
  test DSN and the placeholder `url: ` line is gone — i.e. it checks the
  *runtime-effective* setting changed, not just "some string in the file
  changed."

Both `TestTemplateGalleryE2E_Integration_PostgresS3` and
`TestTemplateGalleryE2E_Integration_PostgresCDCKafka` now go through this
helper.

## Gates run locally (no Docker available in this sandbox)

- `go build ./...` — clean
- `go vet -tags templates_e2e ./cmd/conduit/root/pipelines/...` — clean
- `go test -tags templates_e2e -c -o /dev/null ./cmd/conduit/root/pipelines/` (compiles the E2E test binary without running it) — clean
- `go test -short ./cmd/conduit/root/pipelines/...` (all non-Docker tests in the package) — 113 passed
- `golangci-lint run --build-tags templates_e2e ./cmd/conduit/root/pipelines/...` — no issues
- `make generate && git diff --exit-code` — no unrelated diff
- Rebased cleanly onto current `origin/main` (2 commits, including the merged sev-0 fix #2680) — no conflicts

**I could not run the Docker-gated `postgres-s3` / `postgres-cdc-kafka` E2E
jobs themselves in this sandbox.** I'm confident this fixes the specific bug
diagnosed (the patch now demonstrably lands on the real `url:` setting
instead of the comment), but I have not observed the connector actually
connect and move rows through the real compose infra — that's what CI's
Docker-gated jobs need to confirm. Not claiming green until they report it.

Not merging — leaving this for CI + review.

devarismeroxa:
## Update: postgres-s3 E2E follow-up (postgres-cdc-kafka confirmed green)

Pulled the failing job log (`gh api repos/ConduitIO/conduit/actions/jobs/89347354007/logs`,
run 30049222733). The `url:` patch is working correctly here too — the
log shows the postgres source connecting, running its snapshot, and
fetching all 3 rows fine:

```text
INF "my_table" snapshot completed ... rows=3 table=my_table
```

The actual failure is on the **S3/MinIO destination side**:

```text
ERR error writing record error="operation error S3: PutObject, https response
error StatusCode: 400, ... api error MalformedXML: The XML you provided was
not well-formed or did not validate against our published schema."
```

## Root cause

`conduit-connector-s3`'s writer (`destination/writer/s3.go`) builds its AWS
client with plain `config.LoadDefaultConfig` + `s3.NewFromConfig` — no
override for checksum behavior. Since `aws-sdk-go-v2` v1.30 / `config`
v1.27 (2024), the **default** `RequestChecksumCalculation` is
`when_supported`: `PutObject` unconditionally attaches a trailing CRC32
checksum via `aws-chunked` transfer encoding. `test/compose-templates.yaml`'s
`minio/minio:latest` can't parse that chunked+trailer framing and returns
the 400 MalformedXML seen above — a known aws-sdk-go-v2/MinIO
compatibility gap, not a credentials/bucket/endpoint problem (the
connector did reach MinIO with the right bucket and creds; if those were
wrong we'd see 403/NoSuchBucket, not a parse error on the request body).

This is not the same duplicate-literal/first-occurrence bug as the `url:`
setting — the bucket/access-key/secret placeholders in
`templates/postgres-s3/pipeline.yaml` are each unique strings and were
already being patched correctly.

## Fix

`config.LoadDefaultConfig` resolves `RequestChecksumCalculation` (and its
response-side counterpart, `ResponseChecksumValidation`) from the
`AWS_REQUEST_CHECKSUM_CALCULATION` / `AWS_RESPONSE_CHECKSUM_VALIDATION`
env vars (`aws-sdk-go-v2/config`'s `env_config.go`) — the same mechanism
the test already relies on for `AWS_ENDPOINT_URL`. Setting both to
`when_required` in `TestTemplateGalleryE2E_Integration_PostgresS3` reaches
the connector's client without any `conduit-connector-s3` code change.

## Gates run locally (still no Docker in this sandbox)

- `go build ./...` — clean
- `go vet -tags templates_e2e ./cmd/conduit/root/pipelines/...` — clean
- `go test -tags templates_e2e -c -o /dev/null ./cmd/conduit/root/pipelines/` — compiles
- `go test -short ./cmd/conduit/root/pipelines/...` — 113 passed
- `golangci-lint run --build-tags templates_e2e ./cmd/conduit/root/pipelines/...` — no issues
- `make generate && git diff --exit-code` — no unrelated diff

I'm confident this targets the actual failure mechanism the CI log shows
(a genuine S3-wire-protocol response, not a test-side placeholder bug),
but I have not observed a passing run — CI is the real validation here.
Not claiming green, not merging.

PR #2679, opened 2026-07-23
docs(design): Source.Ack persist-ordering fix + sev-0 postmortem

devarismeroxa:
## Summary

Requires DeVaris Tier-1 sign-off on the fix APPROACH before the code PR.

`Source.Ack` (`pkg/connector/source.go:207-238`) sends the ack to the plugin (line 218) — driving
the plugin's own upstream commit, e.g. a Postgres replication slot's `confirmed_flush_lsn` advance
→ WAL prune — **before** `persister.Persist` (lines 223-235) durably persists the resulting
position. `Persister.Persist` (`pkg/connector/persister.go:137-170`) only enqueues into an
in-memory batch, flushed asynchronously (`flushNow`, `persister.go:238-271`) on a ~1s debounce
(`DefaultPersisterDelayThreshold`) or a 10k-item threshold. A crash in that window loses the
durable position write after the plugin has already been told to commit.

**Confirmed, not hypothetical:** the DBZ-1 SIGKILL chaos test (PR #2677, `tests/chaos/`) drives
this exact window against a synthetic upstream with a `prune` toggle — identical engine code,
identical crash timing. Against a durable/retention-based upstream (Kafka-like) the window
produces only a benign duplicate. Against a pruning upstream (Postgres-replication-slot-like) it
produces a real, structural `OPEN_GAP_ERROR` on restart. This is a **latent, pre-existing sev-0**
(violates invariants 1, 2, 3) — the repo's first chaos test is what caught it.

This PR is **documentation only**: a Tier-1 design doc and a blameless postmortem. No code
changes; the fix is a separate PR after sign-off.

## What's included

- `docs/design-documents/20260723-source-ack-persist-ordering-fix.md` — problem, mechanism traced
  file:line, three approaches (**A, recommended: ack-follows-durable-flush** via a post-flush
  callback + per-partition high-water-mark tracking, no protocol change; **B**: synchronous
  persist-then-ack, rejected on throughput grounds; **C**: new connector-protocol durability
  signal, rejected as a breaking protocol change that spreads the burden to every connector),
  failure-mode analysis crash-by-crash, graceful-shutdown handling (`StopAndWait` must wait for
  the deferred ack, not just the flush), the MySQL/Mongo must-verify flag, regression-gate plan
  (un-skip DBZ-1's `prune=true` assertion once fixed), and a benchi-gated rollout plan.
- `docs/postmortems/20260723-source-ack-persist-ordering.md` — blameless postmortem: what the bug
  is, that it's latent across all shipped releases, impact (Postgres CDC / KC-wrapper
  Debezium-Postgres confirmed; MySQL/Mongo not yet classified), how it was detected (DBZ-1, the
  repo's first chaos test), timeline, remediation, and follow-ups.

## Adversarial self-review

- Verified every file:line citation against the actual tree (`pkg/connector/source.go`,
  `pkg/connector/persister.go`, `pkg/lifecycle-poc/funnel/worker.go`, `pkg/lifecycle/service.go`)
  and against PR #2677's description rather than re-deriving from memory.
- Checked that Approach A's graceful-shutdown story doesn't just move the bug: `StopAndWait`
  already blocks on `WaitPendingWrites` (flush) today, but under A it must additionally block on
  the *deferred ack* firing, or a graceful teardown could still drop the final ack silently. Called
  this out as a required part of the fix, not a nice-to-have.
- Did not claim the MySQL/Mongo connectors are safe — flagged both as inference-only,
  must-verify-before-claiming-fixed.
- Confirmed markdownlint clean (`npx markdownlint-cli2@0.18.1` on both files, 0 errors) and that
  `git diff` touches only these two new files.

## Gates

- `npx markdownlint-cli2@0.18.1`: 0 errors on both new files.
- `git diff --stat`: two new files only, no code changes.

## Risk tier

**Tier 1** (data path: ack/position/checkpoint ordering). No code change in this PR; the
follow-up fix PR will carry the full Tier-1 bar (chaos test update, failure-mode analysis, human
sign-off). This PR itself requires DeVaris's sign-off on the **approach** before that follow-up is
opened.

**DO NOT MERGE** without that sign-off.

### DIFF
--- a/docs/design-documents/20260723-source-ack-persist-ordering-fix.md
+++ b/docs/design-documents/20260723-source-ack-persist-ordering-fix.md
@@ -1,0 +1,325 @@
+# Source.Ack persist-ordering fix
+
+## Summary
+
+**Confirmed sev-0, not hypothetical:** `Source.Ack` (`pkg/connector/source.go:207-238`) sends the
+ack to the plugin (`s.stream.Send`, line 218) **before** the resulting position reaches durable
+storage (`s.Instance.persister.Persist`, lines 223-235, which only enqueues into an in-memory batch
+— see `pkg/connector/persister.go:137-170`). The batch is flushed asynchronously by `flushNow`
+(`persister.go:238-271`) on a debounce: `DefaultPersisterDelayThreshold` (1s, `persister.go:28`) or
+`DefaultPersisterBundleCountThreshold` (10k items, `persister.go:29`), whichever comes first. Any
+crash inside that window loses the position write while the plugin has already been told "you may
+commit."
+
+For a plugin whose upstream commit is reversible/idempotent from Conduit's point of view (Kafka:
+retention-based, replay from an older offset just redelivers), the crash window degrades to a
+**benign duplicate** — consistent with the at-least-once floor (invariant 3). For a plugin whose
+upstream commit **prunes** data it considers committed (Postgres logical replication: acking a
+position drives `confirmed_flush_lsn` forward, and WAL segments behind that point can be recycled),
+the identical crash window produces a **structural gap**: Conduit resumes from the last _durably
+persisted_ position, which is now known to be behind data the upstream has already discarded.
+Confirmed empirically by the DBZ-1 chaos test (`gh pr diff 2677`, `tests/chaos/`) against a
+synthetic `upstreamStore` with a `prune` toggle — identical engine code, identical crash timing,
+`prune=false` → duplicate delivery, `prune=true` → `OPEN_GAP_ERROR` on restart.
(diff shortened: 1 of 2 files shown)


devarismeroxa:
Superseded by #2680, which merged the design doc AND postmortem to main in their final form — aligned to the shipped connector-level FIFO seq (not the per-partition HWM this draft described), with the bounded-Teardown failure mode and the process-wide head-of-line-blocking note added, per the fresh-context review and DeVaris's Tier-1 sign-off. Both files are on main; this draft is stale. Closing to avoid reverting the alignment.

PR #2692, opened 2026-07-26
docs(design): DBZ-2 CDC correctness / acceptance-test suite (v0.20 WS6)

devarismeroxa:
## What

Tier-1-adjacent design doc for **DBZ-2** — the CDC correctness / acceptance-test suite that defines "done" for every CDC connector and is the critical-path gate before DBZ-3. v0.20 Workstream 6. Advances the Debezium-compete roadmap (DBZ-2 row).

## Scope (hybrid, post-review)

Generalizes the DBZ-1 chaos harness (`tests/chaos`). Two parts:
- **Engine-side core** (this v0.20 deliverable, in `tests/chaos`, gated by the required `tests/chaos (race, x3)` check): shared engine guarantees — at-least-once crash-safety under `kill -9` on the ack→persist window, strict-FIFO ack ordering + cumulative-position monotonicity (#2680 contract), per-partition ordering (engine half), transport half of invariant-6.
- **Connector-facing contract**: an SDK-acceptance profile each native connector runs against its real DB + a prose checklist for the JVM KC-wrapper. This — not the engine core — is what catches connector-internal CDC bugs (e.g. conduit-connector-mysql#182).

## Load-bearing boundary

Real DDL reconstruction/policy and real per-partition offset independence are **deferred to DBZ-3 / the contract**; faking a schema engine synthetically is explicitly rejected as false confidence.

## Adversarial review — SOUND-WITH-CHANGES, all folded

Independent review (grounded in harness + `pkg/connector`) confirmed the load-bearing splits are correct and code-grounded. Caught + fixed: mid-handoff / Property-1-base were engine-vacuous (engine persists position two-state empty|valid, `child.go:121-139`/`source.go:394` — no distinct "boundary" state; re-scoped to a producer-pacing variant + happy-path smoke test, honest two-state discriminator); #182 framing corrected to contract-only; wrapper claim narrowed to the engine-seam half; per-partition guarantee asserted (not an open question); Property-4 flagged as a new `pkg/lifecycle-poc/funnel`+DLQ integration.

## Phasing

Phase 1 (DBZ-3 gate): two-phase snapshot/stream synthetic plugin + mid-handoff/mid-position-write SIGKILL + strict-FIFO ack assertion. Phase 2: multi-key ordering. Phase 3 (may trail): transport-level invariant-6 (needs synthetic destination + DLQ).

## Open questions for maintainer

1. Invariant-6 transport half in DBZ-2 core now, or defer all invariant-6 to DBZ-3?
[...]

PR #2693, opened 2026-07-26
test(chaos): DBZ-2 Phase 1a — CDC correctness Properties 1, 2, 3

devarismeroxa:
## What

DBZ-2 Phase 1a — implements CDC correctness **Properties 1, 2, 3** from the design doc (`docs/design-documents/20260726-dbz2-cdc-correctness-suite.md`, merged #2692). Generalizes the DBZ-1 chaos harness (`tests/chaos`, whose `doc.go` calls for exactly this). v0.20 Workstream 6. These three properties are the definition-of-done that unblocks DBZ-3.

## Properties

- **P1 — snapshot→stream continuity (no-crash smoke test).** Two-phase `chaosPlugin` (fast snapshot burst 1..K, then paced stream K+1..N) with a `HANDOFF` marker; asserts every position committed once, in order, monotone resume. Honestly labeled a smoke test — the engine has no distinct snapshot/stream state, so this is a tripwire + the fixture P2 builds on, **not** handoff-atomicity coverage (that's a connector property → DBZ-3).
- **P2 — at-least-once under SIGKILL at three points** (mid-snapshot / mid-handoff / mid-position-write), each against **both** `prune=false` (Kafka-like) and `prune=true` (Postgres-slot-like) upstreams. Asserts: resume ≥ upstream watermark at kill (no gap), no `OPEN_GAP_ERROR` even against pruning, no `CORRUPT_POSITION` (inv 2), `DONE` with `committed == total` (at-least-once). Uses the honest **two-state** `RESUME_POSITION` discriminator (empty vs valid-stale — there is no third "boundary" state; mid-handoff is documented as a producer-pacing variant that lands in the same state as mid-snapshot).
- **P3 — per-partition ordering.** Multi-key plugin + a real per-key delivery **ledger** (not a max-position counter): asserts every `(key,seq)` acked exactly once (gaps AND dupes), Ack-call-order FIFO arrival, and strict per-key monotonicity in delivery order. Exercises `onPersistFlushed`'s head-of-queue drain by `seq`.

## The tests actually catch regressions (not tautologies)

Proven by perturbation (each reverted, zero net diff):
- Reverting `Source.Ack` to pre-sev-0 (synchronous ack before persist) → all three P2 pruning-upstream subtests **fail** on `resume ≥ watermark`.
- Reversing `onPersistFlushed`'s delivery order → P3 **fails** with a per-key ordering-regression message.

## Verification (in-process — no docker/DB, so this is real, not CI-deferred)

`go test -race -count=3 -shuffle=on ./tests/chaos/...` → **PASS, 48 subtests, zero races** (run independently by the reviewer, and twice by the author). `go build`/`go vet` clean; repo-pinned `golangci-lint` (tools/go.mod) → 0 issues. DBZ-1's original `sigkill_test.go` is byte-for-byte unaffected (new plugin capabilities are opt-in via zero-valued fields).

## Scope

Properties 1+2+3 only. **Property 4** (invariant-6 transport half — the funnel+DLQ+destination integration) and the **SIGTERM/invariant-7** case land in a follow-up PR (Phase 1b), per the design doc's decisions.

## Risk tier

Tier-1-adjacent test infrastructure (does not change engine behavior — it tests it). Gated by the required `tests/chaos (race, x3)` check. Adversarial self-review: assertions verified specific + non-tautological (perturbation proof above); two-state discriminator honest; no false-confidence handoff claim.

Advances the Debezium-compete roadmap (DBZ-2). Design: #2692.

### DIFF
--- a/tests/chaos/child.go
+++ b/tests/chaos/child.go
@@ -38,19 +38,24 @@ import (
 // conduit run's actual config surface (see doc.go / the design doc's
 // observability section on this exact point).
 const (
-	envChild        = "CONDUIT_CHAOS_CHILD"
-	envDBDir        = "CONDUIT_CHAOS_DB_DIR"
-	envUpstreamDir  = "CONDUIT_CHAOS_UPSTREAM_DIR"
-	envPrune        = "CONDUIT_CHAOS_PRUNE"
-	envPaceMS       = "CONDUIT_CHAOS_PACE_MS"
-	envTotal        = "CONDUIT_CHAOS_TOTAL"
-	instanceID      = "chaos-source"
-	instancePlugin  = "chaos-plugin"
-	instancePipe    = "chaos-pipeline"
-	markerOpenGap   = "OPEN_GAP_ERROR"
-	markerDone      = "DONE"
-	markerFatal     = "FATAL"
-	markerCorruptPo = "CORRUPT_POSITION"
+	envChild          = "CONDUIT_CHAOS_CHILD"
+	envDBDir          = "CONDUIT_CHAOS_DB_DIR"
+	envUpstreamDir    = "CONDUIT_CHAOS_UPSTREAM_DIR"
+	envPrune          = "CONDUIT_CHAOS_PRUNE"
+	envPaceMS         = "CONDUIT_CHAOS_PACE_MS"
+	envTotal          = "CONDUIT_CHAOS_TOTAL"
+	envSnapshotK      = "CONDUIT_CHAOS_SNAPSHOT_K"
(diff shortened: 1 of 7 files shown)


devarismeroxa:
**Tier-1 sign-off — DeVaris, 2026-07-26.** Approved. Properties 1+2+3 land as the engine-side DBZ-3 gate. Assertions verified specific + non-tautological (perturbation proof in the PR body); two-state RESUME_POSITION discriminator honest; chaos gate (race, x3) green in CI. Property-4 transport half + SIGTERM case follow in Phase 1b.

PR #2694, opened 2026-07-26
test(chaos): DBZ-2 Phase 1b — invariant-6 transport half + SIGTERM/invariant-7

devarismeroxa:
## What

DBZ-2 **Phase 1b** — completes the DBZ-2 engine-side core: **Property 4 (invariant-6 transport half)** and the **SIGTERM/invariant-7** case (design doc `docs/design-documents/20260726-dbz2-cdc-correctness-suite.md`, merged #2692; builds on Phase 1a #2693). v0.20 WS6.

## Property 4 — invariant-6 transport half (never silently mangled)

The chaos harness previously acked directly, bypassing all routing. This stands up a **real** funnel pipeline in-process: `SourceTask` (wrapping a real `*connector.Source`, identical construction to Phase-1a) → `driftDetectTask` → `DestinationTask` (synthetic destination), with a **real `funnel.DLQ`**. Per the design's Q1 boundary, `driftDetectTask` is **not a schema engine** — it only marks `chaosPlugin`'s synthetic drift marker `Nack`'d; the real DLQ `windowNackThreshold` makes the deliver/DLQ/halt decision, exactly as production policy would.

- **DLQ-routing** (window disabled): the drift record lands in the DLQ destination (never the main), is acked upstream (handled, not dropped); `committed` reaches `total`.
- **Halt** (DLQ disabled): the pipeline halts on the drift record; positions before it are delivered/acked; the drift record and everything after are delivered **nowhere**; the upstream **and** Conduit's own persisted position **never advance past the last handled record**; a modeled restart **redelivers exactly the drift record** (at-least-once preserved).

## SIGTERM / invariant-7

Graceful `Teardown` must drain the deferred ack before `stopStream` (`source.go:249-326`). Lands in the mid-position-write window (a real deferred ack in flight when the signal arrives), asserts the upstream watermark advances past what was durable before the signal (deferred ack **flushed, not dropped**), and cross-checks Conduit's own persisted position against the upstream watermark (catching an invariant-1 violation).

## Verified non-tautological (perturbation proof)

Breaking drift-detection (making it never fire → drift record silently passes through) makes **both P4 tests fail** — the halt test times out waiting for a halt that no longer happens. Reverted (zero net diff) → green. The SIGTERM assertion (`committedAfter > committedBefore`, an explicit SEV-0-class failure otherwise) was verified by review. The P4 halt assertion (`committed==2`-not-`total`, position never advances, exact drift record redelivered) is manifestly specific.

## Verification (in-process — real funnel + DLQ + synthetic destination, no docker)

`go test -race -count=3 -shuffle=on ./tests/chaos/...` → **57 passed** (48 from Phase 1a + 9 new subtests), zero races, single uncontended run. `go build`/`go vet` clean; repo-pinned `golangci-lint` → 0 issues. DBZ-1's `sigkill_test.go` + Phase-1a tests unaffected (new capabilities opt-in via zero-valued fields; the SIGTERM child dispatches only via an env flag).

## Scope / status

This completes **DBZ-2 Phase 1** (Properties 1–4 + SIGTERM). The definition-of-done that unblocks DBZ-3 (Properties 1+2+3) was already met by #2693; this adds the invariant-6 transport gate the AI-pipeline record work depends on. Real DDL reconstruction/policy remains deferred to DBZ-3 against a real DB, per the design boundary.

## Risk tier

Tier-1-adjacent test infrastructure (tests engine behavior, doesn't change it). Gated by the required `tests/chaos (race, x3)` check. Adversarial self-review: real funnel/DLQ (not fakes), honest Q1 boundary (drift-detect is not a schema engine), assertions specific + perturbation-proven.

Design: #2692 · Phase 1a: #2693

### DIFF
--- a/tests/chaos/child.go
+++ b/tests/chaos/child.go
@@ -18,8 +18,12 @@ import (
 	"context"
 	"fmt"
 	"os"
+	"os/signal"
 	"strconv"
 	"strings"
+	"sync"
+	"sync/atomic"
+	"syscall"
 	"time"
 
 	"github.com/conduitio/conduit-commons/database"
@@ -47,6 +51,8 @@ const (
 	envSnapshotK      = "CONDUIT_CHAOS_SNAPSHOT_K"
 	envSnapshotPaceMS = "CONDUIT_CHAOS_SNAPSHOT_PACE_MS"
 	envNumKeys        = "CONDUIT_CHAOS_NUM_KEYS"
+	envDriftAt        = "CONDUIT_CHAOS_DRIFT_AT"
+	envSigtermMode    = "CONDUIT_CHAOS_SIGTERM_MODE"
 	instanceID        = "chaos-source"
 	instancePlugin    = "chaos-plugin"
 	instancePipe      = "chaos-pipeline"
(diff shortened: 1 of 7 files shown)


devarismeroxa:
**Tier-1 sign-off — DeVaris, 2026-07-26.** Approved. Completes DBZ-2 Phase 1: invariant-6 transport half (real funnel.Worker + DLQ, drift record never silently acked without handling, position never advances past an unhandled record on halt) + SIGTERM/invariant-7 (Teardown drains the deferred ack). Real funnel/DLQ (not fakes), honest Q1 boundary (drift-detect is not a schema engine), assertions perturbation-proven, 57 chaos subtests green under race x3. Merging (test job green after #2695 de-flake).

PR #2695, opened 2026-07-26
test(connector): de-flake the Teardown deferred-ack regression tests

devarismeroxa:
## What

De-flakes the two #2680 sev-0 regression tests in `pkg/connector` — `TestSource_Teardown_SendsPendingDeferredAckBeforeReturning` and `TestSource_Teardown_FastFlushCompletesWithinBoundedTimeout`. They flake in CI (pass locally), intermittently reddening the `test` job on unrelated PRs (surfaced on #2694).

## Root cause — a test-side goroutine race, not an engine bug

Each test spawns a receiver goroutine (`stream.Recv()` → chan) and then checks the chan with a **non-blocking `select`/`default`** immediately after `Teardown` returns. Even when `Teardown` correctly sent the deferred ack before returning, the receiver goroutine may not have been scheduled yet when the `default` branch fires → false failure. CI's slower, contended scheduling loses that race; local runs win it.

## Fix — without masking the regression these guard

A naive bounded wait would **mask** the exact regression these tests exist to catch: with `delayThreshold=time.Hour`, a regression where `Teardown` returns *without* waiting for the flush would let the async flush deliver the ack a few microseconds late — a bounded wait would still receive it and pass.

So `SendsPendingDeferredAckBeforeReturning` is made **deterministic** via the existing `blockingDB`: while the flush (the only thing that sends the ack) is blocked, assert `Teardown` has **not** returned and the ack has **not** been sent; unblock, then assert both complete. `FastFlush` keeps its fast-path/no-false-timeout job with a bounded wait for the receiver goroutine — the strict before-return guarantee is now owned deterministically by the test above.

## Verified

- **Perturbation proof:** removing `Teardown`'s `WaitPendingWritesContext` (the flush-wait) makes the deterministic test **fail** ("did not wait to drain the ack") — proving it catches the regression a bounded wait would mask. Reverted, green.
- **Stress:** `go test -race -count=100` on all three Teardown tests → **300 passed**.
- Test-only; no engine change. `blockingDB` (already in the file) reused.

## Risk tier

Tier 3 (test-only), but hardens a sev-0 (#2680) regression gate's reliability. No `pkg/connector` production code touched.

### DIFF
--- a/pkg/connector/source_test.go
+++ b/pkg/connector/source_test.go
@@ -394,14 +394,30 @@ func TestSource_Teardown_SendsPendingDeferredAckBeforeReturning(t *testing.T) {
 	ctx := context.Background()
 	ctrl := gomock.NewController(t)
 
-	// A long delay threshold and high bundle-count threshold mean the ack
-	// below would NOT be auto-flushed for the lifetime of this test - the
-	// only thing that can flush it is Teardown's own forced Flush call.
+	// This test proves invariant 7 (Teardown drains the deferred ack before
+	// returning) DETERMINISTICALLY, without racing a receiver goroutine's
+	// scheduling against a non-blocking check. It does so by gating the flush -
+	// the only thing that sends the deferred ack - on a blockingDB the test
+	// controls: while the flush is blocked, Teardown must NOT return and the ack
+	// must NOT be sent; only once the flush is unblocked does the ack go out and
+	// Teardown return. A regression that let Teardown return WITHOUT waiting for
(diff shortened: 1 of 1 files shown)


devarismeroxa:
**Sign-off — DeVaris, 2026-07-26.** Approved. Test-only de-flake of the #2680 sev-0 Teardown regression tests; the deterministic blockingDB rewrite provably catches the flush-wait regression a bounded wait would mask (perturbation-verified), 300x -race stress green. Merging.

PR #2687, opened 2026-07-24
docs(design): AI-pipeline components (v0.20 WS8)

devarismeroxa:
## Summary

Gating design doc for v0.20 Workstream 8 — the canonical Conduit AI workload, CDC -> chunk ->
embed -> vector store. Per CLAUDE.md, this is a **new subsystem** and gets a design doc before any
code; per the v020 execution plan, **no WS8 implementation PR opens until this doc has DeVaris's
sign-off.**

- Names the new dedicated repos: `conduit-processor-ai` (chunking + embedding processors),
  `conduit-connector-pgvector` (v0.20), `conduit-connector-qdrant` (fast-follow/contingency).
  `conduit-processor-sdk` is Go/WASM-only and these don't belong in it.
- Embedding-provider decision held to the same rigor `conduit generate`'s LLM-provider decision
  got: OpenAI / Voyage / local-via-Ollama, deterministic resolution, zero-new-dependency adapters,
  cost "bounded not quantified" per CLAUDE.md's no-benchmark-no-claim rule.
- Resolves the load-bearing architecture question: embedding needs outbound HTTP, but
  `conduit-processor-sdk` standalone processors run on WASI Preview 1 (no sockets). Chosen answer:
  extend the existing host-mediated capability channel (`pprocutils.SchemaService` precedent) with
  one new, allowlisted, host-executed HTTP capability — not a new plugin transport, not a built-in
  escape hatch. This is the one core-engine-adjacent change the doc calls out for its own scoped
  review (Open Questions #2).
- Record-path failure modes per invariants 1–7: provider down/rate-limited/partial-batch (no early
  ack, no silent drop), upsert idempotency under retry (deterministic chunk IDs + `ON CONFLICT`
  upsert), dimension validation at pipeline start (fail fast, coded error), token-cost backpressure
  (bounded batching, synchronous `Process` semantics give backpressure for free), tombstones ->
  vector deletes (matched by `source_key`, not a guessed chunk-ID set).
- The RAG-sync template is the first template needing non-built-in components, so it's designed as
  registry-install-backed, not vendored — which requires a scoped exception to the templates
  gallery's existing built-in-only CI rule (flagged, not resolved unilaterally, in Open Questions).
- >=2 alternatives considered per major decision (processor transport, single vs. per-store vector
  connector, in-WASM vs. delegated local embedding, write-path-only scope) with why they lost.

## Risk tier

**Tier-1-adjacent.** Embedding processors and vector destinations sit on the record path; this
[...]

PR #2696, opened 2026-07-26
docs(design): WASM host-egress capability (AI-pipeline prerequisite, SSRF-hardened)

devarismeroxa:
## What

Focused design for the **WASM host-mediated HTTP egress capability** — the one piece of the AI-pipeline subsystem that touches the core engine and is a NEW security boundary. Per the AI-pipeline sign-off (#2687), this gets its own review BEFORE the embedding processor is built against it. This is that doc.

## Mechanism

One new `http_request` export on the wazero `conduit` host module (same `handleWasmRequest` park/resize protocol + `pprocutils.Error` band as the existing schema calls). The guest marshals a request; the HOST (full `net/http`) performs it under policy. Buffered, single-shot, no streaming; the guest never gets a socket. Response body via `io.LimitReader`.

## Security boundary (the point of the doc)

- **Deny-by-default**, three-tier allowlist (deny-all → per-processor pipeline YAML → engine-level ceiling intersection), bound host-side per-processor instance, immutable from guest.
- **The load-bearing gate:** a `net.Dialer.Control` hook validates the **resolved IP on every dialed candidate**, refusing RFC1918/loopback/link-local/ULA/v4-mapped/**NAT64**/metadata (169.254.169.254) unless that exact IP is explicitly allowlisted. Closes the DNS-rebinding/TOCTOU window a string-only allowlist leaves open.
- Redirects not followed; `Host` and `Accept-Encoding` host-controlled; per-call timeout + size cap.

## Adversarial security review — SOUND-WITH-CHANGES, all 6 must-fixes folded

A `security-engineer` review attacked it as an adversary and confirmed the architecture is right, but found (all now fixed in the doc):

1. **Proxy-env bypass (was a complete bypass):** `HTTP_PROXY`/`HTTPS_PROXY` would tunnel every dial through a `CONNECT` the resolved-IP check never inspects → `Transport.Proxy` pinned `nil`, env ignored, tested.
2. **NAT64 `64:ff9b::/96`** (embeds the metadata IP, missed by v4-mapped unwrap) → added to refused ranges + unwrap + fuzz test.
3. **Carve-out scoped to (IP, port)**, not IP-only (else allowlisted Ollama IP could reach Redis on another port) → + test.
4. **`Control` governs http AND TLS dials** (no `DialTLSContext` bypass); rebinding test runs against real `https`.
5. **`Accept-Encoding` reserved host-side** so the decompression cap stays on the decompressed stream.
6. **Per-processor policy isolation is a hard, tested invariant** — the reviewer traced the real code and found the mirrored `schemaService` is a process-global singleton; a naive impl would leak one pipeline's allowlist cross-tenant. No policy field may live on `Registry`; it must be threaded per-call (three widened signatures). Multi-tenant isolation test required. (Verified against the live-hot-reload path that reconfigure rebinds policy — the doc states this with citation.)

**Resolved (was open question 3):** API keys are **host-injected from the secret store by name — mandatory, no guest-supplied `Authorization` path**, so a compromised/third-party WASM binary categorically cannot steal the vendor key (it never enters the sandbox).

## Status

Design only. Prerequisite for the AI-pipeline **embedding processor** (the vector destinations + chunking processor don't depend on it and can proceed independently). Gates that impl on your sign-off.

## Risk tier

New security boundary → needs sign-off before implementation.

### DIFF
--- a/docs/design-documents/20260726-wasm-host-egress-capability.md
+++ b/docs/design-documents/20260726-wasm-host-egress-capability.md
@@ -27,14 +27,18 @@ allowlisted, host-executed HTTP call.
   host (full Go `net/http`) performs the I/O.
 - **Security boundary:** a host-side allowlist enforced in **two independent stages** — a coarse
   hostname/scheme pre-filter, then the **load-bearing resolved-IP check in the dialer's `Control`
-  hook on every dial** (the DNS-rebinding / TOCTOU defense). Private, link-local, loopback, and
-  reserved ranges are refused unless that exact IP is itself explicitly allowlisted (the local-Ollama
-  case). Redirects to non-allowlisted targets are not followed. Per-call timeout and response-size
-  cap are host-enforced, never guest-requestable.
-- **Configuration:** default **deny-all**. Egress is opt-in via a **host-reserved** config key,
-  clamped by an **engine-level ceiling** the Conduit operator sets — the guest can never widen its
-  own allowlist, because the resolved policy is bound host-side into the per-processor host-module
-  instance at `newWASMProcessor` time and is never mutable from guest code.
+  hook on every dial, for http and TLS alike** (the DNS-rebinding / TOCTOU defense). Private,
+  link-local, loopback, reserved, and embedded-v4 ranges (v4-mapped **and** NAT64 `64:ff9b::/96`) are
+  refused unless that exact **(IP, port)** pair is explicitly allowlisted (the local-Ollama case). The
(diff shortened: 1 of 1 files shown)


devarismeroxa:
**Sign-off — DeVaris, 2026-07-26.** Approved. WASM host-egress capability design (AI-pipeline prerequisite). Security-reviewed as an adversary: core resolved-IP dial-time gating confirmed sound; 6 must-fixes folded (incl. the proxy-env complete-bypass); API keys host-injected from the secret store (mandatory, never in the sandbox); per-processor policy isolation a hard tested requirement. Unblocks the embedding-processor impl. Merging.

PR #2698, opened 2026-07-26
test: de-flake Property-4 halt watermark + lifecycle FailureEvent under CI load

devarismeroxa:
## What

De-flakes two timing-sensitive tests that pass reliably locally (20–30× under `-race`) but flake **only** under the full `-race` test suite's CPU contention in CI — intermittently reddening the `test` job on unrelated PRs (surfaced on #2697).

## The flakes

**1. `tests/chaos` Property-4 halt (`TestSchemaDrift_TransportHalf_Halt_PositionNeverAdvances`) — I introduced this flake in #2694; owning + fixing it.** The fatal-error halt path (`waitHalted`) returns as soon as `Do` propagates the error, but the funnel's acks for records handled *before* the halt — and the upstream commits they drive — land **asynchronously**. Reading the watermark the instant `waitHalted` returned raced that pipeline (`committed==1` vs expected `2`), and a commit still in flight when the test returned could even race `t.TempDir` cleanup (`rename ... no such file`). The DLQ-routing sibling doesn't flake because `stopGracefully` drains first.
Fix: replace the instant reads with `waitUpstreamCommitted`/`waitPersistedPosition` polls (mirroring the existing `waitDelivered`). They fail loudly if the watermark ever creeps **past** the expected value — the actual invariant-6 violation this test guards, so the assertion is *strengthened*, not weakened.

**2. `pkg/lifecycle` `TestServiceLifecycle_PipelineError` (pre-existing).** The `OnFailure` event — emitted as part of the degradation `WaitPipeline` already observed, but delivered on a separate goroutine — was awaited with a **200ms** timeout, too tight under CI load. Bumped to 10s; the event still must arrive, so the assertion is unchanged.

## Why this matters

The Property-4 flake is in the **DBZ-2 suite that's supposed to be a *reliable* gate** — a flaky test there undermines the whole point. This is the third flaky-test incident this session (the `pkg/connector` teardown tests were fixed in #2695); the pattern is timing-sensitive assertions that only break under CI contention.

## Verification

`go build`/`go vet` clean; Property-4 tests **20 passed** under `-race -count=10`, lifecycle **10 passed** under `-race -count=10`; `golangci-lint --new-from-rev=origin/main` → 0 issues. Test-only; no production code.

## Risk tier

Tier 3 (test-only), but hardens the DBZ-2 correctness gate's reliability. Unblocks #2697 (whose `test` job these were flaking).

### DIFF
--- a/pkg/lifecycle/service_test.go
+++ b/pkg/lifecycle/service_test.go
@@ -343,7 +343,13 @@ func TestServiceLifecycle_PipelineError(t *testing.T) {
 	// pipeline errors contain only string messages, so we can only compare the errors by the messages
 	t.Log(pl.Error)
 
-	event, eventReceived, err := cchan.Chan[FailureEvent](events).RecvTimeout(ctx, 200*time.Millisecond)
+	// The OnFailure event is emitted as part of the degradation that WaitPipeline
+	// above already observed, so it arrives ~immediately - but it's delivered on a
+	// separate goroutine, so a too-tight timeout flakes under CI load (the full
+	// -race suite saturating CPUs can delay the send past a sub-second window).
+	// A generous bound removes the flake without weakening the assertion: the
+	// event still MUST arrive.
+	event, eventReceived, err := cchan.Chan[FailureEvent](events).RecvTimeout(ctx, 10*time.Second)
 	is.NoErr(err)
 	is.True(eventReceived)
 	is.Equal(pl.ID, event.ID)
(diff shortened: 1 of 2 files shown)


devarismeroxa:
**Sign-off — DeVaris, 2026-07-26** (approved the de-flake). Fixes the Property-4 halt flake I introduced in #2694 (poll for the watermark instead of an instant read; strengthened to fail if it creeps past the expected value) + the pre-existing pkg/lifecycle 200ms→10s FailureEvent timeout. Full suite green, no whack-a-mole. Merging.

PR #2720, opened 2026-07-31
feat(lifecycle-poc): drain-safe StopAndWait/ReconfigureProcessor for arch-v2

devarismeroxa:
## What

Plan 02 of the arch-v2 graduation epic: **drain-safe live reconfigure** for the arch-v2
(funnel) lifecycle. Replaces the two hard `CodeStopAndWaitUnsupported` refusals
(`StopAndWait`, `ReconfigureProcessor`) with a real, bounded, drain-then-durable-flush
implementation, so `provisioning.ApplyPlanLive` (HTTP apply + `--dev` hot-reload) works
under `Preview.PipelineArchV2` via graceful drain-and-restart.

Roadmap: v0.20 / Phase 2 arch-v2 graduation. Closes the "Open parity item" the refusal
doc-comment tracked. Design doc: `docs/design-documents/20260731-archv2-drain-reconfigure.md`.

## Risk tier: **Tier 1 (data path — drain/ack/position on live reconfigure)**

Preview-gated (`Preview.PipelineArchV2`, off by default) → nil production blast radius.

### Open questions resolved (in-impl, proven)
- **O1** — `ReconfigureProcessor` returns the v1 `ErrProcessorNotLiveReconfigurable` sentinel (arch-v2 has no in-place hot-swap), so `applyInPlace` falls back to restart with no arch-v2 branch. Proven with a real (non-mocked) v2 service test.
- **O2** — `StopAndWait` bounds the whole `Stop → WaitPipeline → WaitPersisted` by `DefaultStopAndWaitTimeout` (30s) or a tighter caller ctx deadline. On timeout: coded `CodeStopAndWaitTimeout`, **force-kills nothing** — degrades to the same benign-duplicate-on-restart contract `Source.Teardown` already guarantees, never a gap.
- **O3 (the race the recovery port created)** — a deliberate `Stop(force=false)` that hits a transient error mid-drain would auto-restart the pipeline the operator just stopped. Fixed with a per-run `intentionalStop atomic.Bool`, set before `Worker.Stop`, checked in `runPipeline`'s cleanup switch (after `isGracefulShutdown`, before recovery `default`) → finalize `UserStopped`. **Rolled back if `Stop` errors** (worker stop flag never set → pipeline still running → a later unrelated transient error must stay recovery-eligible).
- **O4** — DLQ shares the connector `Persister`, so `WaitPersisted` already covers it. Audited, no gap.

## Failure-mode analysis

| Failure | Behavior | Guarantee |
| --- | --- | --- |
| Clean drain | tomb `ErrStillAlive` → UserStopped | inv 1/3 (processingLock quiescence + Teardown flush) |
| Transient error mid-drain | `intentionalStop` → UserStopped, no recovery | inv 3/7 (O3) |
| Wedged destination | O2 bound → `CodeStopAndWaitTimeout`, no force-kill | benign duplicate on restart, never a gap |
| Read-but-unprocessed batch at stop | thrown away un-acked → re-read on restart | benign duplicate, never a gap |
| `kill -9` mid-drain | existing SIGKILL chaos suite | no gap |

The funnel drain audit (design doc §3.1) establishes `processingLock` gives the same "no batch mid-flight at teardown" guarantee v1 gets from its msg tracker. **No data-loss gap found.**

## Known follow-up (flagged, not hidden)
`DefaultStopAndWaitTimeout` is a process-wide constant, not yet per-pipeline configurable — design doc Consequences. `waitBounded` does not cancel the wrapped `WaitPipeline`/`WaitPersisted` goroutine on timeout (mirrors `Persister.WaitPendingWritesContext`) — at most one leaked goroutine per genuinely-wedged-destination timeout, documented.

## Tests
- `TestServiceLifecycle_Stop_TransientErrorMidDrain_NoRecovery` (O3 crux) — **fails without the fix** (infinite restart loop under `InfiniteRetriesErrRecovery`), passes with it, `-race -count=20`.
- `TestApplyPlanLive_ArchV2_ProcessorUpdate_FallsBackToRestart` — real v2 service, restart fall-through.
- `StopAndWait` ordering / not-running / never-force-stops unit tests.
- Chaos: `TestArchV2_LiveApplyDrain_UnderLoad_NoGap` (AC-1/AC-3), `TestArchV2_LiveApplyDrain_StuckDestination_O2Bound` (AC-6) — `-race -count=10`.

## Guard removal
`CodeStopAndWaitUnsupported` deleted (no other references); `CodeStopAndWaitTimeout` registered; `llms-full.txt` regenerated via `make generate` (single-line diff, CI-pinned tools, no stray `_string.go`/mock churn).

## Adversarial self-review
- Verified the `intentionalStop` arm sits between `isGracefulShutdown` and recovery `default`, and that force-stop still routes to the fatal arm (unchanged).
- Verified O2 timeout tears down nothing → "safe to retry" claim holds; a drain-phase timeout still finalizes UserStopped when the wedged write eventually returns.
- Re-ran full build + race suites + O3 fail-without + chaos + lint from clean `origin/main` (a mid-run accidental `git checkout` of `service.go` was caught and the diff reconstructed + re-verified byte-for-diffstat).

### DIFF
--- a/docs/design-documents/20260731-archv2-drain-reconfigure.md
+++ b/docs/design-documents/20260731-archv2-drain-reconfigure.md
@@ -214,6 +214,15 @@ every connector it manages** — including the DLQ destination, which `lifecycle
 creates through the same `s.connectors.Create` (and thus the same persister) as every other
 connector. There is no DLQ-specific persistence path this misses. **No gap found.**
 
+The same fact has one benign side effect (raised by adversarial review, Finding 3): because
+`WaitPersisted` is a **process-global** barrier over the shared persister — not scoped to a single
+`pipelineID` — a `StopAndWait` can hit its persist-phase O2 timeout because of an _unrelated_
+pipeline's slow/stuck write, not its own. This only ever over-waits (it can never return early
+before this pipeline's writes are durable), and a persist-phase timeout fails safe: `ApplyPlanLive`
+aborts on the error and mutates nothing (invariant 1 preserved). It is a liveness/observability
+wrinkle, not a correctness one; per-pipeline persistence scoping is a possible future refinement,
+not required for this interim.
+
 ### §3.1 — the funnel drain audit
 
 This is the audit the original `CodeStopAndWaitUnsupported` guard was waiting on. It walks the same
--- a/docs/design-documents/20260731-archv2-drain-reconfigure.md
+++ b/docs/design-documents/20260731-archv2-drain-reconfigure.md
@@ -1,0 +1,361 @@
+# Drain-safe live reconfigure for arch-v2 (Preview.PipelineArchV2 / funnel lifecycle)
+
+## Summary
+
+Ports `StopAndWait` and `ReconfigureProcessor` from `pkg/lifecycle` (v1) to the experimental
+`pkg/lifecycle-poc` (arch-v2 / funnel) lifecycle service, removing the `CodeStopAndWaitUnsupported`
+refusal guard those two methods have carried since the `Preview.PipelineArchV2` flag was introduced.
+This closes the "Open parity item" flagged in
+[20260708-live-server-deploy-apply.md](20260708-live-server-deploy-apply.md): `ApplyPlanLive`
+(`provisioning.Service`) can now apply live changes to a running pipeline under arch-v2 exactly as it
+already does under v1, with the same drain/durability guarantee. **Tier 1** — this is a data-path
+change (stop/drain/restart of a running pipeline) plus a fix to the arch-v2 error-recovery loop the
+same PR train landed a few commits ago (`a61d4bc`, #2718).
+
+This PR also fixes a race the recovery port introduced (O3, below): a deliberate, operator-initiated
+`Stop(force=false)` that races a transient (non-fatal) error surfacing from the drain itself was
+being misclassified as a spontaneous failure and auto-restarted — restarting a pipeline the operator
+just told to stop. This is fixed generically, for any caller of `Stop`, not just `StopAndWait`.
+
+## Context
+
+`pkg/lifecycle-poc` is the experimental `funnel`-based pipeline runtime behind
+`Preview.PipelineArchV2` (`runtime.go:444`). Until this PR, its `StopAndWait` and
+`ReconfigureProcessor` unconditionally refused with `CodeStopAndWaitUnsupported`
+(`lifecycle-poc/service.go`, pre-this-PR): the funnel's `Stop`/`WaitPipeline`/`Persister` interaction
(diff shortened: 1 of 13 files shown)


devarismeroxa:
## Independent adversarial review → 1 HIGH bug found and fixed (pushed)

A separate fresh-context review of this Tier-1 change found a **confirmed HIGH** defect the self-review and the O3 test both missed:

**Finding 1 (HIGH) — `intentionalStop` rollback was too broad.** The rollback cleared the flag on *any* `rp.w.Stop` error. That is correct only for the lock-acquisition-timeout path (flag never armed). `funnel.Worker.Stop`'s **other** error path sets `w.stop.Store(true)` and *then* fails in `tearDownSource` (a wedged/dead source) — returning an error with the flag already armed. The unconditional rollback wrongly cleared it there, and the subsequent non-fatal `Worker.Close` teardown error then fell into the recovery arm and **auto-restarted a pipeline the operator just stopped** — exactly the O3 bug this field exists to prevent. The existing O3 test can't catch it (its `Stop` *succeeds*).

**Fix:** gate the rollback on `!rp.w.Stopping()` (new `funnel.Worker.Stopping()` accessor reporting `w.stop`), so the marker is cleared only when the worker genuinely never began stopping. New regression test `TestServiceLifecycle_Stop_SourceTeardownFails_NoRecovery` — **fails without the fix** (infinite recovery restart loop under `InfiniteRetriesErrRecovery`; verified by reverting the guard → hang), **passes with it** (`-race -count=5`). Added `DestinationPluginWithControlledBlock` / `SourcePluginWithTeardownError` mock helpers.

**Finding 2 (LOW) — fixed.** The `CodeStopAndWaitTimeout` message now differs by phase: `stop` → pipeline still running, cleanly retriable; `drain`/`persist` → already stopping, don't force-restart. (Errors are API.)

**Finding 3 (informational) — documented.** `WaitPersisted` is a process-global barrier: a `StopAndWait` can hit its persist-phase timeout because of an unrelated pipeline's stuck write. Over-waits, fails safe (ApplyPlanLive aborts, no mutation). Noted in the design doc.

**Verified CORRECT by the review (no change needed):** switch-arm ordering, no carry-over across restart, guard removal / catalog consistency, `waitBounded` leaked goroutine is resource-only (never mutates state; no restart to collide with), invariant 1 via `ApplyPlanLive` abort-on-error, the §3.1 drain audit, and the recovery-vs-StopAndWait race (at-least-once either way).

Also rebased onto current `main` and added `WaitPersisted()` to `#2719`'s `lcConnectorService` chaos double (the merge-preview compile failure that reddened golangci-lint on the first run). Full-repo `golangci-lint` now 0 issues.

PR #2721, opened 2026-08-01
feat(lifecycle-poc): M-destination fan-out with per-record ack accounting (3a)

devarismeroxa:
## What

Slice **3a** of the arch-v2 multi-connector epic: **M-destination fan-out with correct per-record
ack accounting** (`multiAckNacker`). One source, many destinations. This is the crux and the
highest data-loss-risk code in the arch-v2 graduation.

Scope is deliberately destinations-only — the multi-source guard stays in place (that's 3b).

Roadmap: v0.21 arch-v2 graduation. Design doc: `docs/design-documents/20260731-archv2-multiconnector.md`.
ADR: `docs/architecture-decision-records/20260731-archv2-fanout-ack-model.md`.

## The correctness bar

A per-*batch* counter (what the old stub sketched) is a **silent data-loss bug**: destinations
diverge per record — dest1 may ack records 0-4 while dest2 nacks record 3. `multiAckNacker`
therefore accounts **per original source position**:

1. **Ack-only-when-unanimous** (inv 1) — a position is acked to the source only when *all* M
   branches acked it.
2. **Nack-wins** (inv 3) — any branch nacking routes the position to the DLQ exactly once. A
   partially-written record (dest1 ok, dest2 failed) is **not** a success.
3. **Idempotent / race-free** — terminal transition fires the parent action exactly once; a late
   ack vote for a nack-terminal position is a no-op; out-of-order branch completion is safe.
4. **In-source-order release** (inv 4) — completed positions are released in ascending order,
   buffering early completers, because `Source.Ack` sets `State.Position = p[last]`. Ack runs are
   coalesced; nacks release one at a time (different branches nack under different task IDs, so
   coalescing would misattribute DLQ metadata).

## Three data-loss/liveness bugs found and fixed in this PR

**1. `Batch.clone()` shared `positions` with full capacity.** `SplitRecord` grows positions via
`append`, and Go reuses the backing array when spare capacity exists — which it always did after
slicing. Two fan-out branches from one `clone()` shared one growable array, so branch A's split
could silently overwrite branch B's positions while B read them concurrently. Blast radius: a
wrong position reaching `Source.Ack` (invariant 2). Fixed with `slices.Clip`; regression test
verified to fail without it.

**2. `Batch.clone()` shared the `splitRecords` map by reference.** Two branches splitting
concurrently is a *fatal* concurrent map write, not a benign race. Fixed by per-branch copy.

**3. Duplicate source positions silently stalled the entire batch** (found during review, fixed in
`63bddeb`). `posIndex` maps position bytes to a slot; two records with identical positions
(including two empty/nil ones) made the later entry shadow the earlier. The shadowed slot never
accumulated votes, never became terminal, and the strictly-in-order release stopped there forever.
Probe before the fix: 3 positions where #0 and #2 collide produced **zero** `parent.Ack` calls —
the whole batch never acked, silently, unboundedly, while the pipeline kept running. At-least-once
held (nothing acked ⇒ nothing lost; records replay on restart), but a silent liveness failure beats
no error only in the wrong direction. Now rejected at construction with a registered code
`pipeline.duplicate_source_position` + actionable suggestion.

Bugs 1 and 2 were latent in code disabled before this slice (the old stub hit `panic`/error first),
so nothing shipped was ever exposed — but they'd be the first thing a real multi-destination
pipeline with a splitting processor hit.

## Failure-mode analysis

| Failure | Behavior | Guarantee |
| --- | --- | --- |
| One destination nacks a record | that position → DLQ once; others' copies do not spuriously ack it | inv 1/3 |
| Branch completes out of order | buffered; released in source order | inv 4 |
| Fatal DLQ error in a branch | propagates; branch pool errors; worker tombs | matches single-dest path |
| `kill -9` mid-fan-out (written to a subset) | source position never advanced ⇒ replay to **both** destinations on restart | inv 2/3: duplicates allowed, gaps forbidden |
| Duplicate/empty positions in a batch | coded error, pipeline fails fast | no silent stall |

**No durable multi-ack state exists** — the tally is in-memory and rebuilt from replay, so there is
nothing to migrate and no serialized-format change.

## Tests
- `multiAckNacker` unit suite: unanimity, nack-wins in any interleaving, partial per-record
  divergence, out-of-order completion, in-source-order release, fatal DLQ propagation, duplicate/
  empty-position rejection.
- **OQ-1 property test** (200 seeded iterations, differential per-branch splits + random nacks):
  every source record is acked **exactly once XOR** DLQ'd once — never both, never neither.
- `Batch.clone()` positions-aliasing + splitRecords regression tests.
- Slow-destination: source position must not advance past a withheld record.
- Integration: 1 source → 2 file sinks.
- **Chaos** `TestSIGKILL_MidFanout_GaplessResume`: real subprocess `kill -9` mid-fan-out, both
  destinations' fsync'd ledgers asserted gapless. `-race -count=10`, no flakes.

## Verification
`go build ./...` clean · `go test -race ./pkg/lifecycle-poc/... ./tests/chaos/...` green (chaos 108s)
· `golangci-lint` 0 issues · `make generate` clean (new error code in `llms-full.txt`).

## Adversarial self-review
Reviewed the fan-out call site, the release logic, and the lock discipline directly rather than
trusting the tests. The duplicate-position stall (bug 3) was found that way — by asking what
`posIndex` does when two keys collide, then proving it with a probe before fixing it.

### DIFF
--- a/cmd/conduit/root/run/run_test.go
+++ b/cmd/conduit/root/run/run_test.go
@@ -55,7 +55,7 @@ func TestRunCommandFlags(t *testing.T) {
 		{longName: "pipelines.error-recovery.max-retries-window", usage: "amount of time running without any errors after which a pipeline is considered healthy"},
 		{longName: "schema-registry.type", usage: "schema registry type; accepts builtin,confluent"},
 		{longName: "schema-registry.confluent.connection-string", usage: "confluent schema registry connection string"},
-		{longName: "preview.pipeline-arch-v2", usage: "enables experimental pipeline architecture v2 (note that the new architecture currently supports only 1 source and 1 destination per pipeline)"},
+		{longName: "preview.pipeline-arch-v2", usage: "enables experimental pipeline architecture v2 (note that the new architecture currently supports only 1 source, but multiple destinations, per pipeline)"},
 		{longName: "preview.pipeline-arch-v2-disable-metrics", usage: "disables metrics about amount of data (in bytes) moved in pipeline architecture v2 (increases performance)"},
 		{longName: "dev.cpuprofile", usage: "write CPU profile to file"},
 		{longName: "dev.memprofile", usage: "write memory profile to file"},
--- a/docs/architecture-decision-records/20260731-archv2-fanout-ack-model.md
+++ b/docs/architecture-decision-records/20260731-archv2-fanout-ack-model.md
@@ -1,0 +1,126 @@
+# arch-v2 destination fan-out: per-record ack tally, not per-batch
+
+## Summary
+
+For `funnel.Worker`'s M-destination fan-out (slice 3a of the arch-v2 multi-connector epic),
+`multiAckNacker` tracks ack/nack votes **per original source position**, acks the source only once
+**every** destination branch has acked a position (unanimity), routes a position to the DLQ exactly
+once the instant **any** branch nacks it (nack wins), and releases resolved positions to the source
+strictly in ascending source order. A per-batch counter — the shape the pre-existing (disabled) sketch
+code implied — is rejected as a data-loss bug: destinations diverge per record, not per batch, and a
+batch-level counter cannot represent that divergence at all.
+
+## Context
+
+`pkg/lifecycle-poc/funnel/worker.go` had a `multiAckNacker` type since the funnel architecture's
+introduction (#1913), but it was never completed: `Ack`/`Nack` both `panic("not implemented")`, and it
+was constructed as `newMultiAckNacker(parent, count)` — a single `*atomic.Int32` counter, decremented
+once per branch's call, intended to fire the parent action once it reached zero. The `doNextTask`
+caller that would have used it was itself disabled (`"multiple next tasks not supported yet"`).
+
+Bringing this online for slice 3a required deciding how to represent "M destinations, each
+independently voting ack or nack, for a batch of N records" — the counter's granularity (per batch,
+per record, or something else) determines whether the implementation can even express the scenarios
+a real fan-out produces.
+
+## Decision
+
+Track ack/nack state **per original source position**, not per batch. `multiAckNacker` holds parallel
+slices (indexed by position, captured once at fan-out time from `b.originalBatch()`): an ack-vote
+count, a terminal flag, and — for a nack — which branch's record/error/task-ID to use. `Ack`/`Nack`
+each update every position in the batch they're handed (after collapsing splits back to root positions
+via `originalBatch()`), then attempt to release a leading, contiguous run of now-terminal positions to
+the parent, in source order.
+
+Unanimity (ack) and nack-wins are enforced directly by this per-position structure: a position is
+`acked[i]` only when its vote count reaches the branch count; it is `nacked` immediately on the first
+nack vote, and a later vote (of either kind) for an already-terminal position is a no-op.
+
+## Alternatives considered
+
+### A per-batch counter (the pre-existing sketch's implied shape)
+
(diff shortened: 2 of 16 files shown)


devarismeroxa:
## Independent adversarial review → 1 HIGH bug in this PR fixed; 2 pre-existing bugs filed

A separate fresh-context review traced the two things I flagged as highest-risk and **cleared both**: a single branch cannot vote twice (the tainted loop partitions indices disjointly, including through the `Retry` recursion), and the lock order is consistently `multiAck.mu → DLQ.m` with no re-entry, so no deadlock. It also confirmed the `clone()` field audit is complete — every field is either deep-copied or a value copy, nothing mutable remains shared.

It then found what the tests didn't.

### F2 (HIGH, in this PR's own guard) — FIXED in `ce9045b`

My duplicate-position guard keys on `string(p)`, so **two** nils collide and were rejected — but a **single** nil sailed straight through. That's reachable: `SplitRecord` gives every piece after the first a nil position, so a sub-batch covering only the tail of a split run collapses to exactly `[nil]`. It reached `Source.Ack`, which does `State.Position = p[len(p)-1]` unconditionally — **overwriting the durable source position with nothing**. On restart: full re-snapshot for Postgres, offset 0 for file/Kafka. Invariant 2, and a far worse blast radius than the stall the guard was written to close.

Fixed at the **corruption site** (`validateAckPositions` in `Worker.Ack`), so the protection also covers the M=1 path where the same `[nil]` sub-batch would wipe the position with no fan-out involved. Plus explicit rejection in `newMultiAckNacker`. New code `pipeline.empty_source_position`. Regression tests: `TestMultiAckNacker_SingleEmptyPosition_Rejected`, `TestValidateAckPositions_RejectsEmpty`.

### F4 (MEDIUM) — FIXED in the same commit
My duplicate error said "this is a bug in the source connector". Nil positions come from `SplitRecord` — the *processor chain* — so an operator would have gone and audited the wrong component. Errors are API; both messages now attribute correctly.

### F1 (HIGH, pre-existing root cause) — filed as #2723, ADR narrowed in `3f45c84`
`Nack` propagates across a split run; `Retry` and `Filter` do not. So a sub-batch can cover only the *head* of a split run and ack the original position before the tail is delivered. The nil-tail half is now contained by `validateAckPositions` (fails loud instead of corrupting); the head case remains open. The ADR claimed this composition was sound — **narrowed to uniform-flag runs with the gap named**, rather than left asserting more than was proven.

### F3 (HIGH, pre-existing on `main`) — filed as #2722
`idx += len(subBatch.positions)` overshoots when a retried sub-batch grows via `SplitRecord`, skipping records **and then advancing the persisted position past them**. Verified: `p2`/`p3` never delivered, `State.Position` advanced to `p5`. Silent permanent loss, invariants 2/3. Not this PR's to fix, but a graduation blocker.

Worth noting: the fan-out path is *immune* to F3 — skipped positions never go terminal, so the strictly-in-order release stalls at the gap rather than releasing past it. That's an accidental argument for this ADR's in-order design.

### Chaos-coverage claim, corrected
The review is right that the PR body oversold it. `fanout_child.go` uses synthetic destinations, no processors, no nacks — each destination acks the whole batch in one call, so the per-position tally degenerates to the per-batch counter the ADR rejects. What `TestSIGKILL_MidFanout_GaplessResume` genuinely proves is **the unanimity gate under crash** (a record in destA's ledger but not destB's is never source-acked) — real and valuable. It does **not** prove per-record divergence, the nack/DLQ path, or split composition under crash. Those belong in 3c's chaos work.

Issue #2723, opened 2026-08-01
Split-run head can be acked before its tail is delivered (Retry/Filter don't propagate across a split run)


devarismeroxa:
## Summary

A processor that splits a record produces a *split run* whose pieces share one original position. `Batch.Nack` propagates across such a run (`setFlagWithErr`, `batch.go:280-292`), but **`Retry` and `Filter` do not**. A sub-batch can therefore cover only the *head* of a split run and vote ack for the original position **before the tail has been delivered anywhere**.

Found by adversarial review of #2721.

## Reproduction (verified)

Processor A splits `p0` into 3; a later per-destination processor B returns only 1 of the 3 (`ProcessorTask.Do` emits `Retry` whenever a processor returns fewer records than it received — `processor.go:110-113, 172-175`, documented behavior):

```
positions: [p0, nil, nil, p1, p2]
statuses:  [ack, retry, retry, ack, ack]

sub(0,1).originalBatch(): positions=[p0]        <-- votes ACK for p0
sub(1,3).originalBatch(): positions=[<nil>,<nil>]
ACK calls in order: [[p0] [<NIL> <NIL>] [p1 p2]]
```

`p0` is acked while pieces 1 and 2 have gone nowhere. Under fan-out, if the sibling branch also acks `p0` (it will — it took the straight-through path), unanimity fires and `Source.Ack` persists `p0`. The branch then errors on the nil tail and the pipeline tombs, but the position is already durable: on restart the source resumes past `p0` and the two derived records are gone.

**Invariants 1 and 3 violated.**

## Status / mitigation

The root cause is pre-existing (the same premature `Source.Ack` exists at M=1). Two partial mitigations are already in place from #2721:

- `validateAckPositions` (in `Worker.Ack`) now refuses to ack an empty position, so the *nil-tail* half of this shape fails loud instead of wiping the source position.
- That converts the silent-corruption case into a stopped pipeline with a coded error.

What remains unfixed is the **head** case: `p0` itself can still be acked before its tail is delivered.

## Fix direction

Make `Retry` and `Filter` propagate across a split run the way `Nack` already does, so a split run is always flagged uniformly and can never be partitioned mid-run. Needs its own design note — the interaction with `subBatchByFlag`'s partitioning is the tricky part.

Ships with a regression test: a split run broken by a partial-output processor, asserting the original position is not acked until every piece is durably handled.

## Doc correction needed

[...]

PR #2725, opened 2026-08-01
fix(funnel): propagate Retry/Filter across split runs (split-run head acked before its tail)

devarismeroxa:
## What

Fixes **#2723** — a split-run's head could be acked to the source **before its tail was delivered anywhere** (invariants 1 and 3).

## The bug

A processor that splits a record produces a *split run*: N batch entries belonging to ONE original source position. `Batch.Nack` propagated its flag across a whole run; **`Retry` and `Filter` did not**. `ProcessorTask.Do` emits `Retry` whenever a processor returns fewer records than it received — documented, real behavior.

So `doTask`'s tainted loop could build a sub-batch covering only the run's **head**, which collapses via `originalBatch()` to the original position and gets acked while the rest of the run has gone nowhere. Under fan-out both branches ack it, unanimity fires, `Source.Ack` persists it — then the pipeline errors on the tail, but the position is already durable. On restart the source resumes past it and the derived records are gone.

## Fix — flag escalation across split runs

A tier ranking now governs all flag writes: `Nack`(2) > `Retry`(1) > `Ack`/`Filter`(0). A write may only **escalate** a record's tier, never downgrade it. `Retry`/`Filter` route through `setFlagEscalating` → `escalateFlag`, which applies the write and propagates it to every other piece of the same split run — but only where it is a strict upgrade for that sibling.

**`Ack` and `Filter` deliberately share tier 0**: `subBatchByFlag` already groups them into one sub-batch, so that pair can never partition a run. The asymmetry also matters for correctness — it stops `Filter` from clobbering a legitimate `Ack` sibling it didn't choose (which would silently drop real processor output), while still letting `Retry`/`Nack` force a lagging sibling to wait for the rest of the run.

### Latent bug fixed on the way
The pre-existing `Nack` propagation never adjusted `filterCount` when overwriting a `Filter`-flagged sibling, so `HasActiveRecords`/`ActiveRecords` could go stale. Centralizing every write through `setRecordFlag` fixes that for `Nack` too.

### Defensive backstop
`validateSplitRunBoundary` runs at the actual partition decision point (`subBatchByFlag`, before `sub()`) and raises the coded `pipeline.split_run_partitioned` if a boundary would ever cut a run. Propagation should make this unreachable; it exists because a future flag-setting path that forgets to route through the escalating helpers would otherwise silently reintroduce this bug. Same trade as `CodeEmptySourcePosition`.

## Failure-mode analysis

| Failure | Before | After |
| --- | --- | --- |
| Processor returns fewer records than received, mid split run | head acked before tail delivered → records lost on restart | whole run escalates to `Retry`, stays intact |
| `Filter` on part of a split run | run partitioned | escalates across the run |
| `Nack` on part of a run | already propagated (correct) — but `filterCount` went stale | propagated, count maintained |
| A future flag path bypasses the helpers | silent premature ack | loud coded error at the boundary |

## Verification

Fail-without/pass-with confirmed independently (I disabled only the `Retry` propagation and re-ran):

```text
err: split run spanning batch indices [0,2] was about to be partitioned by a
sub-batch boundary [0,1): this would ack or nack part of a split record before
the rest of it was durably handled
--- FAIL: TestSplitRun_RetryHead_NotAckedBeforeTailResolved
```

— the backstop catching it loudly with the primary fix removed, i.e. defence in depth actually works.

Tests: `Retry`-head and `Filter`-head shapes, a run never partitioned across sub-batches, fan-out composition with no premature unanimity, the forced-partition backstop, and the existing split/DLQ/nack behaviour unchanged.

`go build ./...` clean · `go test -race ./pkg/lifecycle-poc/...` green · `golangci-lint` 0 issues · `make generate` leaves the tree clean (new code in `llms.txt`/`llms-full.txt`).

Chaos: green on two clean runs. One earlier run failed on the known-flaky `/mid-snapshot` timing tests while three agents were running concurrently — reporting that rather than hiding it; CI is the arbiter.

## ADR
The "Known gap" bullet added to `20260731-archv2-fanout-ack-model.md` in the same release described exactly this bug and pointed here. Updated to record the fix and the mechanism. The Decision/Alternatives sections are untouched — this was a forward-looking caveat on an open bug, not a settled decision being reversed.

### DIFF
--- a/docs/architecture-decision-records/20260731-archv2-fanout-ack-model.md
+++ b/docs/architecture-decision-records/20260731-archv2-fanout-ack-model.md
@@ -111,13 +111,21 @@ if profiling ever shows otherwise.
   regresses to dropping that map, `multiAckNacker`'s correctness claim above no longer holds for
   pipelines with pre-fan-out splitting.
 
-  **Known gap (narrowing this claim):** convergence assumes a split run carries a _uniform_ status
-  flag. `Batch.Nack` propagates across a run, but `Retry` and `Filter` do not — so a sub-batch can
-  cover only the _head_ of a split run and vote ack for the original position before the tail has
-  been delivered. Tracked as issue #2723; the nil-tail half of that shape is already contained by
-  `validateAckPositions` (an empty position is refused rather than acked, which would otherwise
-  overwrite the durable source position), but the head case remains open. Do not read the
-  convergence claim as covering `Retry`/`Filter`-partitioned split runs.
+  **Known gap, closed:** convergence assumes a split run carries a _uniform_ status flag.
+  `Batch.Nack` propagated across a run from the start, but `Retry` and `Filter` did not — so a
+  sub-batch could cover only the _head_ of a split run and vote ack for the original position
+  before the tail had been delivered. Filed as issue #2723; the nil-tail half of that shape was
+  already contained by `validateAckPositions` (an empty position is refused rather than acked,
+  which would otherwise overwrite the durable source position), but the head case remained open.
+  Fixed by giving every flag that can leave a record "not yet done" (`Retry`, `Filter`, alongside
+  the pre-existing `Nack`) the same whole-run propagation, ranked so a stronger flag can never be
+  downgraded by a weaker one touching a different piece of the same run (`splitRunFlagTier` in
+  `pkg/lifecycle-poc/funnel/batch.go`) — a split run now always carries one uniform flag, so
+  `Worker.subBatchByFlag`'s same-flag grouping can no longer land on a partial run. A defensive,
+  coded backstop (`CodeSplitRunPartitioned`, `validateSplitRunBoundary` in worker.go) rejects a
+  partitioned boundary outright if this guarantee ever regresses, rather than silently allowing a
+  premature ack. The convergence claim above now covers `Retry`/`Filter`-partitioned split runs
+  too, not just `Nack`-partitioned ones.
 - **Extending to N sources is out of scope for this decision.** This ADR only covers 1 source, M
   destinations. A future N-source slice would need its own decision about how (or whether) acks
   compose across multiple `funnel.Worker` instances — not addressed here.
--- a/llms-full.txt
+++ b/llms-full.txt
@@ -614,7 +614,7 @@ Author: Meroxa, Inc.
 | `sdk.schema.extract.key.enabled` | bool | true | no | Whether to extract and decode the record key with a schema. |  |
 | `sdk.schema.extract.payload.enabled` | bool | true | no | Whether to extract and decode the record payload with a schema. |  |
 
-## 3. Error codes (90)
+## 3. Error codes (91)
 
 | Reason | gRPC status | Description |
 | --- | --- | --- |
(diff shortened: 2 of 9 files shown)


devarismeroxa:
## Adversarial review: REQUEST CHANGES — converting to draft, do not merge

An independent review found the PR's central claim ("a split run now always carries one uniform flag") is **false**, with a deterministic repro. Recording the findings here rather than quietly reworking.

### A — CRITICAL, confirmed. `SplitRecord` writes flags outside the tier model; #2723 is NOT closed
`batch.go:212-217`: `make([]RecordStatus, len(recs)-1)` zero value **is** `RecordFlagAck`. The comment justifying it ("the original record must have been acked, otherwise the processor wouldn't receive it") is **invalidated by this PR** — escalation can leave the original flagged `Retry`, and because `ProcessorTask.Do` marks ranges end→start, a `SplitRecord` at a *lower* index in the *same* pass then injects `Ack` pieces into an already-`Retry` run.

Result on this branch: the run goes mixed, `subBatchByFlag` proposes a partitioning boundary, and `CodeSplitRunPartitioned` **fires fatally**. Result on `main` with identical input: the original data-loss shape of #2723 is still reachable. So this PR converts the silent bug into a fatal stop rather than closing it — and the error text says "internal invariant violation… please report it" while being triggered by ordinary processor behaviour (a processor that splits **and** doesn't consume the whole batch — i.e. the chunk→embed RAG shape under a rate limit).

Measured: **607/3000 (20.2%)** randomised split+partial-return pipelines hit the fatal guard.

### B — HIGH, confirmed. Escalation mutates `filterCount` mid-`Task.Do`, corrupting the active-index mapping
`ProcessorTask.Do` marks end→start precisely so the active-index mapping stays stable for lower-index calls still to come. Escalation can now *unfilter* a sibling and decrement `filterCount` mid-pass; when it reaches 0, `activeRecordIndices()` returns nil and subsequent calls use raw indices. Net effect: one chunk's content destroyed, another's duplicated, and a `Filter` decision silently lost. The reviewer's test passes on `main` and fails on this branch.

### C — HIGH, confirmed. `Filter` inside a `Retry` run is refused → non-convergent retry → unbounded recursion
Because marking runs end→start, the `Retry` range escalates the run to tier 1 first, and the processor's `Filter` at a lower index is then refused by the tier guard. The retry re-feeds identical input, the processor returns an identical result — fixpoint. `worker.go`'s retry branch is **direct self-recursion with no depth bound**, so this is not a hang but unbounded stack growth to Go's 1 GB limit: `fatal error: goroutine stack exceeds limit` kills the **whole Conduit process**, not just the pipeline.

This PR's own test can't catch it — its processor B is an identity transform, so the retry converges by accident.

### D — MEDIUM, confirmed. Quadratic escalation
`escalateFlag` does an O(run) lookup *plus* an O(run) sibling loop on **every** write. Measured on a single `Retry(0,n)`: 16k-piece run goes 10.5 µs → **439 ms** (~42,000×). A 16k-chunk document is a normal RAG batch.

### E / F — behaviour change (already-processed pieces re-fed to non-idempotent processors) undocumented; `Batch.Ack` is an unguarded hole with an unwritten caller contract.

### Confirmed sound (no action)
Ack/Filter grouping is unconditional — the tier hole is `SplitRecord`, not a flag pair. `filterCount` *arithmetic* is correct and genuinely fixes `main`'s pre-existing over-count. `validateSplitRunBoundary` has **zero** false positives across 3000 randomised pipelines — it is firing on a true positive. No interaction with `7af7a05` or `multiAckNacker`.

### Plan
Rework on the reviewer's recommended design, which fixes B, C and D together: propagate the split record's current status onto new pieces in `SplitRecord` (verified to take the guard-fire rate to 0/3000), and **normalise split-run flags in one linear pass after `Task.Do` returns** instead of escalating on every individual write. The unbounded retry recursion is a pre-existing `main` bug and gets filed separately.

**#2723 stays open.**

devarismeroxa:
Superseded by #2731 (merged as 391a184), which fixes #2723 by making the ack path run-aware instead of forcing split-run flags uniform. The four defects found in this PR's design are recorded above and each has a regression test in the merged change.

Issue #2726, opened 2026-08-01
funnel: unbounded self-recursion in doTask's retry branch crashes the whole process on a non-converging processor


devarismeroxa:
## Summary

`Worker.doTask`'s `RecordFlagRetry` branch calls **itself directly with no depth bound**. A processor that never converges — i.e. keeps returning fewer records than it receives, so the batch is re-marked `Retry` every pass — drives unbounded recursion.

This is not a hang. It is unbounded **stack** growth to Go's 1 GB limit, ending in `fatal error: goroutine stack exceeds limit`, which is **not recoverable**: it kills the entire Conduit process, not just the offending pipeline.

Found while adversarially reviewing #2725. **Pre-existing on `main`** and independent of that PR.

## Mechanism

`pkg/lifecycle-poc/funnel/worker.go`, the tainted-batch loop:

```go
case RecordFlagRetry:
    // Retry the sub-batch by passing it to the same task.
    subBatch.Ack(0, len(subBatch.records))
    err := w.doTask(ctx, taskNode, subBatch, acker)   // <-- direct self-recursion, unbounded
```

`ProcessorTask.Do` emits `Retry` whenever a processor returns fewer records than it received (documented behavior). If the retry pass feeds the processor identical input and the processor returns an identical short result, the loop is a fixpoint and recursion never terminates.

## Why it can happen in practice

Any processor whose short return is a deterministic function of its input reaches this fixpoint — e.g. a processor that refuses/skips records above a size threshold, or one whose rate-limit path returns only what it could process and is re-fed the same batch. It does not require a "buggy" processor, only a non-converging one.

## Impact

Whole-process crash. Every pipeline in the instance dies, not just the one with the bad processor. Under `Preview.PipelineArchV2` today; that engine is scheduled to become the default, so this is a graduation blocker.

## Fix direction

Two parts, both wanted:

1. **Bound the retry.** Convert the self-recursion into a loop with an explicit iteration cap, or carry a depth/attempt count.
2. **Detect no progress.** If a retry pass produces a sub-batch identical in size/shape to the previous one, that is a non-converging processor: fail the pipeline with a **registered, actionable error code** naming the offending task ID, rather than recursing. An unbounded retry on the data path should never be silent.

Ships with the regression test that fails without the fix (a deliberately non-converging processor, asserting a bounded coded error instead of a stack overflow).

PR #2727, opened 2026-08-01
fix(funnel): normalize split-run flags in one pass (supersedes #2725)

devarismeroxa:
## What

Fixes **#2723** — a split-run's head could be acked to the source **before its tail was delivered** (invariants 1 and 3).

**Supersedes #2725**, whose per-write escalation design was rejected by adversarial review with four confirmed defects. This is the rework on the design that review recommended. #2725 stays draft and should be closed once this lands.

## Design: normalise once, after `Do` — not on every write

1. **`SplitRecord` replicates the split record's real status** onto the new pieces. Previously it used `make([]RecordStatus, n)`, whose zero value is `RecordFlagAck` — so a `SplitRecord` at a lower index could inject `Ack` pieces into a run already flagged `Retry` (`ProcessorTask.Do` marks end→start). That was the critical defect in #2725: it left the run mixed and fired the guard on **20%** of randomised split+partial-return pipelines. The stale comment ("the original record must have been acked") is corrected.

2. **No flag mutation during `Task.Do`.** `Retry`/`Filter` are plain single-index writes again. After `Do` returns and before partitioning, `doTask` calls `normalizeSplitRuns()` — **one linear pass** that makes each run uniform by rank. This is what fixes the mid-`Do` `filterCount` mutation that corrupted the active-index mapping (#2725 defect B) and the O(n²) blowup (defect D).

3. **`Filter` is never escalated to `Retry`** (defect C). A run may hold both; `groupFlagAt` still treats it as one atomic span, and `resetForRetry` resets only `Retry`-flagged members, leaving `Filter` members excluded from `ActiveRecords`. **That is what makes the retry converge**: the processor's input actually shrinks instead of being re-fed identically. Asserted, not assumed — `TestSplitRun_FilterHead_NotAckedBeforeTailResolved` expects the retry call to receive *only* the undecided piece.

4. **Defensive backstop kept**, wording fixed. `validateSplitRunBoundary` + `CodeSplitRunPartitioned` were confirmed by review to have zero false positives. Its text no longer claims "internal invariant violation… please report it", which was wrong when ordinary processor behaviour triggers it.

A real gap was found during the rework: a naive "always swallow the whole run" grouping would make the backstop **unreachable even when it should fire**. `groupFlagAt` now also detects genuine non-uniformity within a run. Caught by the author's own `TestSubBatchByFlag_ForcedPartition_ReturnsCodedError`, which failed until that check was added.

## Failure-mode analysis

| Failure | Behavior |
| --- | --- |
| Processor returns fewer records mid split run | whole run normalised to `Retry`; head cannot be acked early |
| `Retry` then `SplitRecord` at a lower index in the same pass | pieces inherit the real status; run stays uniform; guard does **not** fire |
| `Filter` + `Retry` in one run | both preserved; retry input shrinks; converges |
| Split of a `Filter`-flagged record | `filterCount` accounted correctly (tested) |
| A future path leaves a run genuinely mixed | loud coded error at the partition point, not a silent premature ack |

## Verification

Three fail-without/pass-with reverts, each rerun explicitly:
- `SplitRecord` zero-value restored → `ack != nack` in `TestBatch_SplitRecord_ReplicatesNackStatus`, and the already-filtered guard stopped panicking.
- `Filter`-escalation reintroduced → gomock `Unexpected call to Process([...3 records...])` — proving the run-shrink (and therefore convergence) never happened.
- `normalizeSplitRuns` disabled → the backstop fired instead of a silent premature ack: `split run spanning batch indices [0,2] carries inconsistent flags`. Defence in depth working as intended.

**Property test: 3000 iterations of split+filter+retry+nack+partial-return — guard-fire 0/3000, mixed-run violations 0/3000** (#2725's design measured 607/3000).

**Benchmark, `normalizeSplitRuns`** (zero allocs): 1k → 3,602 ns · 4k → 11,577 ns · 16k → 56,290 ns. Roughly linear per 4× step; #2725's design went 10.5 µs → 439 ms at 16k.

`go build ./...` clean · `go test -race ./pkg/lifecycle-poc/...` green · full chaos suite green (109s) · `golangci-lint` 0 issues · `make generate` leaves the tree clean (90→91 error codes).

## Not fixed here
**#2726** (unbounded self-recursion in `doTask`'s retry branch — a non-converging processor grows the stack to Go's 1 GB limit and kills the whole process) is pre-existing on `main` and gets its own PR. This change does not make it more likely; the one scenario review named is proven to converge. General convergence is **not** claimed.

## ADR
The known-gap bullet in `20260731-archv2-fanout-ack-model.md` now describes the normalisation design and records the gap as closed.

### DIFF
--- a/docs/architecture-decision-records/20260731-archv2-fanout-ack-model.md
+++ b/docs/architecture-decision-records/20260731-archv2-fanout-ack-model.md
@@ -111,13 +111,28 @@ if profiling ever shows otherwise.
   regresses to dropping that map, `multiAckNacker`'s correctness claim above no longer holds for
   pipelines with pre-fan-out splitting.
 
-  **Known gap (narrowing this claim):** convergence assumes a split run carries a _uniform_ status
-  flag. `Batch.Nack` propagates across a run, but `Retry` and `Filter` do not — so a sub-batch can
-  cover only the _head_ of a split run and vote ack for the original position before the tail has
-  been delivered. Tracked as issue #2723; the nil-tail half of that shape is already contained by
-  `validateAckPositions` (an empty position is refused rather than acked, which would otherwise
-  overwrite the durable source position), but the head case remains open. Do not read the
-  convergence claim as covering `Retry`/`Filter`-partitioned split runs.
+  **Gap closed (was: narrowing this claim):** convergence used to assume a split run carries a
+  _uniform_ status flag by construction, which wasn't true — `Batch.Nack` propagated across a run,
+  but `Retry` and `Filter` did not, so a sub-batch could cover only the _head_ of a split run and
+  vote ack for the original position before the tail had been delivered (issue #2723).
+  `validateAckPositions` already contained the nil-tail half of that shape (an empty position is
+  refused rather than acked, which would otherwise overwrite the durable source position), but the
+  head case remained open through two review rounds (PR #2725, reverted after an adversarial
+  review found the fix's per-write flag-escalation mutated `filterCount` mid-`Task.Do` and could
+  corrupt content or hang on a non-converging processor).
+  The fix that closed it does not touch `Do` at all: `Batch.normalizeSplitRuns` runs once, after a
+  task's `Do` returns and before `Worker.doTask` partitions the batch, and resolves each split run
+  by rank (`Nack` > `Retry` > `Filter`/`Ack`). The one deliberate asymmetry — `Filter` is never
+  escalated to `Retry` even when the rest of its run is — is itself load-bearing for this ADR's
+  convergence claim: it is what lets a retry pass on a split run actually shrink (filtered pieces
+  stay excluded from `Batch.ActiveRecords`) instead of re-feeding the processor an identical-sized
+  input forever. `Worker.subBatchByFlag` (via `Batch.groupFlagAt`) treats a normalized run as one
+  indivisible unit when partitioning, so it can no longer be torn apart at all; a defensive
+  `CodeSplitRunPartitioned` guard covers the case where some future write path bypasses
+  normalization. This closes #2723 for `Retry`/`Filter`-partitioned split runs, with M=1 and M>1
+  (fan-out) both covered by test. It does **not** close #2726 (a non-converging processor's retry
+  recursion has no bound) — a separate, still-open concern this fix was required not to make any
+  more likely.
 - **Extending to N sources is out of scope for this decision.** This ADR only covers 1 source, M
   destinations. A future N-source slice would need its own decision about how (or whether) acks
   compose across multiple `funnel.Worker` instances — not addressed here.
--- a/llms-full.txt
+++ b/llms-full.txt
@@ -614,7 +614,7 @@ Author: Meroxa, Inc.
 | `sdk.schema.extract.key.enabled` | bool | true | no | Whether to extract and decode the record key with a schema. |  |
 | `sdk.schema.extract.payload.enabled` | bool | true | no | Whether to extract and decode the record payload with a schema. |  |
 
-## 3. Error codes (90)
+## 3. Error codes (91)
 
 | Reason | gRPC status | Description |
 | --- | --- | --- |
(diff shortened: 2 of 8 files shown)


devarismeroxa:
## Adversarial review: REQUEST CHANGES — converting to draft. Second attempt, second rejection.

Independent review found the rework closes #2725's findings **B** (mid-`Do` mutation) and **D** (quadratic) — both verified — and that `groupFlagAt` is sound with no false positives on the legal `Filter`+`Retry` pairing. But it introduces **two CRITICAL regressions against `main`**, and #2723 is still not closed.

All findings have runnable repros, each executed against **both** this branch and `origin/main` to separate regression from pre-existing.

### C1 — CRITICAL, confirmed, REGRESSION. Non-convergent retry → whole-process crash
`normalizeRun` escalates every non-`Filter` run member to `Retry`, so the retry input is the **whole run**, not the unprocessed tail. Any processor whose output size is a deterministic function of its input — *an embedding processor with a per-call batch cap, i.e. the exact shape this fix targets* — returns the same count every pass. With no `Filter` present there is no shrink mechanism at all. Fixpoint.

| | behaviour |
| --- | --- |
| `main` | call 1 input=3, call 2 input=**1**, terminates via `CodeEmptySourcePosition` (fail-loud, records replay) |
| this branch | call 1..N input=**3** forever |

`doTask`'s retry is unbounded self-recursion (#2726), so this is stack growth to Go's 1 GB limit → unrecoverable `fatal error`, killing **every pipeline in the process**. It also holds the first task's processing lock throughout, so SIGTERM never drains (invariant 7).

The ADR text this PR adds claims the `Filter` asymmetry is "what lets a retry pass actually shrink" and that the fix "was required not to make [#2726] any more likely". The repro falsifies both: a shape that **terminated on `main` now recurses forever.**

### C2 — CRITICAL, confirmed, REGRESSION. Retry re-feeds transformed records → silent double transformation
`SetRecords` has already written processor output into `b.records`; `resetForRetry` does not restore originals. Escalating an already-`Ack`'d piece to `Retry` hands the processor **its own output**.

| | destination receives |
| --- | --- |
| `main` | `[a\|X, b\|X, c\|X]`, then errors on the nil-position tail |
| this branch | `[a\|X\|X, b\|X\|X, c\|X]` — **source acked, `doTask` returns nil** |

Silently doubly-encoded records reach the destination and the position advances. Invariant 6, permanent and unrecoverable — worse than the data loss being fixed. For an embedding processor: embedding an already-embedded payload. There is no idempotency contract on processors anywhere in this codebase; introducing one is a public-contract change needing a design doc and sign-off, not a doc comment.

### H3 — HIGH, confirmed. #2723 is not closed: two disagreeing definitions of run membership
`splitRecords` is keyed on `b.records[i].Position` (the record's *current* position) while `groupFlagAt`/`checkEdge` look up `b.positions[i]` (the *original source* position). `Batch`'s own field doc says these are deliberately separate "because a processor tries to change the position." When a processor rewrites a position, the head is classified as an ordinary record — **both the grouping and the backstop miss it**, and the head is acked with zero destination writes. Same on `main`, so not a regression, but three doc comments in this PR assert the opposite.

### H4 — HIGH, confirmed. The backstop fires *after* earlier sub-batches were already acked
`codes.go` claims failure happens "**before that ack/nack call is ever made**." False: `subBatchByFlag` is called at the top of each loop iteration, and the previous iteration already ran `acker.Ack`/`Nack`. When the guard fires at sub-batch k>0, 0..k−1 are flushed and their positions advanced. Validation must cover the whole partitioning before the first flush.

### M5 / M6 / L7
M5: the new `setFlagWithErr` comment asserts safety that holds only for `ProcessorTask.Do`'s end→start marking — `DestinationTask.markBatchRecords` iterates **forward**, and the repro shows a destination-failed record acked as success and never DLQ'd. M6: the panic is genuinely unreachable (verified), but its justification cites a calling convention already violated two frames up — two pre-existing bugs found there (a legal `MultiRecord` output causing `index out of range`, and the wrong record being filtered). L7: this PR edits a merged ADR in place, which CLAUDE.md forbids — supersede instead.

### On the CI evidence in the PR body
The property test's 0/3000 is consistent with every finding above: C1/C2 need a deterministic output cap or a non-idempotent transform, H3 needs a position rewrite, M5 needs a destination ack error on a partially-filtered run. **None is in the generator's shape space.** Guard-fire count was never a proxy for these.

### Status
Not merging. #2723 stays open. The pre-existing bugs found here are being filed separately. The blocking question (C1/C2) is a design decision with a public-contract implication and is going to DeVaris rather than a third autonomous attempt.

devarismeroxa:
Superseded by #2731 (merged as 391a184). The two critical regressions found in this PR's normalize-once design — non-convergence for a deterministic output-cap processor, and re-feeding already-transformed records — are both now permanent regression tests in the merged change (TestRunLedger_OutputCapProcessor_ConvergesAndAcksOnce, TestRunLedger_NonIdempotentProcessor_TransformsExactlyOnce).

Issue #2730, opened 2026-08-01
funnel: split-run membership defined two different ways — position rewrite defeats both grouping and backstop


devarismeroxa:
## Summary

Two parts of the funnel disagree about what makes a record part of a split run, and the disagreement makes both the run-grouping **and** its defensive backstop miss real cases.

- `findSplitRecord`/`normalizeSplitRuns` define a run **structurally**: a maximal `[non-nil, nil, nil, …]` block of `b.positions`.
- `groupFlagAt` and `validateSplitRunBoundary.checkEdge` define it by **map lookup**: `positions[idx] == nil || positions[idx] ∈ splitRecords`.

These agree only while a run head's `b.positions[i]` equals its `splitRecords` key. But `SplitRecord` keys the map on `b.records[i].Position` — the record's **current** position — while every consumer looks up `b.positions[i]`, the **original source** position. `Batch`'s own field doc states the two are deliberately kept separate *"because a processor tries to change the position."*

Found while adversarially reviewing #2727; **pre-existing on `main`**. It is why #2723 is not fully closed.

## Reproduction

Processor A rewrites `p0`'s position → processor B splits it into 3 pieces → processor C filters the head piece and leaves the other two unprocessed:

```
SOURCE ACK: ["p0"]        <-- with ZERO destination writes
doTask err = split run spanning batch indices [0,2] carries inconsistent flags ...
```

Because the head's current position no longer matches the map key, `groupFlagAt` reports a span of 1 and `checkEdge` returns nil for it — so the head is treated as an ordinary record, acked to the source, while two of its three pieces were never delivered anywhere.

## Impact

Data loss (invariants 1 and 3) for any pipeline where a processor rewrites a position **before** a later processor splits that record. Note the ack is emitted *before* the guard fires (see the related ordering problem — a backstop that validates lazily inside the flush loop runs after earlier sub-batches have already been acked).

## Fix

Pick **one** definition of run membership, not both:

- key `splitRecords` on `b.positions[i]` (the original source position), **or**
- make `groupFlagAt`/`checkEdge` use `findSplitRecord`'s structural definition.

The first is a one-line change. Either way, the doc comments asserting a boundary "can never land strictly inside a run" and that two-edge checking "is sufficient to catch every possible partition" must be corrected or made true.

Ships with the position-rewrite regression test above.

PR #2731, opened 2026-08-01
fix(funnel): withhold a split run's source position until every member is terminal

### LINKED ISSUE #2723: Split-run head can be acked before its tail is delivered (Retry/Filter don't propagate across a split run)

devarismeroxa:
## What

Closes **#2723** (split-run head acked before its tail is delivered) and **#2730** (run-membership defined two ways).

**Third attempt.** #2725 (per-write flag escalation) and #2727 (post-`Do` run normalisation) were both rejected by adversarial review. DeVaris chose this design: **fix the ack path, not the flags.**

## Design

`Retry`/`Filter`/`Nack` flag semantics are **completely untouched** — that is the point. A new `runAckNacker` decorator intercepts ack/nack decisions and releases a split run's original source position only once **every** member of that run is terminal. This mirrors `multiAckNacker`'s per-position unanimity, already proven in this package.

- **"Terminal"** = acked (reached the true end), filtered, or nacked. `Retry` is explicitly **not** terminal — it is still in flight and may be split further.
- **Ledger lives on `Batch`** (`runs []*splitRun`, shared pointers) because `sub()` produces distinct `*Batch` structs across loop iterations *and* through the retry recursion; a shared pointer is what lets them observe one tally. The release *decision* lives in `runAckNacker`, fresh per acker scope: once per batch-read pass in `Worker.Do`, once per branch in `doNextTask`'s fan-out. A branch votes for a run's original position exactly once — which is precisely what `multiAckNacker`'s one-vote-per-branch assumption requires.
- **#2730 fixed as a prerequisite**: `splitRecords` is now keyed on `b.positions[i]` (original source position) instead of `b.records[i].Position` (current position), so a processor that rewrites a position can no longer detach a run head from its own run. The ledger would be unsound without this.

## Why the two previous designs were wrong

Both forced run flags uniform. That is what closed #2723 *and* what broke everything else:
- retry input never shrank → **non-convergent** for a processor with a deterministic output cap (an embedding processor with a per-call batch limit — the exact target shape) → unbounded recursion (#2726) → whole-process crash.
- already-transformed records were re-fed → **silent double transformation** reaching the destination.

Both failure shapes now have regression tests **in this PR**, and both pass:
- `TestRunLedger_OutputCapProcessor_ConvergesAndAcksOnce` — a 5-piece run through a cap-2 processor converges in exactly 3 rounds, identical to `main`.
- `TestRunLedger_NonIdempotentProcessor_TransformsExactlyOnce` — destination receives `"value|X"`, never `"value|X|X"`.

Neither is possible here because `runAckNacker` never calls `Retry`/`SetRecords` itself; it only intercepts decisions `doTask` already made.

## Failure-mode analysis

| Failure | Behavior |
| --- | --- |
| Sub-batch covers only a run's head | head's position withheld until the tail is terminal |
| Processor with deterministic output cap | converges exactly as on `main` (flags untouched) |
| Non-idempotent processor + retry | each record transformed exactly once |
| Processor rewrites a position, then a later processor splits it | run stays intact (#2730) |
| Run pending on one fan-out branch | no premature unanimity |
| Deferred run followed by later positions | later positions cannot ack past it |

**In-source-order release** is guaranteed structurally: `doTask`'s tainted loop and its retry recursion are sequential per branch, so a run's contiguous span resolves before the loop reaches anything after it. This is a **caller-provided** guarantee (unlike `multiAckNacker`'s self-enforced prefix scan) — flagged explicitly as a limit in the design doc rather than left implicit.

## Decisions taken
- `SplitRecord`'s zero-value `Ack` on new pieces is **kept** — harmless here because completion is tracked independently of flags. Proven by `TestBatch_SplitRecord_ZeroValueAck_SafeUnderRunLedger`, which reproduces #2725's exact "split while a sibling is Retry" shape.
- `validateSplitRunBoundary` / `CodeSplitRunPartitioned` from the rejected attempts are **omitted entirely**: under this design a partitioned run is legal and handled correctly, so the guard's premise no longer holds.

## Verification
Fail-without/pass-with: short-circuiting `runAckNacker.vote` to forward straight through fails every new test (`TestRunLedger_RetryHead…` → `2 != 1`, releasing the head separately from the tail; the fan-out test → `<nil> != [0]`, premature unanimity). Restoring returns the suite to green.

`go build ./...` clean · `go test -race ./pkg/lifecycle-poc/...` green · full chaos suite green (108s) · `golangci-lint` 0 issues · no error codes added, so no `make generate` needed. Property test over split+filter+retry+nack+position-rewrite passes.

## Not fixed here
**#2726** (unbounded retry recursion) — confirmed not fixed and **not made more likely**; nothing here escalates flags or alters retry input. Its own PR.

Design doc: `docs/design-documents/20260801-archv2-split-run-ack-ledger.md` (new — the merged ADR is **not** edited in place; #2727 was pinged for that).

### DIFF
--- a/docs/design-documents/20260801-archv2-split-run-ack-ledger.md
+++ b/docs/design-documents/20260801-archv2-split-run-ack-ledger.md
@@ -227,16 +227,34 @@ layer doesn't need them to.
   a call shape that cannot currently occur was judged speculative complexity against a
   non-existent failure mode (CLAUDE.md's "no speculative generality"); if a future caller changes
   that assumption, this section is the flag to revisit it.
-- **A split run whose disjoint pieces each independently reach the _same, repeated_ destination
-  fan-out point** (a _pre_-fan-out processor splits a run across flag groups such that different
-  groups reach `doNextTask`'s fan-out branch at different times) is not newly handled. It already
-  fails loud today, unrelated to this change: each such group builds its own `multiAckNacker` via
-  `b.originalBatch()`, and a pure-tail group's `originalBatch()` yields nil positions, which
-  `newMultiAckNacker` already rejects with `CodeEmptySourcePosition`. This fix does not touch that
-  path, so behavior there is unchanged: a safe, coded failure, not silent corruption — but also not
-  something this ledger makes work correctly. The scoped shape this design (and its tests) covers is
-  a run split **after** a (possibly repeated) fan-out point, independently per branch, which is the
-  shape #2723 was actually filed against.
+- **A split run cut across a fan-out boundary is explicitly refused, with a coded error.** If a
+  processor _before_ a multi-destination fan-out splits a record and a later pre-fan-out processor
+  resolves only part of that run, the sub-batch reaching `doNextTask` holds only some of the run's
+  members. `Batch.clone` must give each branch its own ledger (branches diverge, and a shared tally
+  would race and bleed decisions across branches), but a clone carries `splitRun.total` — the
+  whole-run member count. The branch could then never complete the run, and neither could the parent
+  side: two disjoint tallies, no join point. `validateRunsWholeBeforeFanOut` detects this before the
+  fan-out and returns `CodeSplitRunStraddlesFanOut`. Nothing is acked, so the records are redelivered
+  on restart (invariant 3).
+
+  **An earlier draft of this document claimed this shape "already fails loud today … this fix does
+  not touch that path." Both halves were wrong**, and adversarial review caught it. The tail group
+  does not always reach `newMultiAckNacker` at all — `doTask` routes a `RecordFlagNack` group
+  straight to `acker.Nack`, and a group with no active records straight to `acker.Ack`, bypassing
+  `doNextTask` entirely. And the ledger _does_ touch that path: wrapping `Worker.Do` in a
+  `runAckNacker` intercepts the nil-position batch and withholds it, so `Worker.Ack`'s
+  `validateAckPositions` never runs either. Withholding silently would therefore have **disarmed both
+  of the backstops that caught this shape before the ledger existed**, turning a loud failure into a
+  permanently withheld position while later positions acked past it — silently skipping the record on
+  restart. The explicit refusal restores the loud failure.
+
+  Joining a run's tally _across_ fan-out branches is possible (hoist the ledger into a per-pass
+  registry, or have each branch complete against the member count it actually received rather than
+  `run.total`). It is deliberately not attempted without a real use case for the shape.
+
+- The shape this ledger positively supports is a run split **after** a fan-out point, independently
+  per branch — such a run is created wholly inside one branch's ledger, so its total and its members
+  agree. That is the shape #2723 was filed against.
 
 ## Testing
 
--- a/docs/design-documents/20260801-archv2-split-run-ack-ledger.md
+++ b/docs/design-documents/20260801-archv2-split-run-ack-ledger.md
@@ -1,0 +1,276 @@
+# Split-run ack ledger: fixing #2723 without touching RecordFlag semantics
+
+## Summary
+
+Fixes **#2723** (a split run's head can be acked to the source before its tail is delivered) and
+**#2730** (split-run membership is defined two different ways, and a position rewrite defeats
+both). This is the **third** attempt: #2725 (per-write flag escalation) and #2727 (normalize-once
+after `Task.Do`) were both rejected by adversarial review, for reasons this design deliberately
+avoids by construction rather than by patching around them — see
+[Why the two prior attempts failed](#why-the-two-prior-attempts-failed).
+
+The fix does not change `RecordFlag` semantics at all. `Retry`, `Filter`, and `Nack` behave in
+`pkg/lifecycle-poc/funnel` exactly as they do on `main`. Instead, a new decorator —
+`runAckNacker`, wrapping the existing `ackNacker` interface (`worker.go`) — sits between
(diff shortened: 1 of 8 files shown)


devarismeroxa:
## Adversarial review round 3 → 1 CRITICAL found and fixed (`c2e071f`)

The review confirmed the **single-destination half genuinely closes #2723** and that #2730 is completely fixed — but found a CRITICAL regression in the fan-out half. Fixed rather than reworked, since the defect was localized.

### The bug
`Batch.clone` must give each fan-out branch its own ledger (branches diverge; a shared tally would race and bleed decisions across branches), but the clone carries `splitRun.total` — the **whole-run** member count. A sub-batch reaching a fan-out may hold only one flag group's members, so the branch needs `total` to complete while seeing only k < total, and the parent side only ever sees the other total−k. Two disjoint tallies, no join point.

Consequence: the run's original position withheld **forever** while every later position acked normally. `Source.Ack` sets `State.Position` from the last position handed to it, so the withheld record was **silently skipped on restart** — no error, no DLQ entry, no replay.

Worse, it **disarmed both backstops that caught this shape before the ledger existed**: `doTask` routes a `Nack` group straight to `acker.Nack` and an inactive group straight to `acker.Ack`, bypassing `doNextTask` (so `newMultiAckNacker`'s empty-position check never runs), and wrapping `Worker.Do` intercepts the nil-position batch before `validateAckPositions` sees it. Both previously failed loud. The review's repro showed `main` DLQ'ing the record and stopping loudly where this branch returned `nil` and acked past it.

### Fix
`validateRunsWholeBeforeFanOut` refuses the shape up front with `pipeline.split_run_straddles_fanout`. Nothing is acked, so records are redelivered on restart (invariant 3). Splitting **inside** a branch, after the clone, is unaffected and stays supported — such a run lives wholly in one branch's ledger, so its total and members agree.

Regression test `TestRunLedger_FanOut_RunCutBeforeFanOut_FailsLoud` — verified fail-without/pass-with. This is exactly the gap the review identified between the existing fan-out test (which splits *inside* the branch, after the clone) and reality.

### Also fixed from the same review
- `splitRun.complete` now returns an error on over-count instead of masking it with `>=`. An over-count releases the run **early** (invariant 1), and `released` only guards a second forward, not a premature first one.
- Corrected comments naming `runAckNacker` as the isolation boundary — it is stateless; the tally lives in `Batch.runs`. **That misattribution is what hid this bug**; the real boundary is `cloneRuns`.
- The fan-out comment now says a branch votes **at most** once, not exactly once — zero votes was the actual failure mode, and zero stalls `multiAckNacker`'s prefix scan for that position and every later one.
- Corrected the design doc's "Limits" section, which claimed this shape "already fails loud today" and that the fix "does not touch that path". Both were false, and the doc now says so explicitly rather than quietly rewriting history.

### Not adopted (deliberate)
The review offered two richer options — hoisting the ledger into a per-pass registry, or making `multiAckNacker` the join point. Both are real designs, but joining a run's tally across branches without a use case for the shape is speculative generality. Failing loud is the boring, honest option and restores `main`'s behaviour.

`go build` clean · `-race` green on `pkg/lifecycle-poc/...` · full chaos suite green (109s) · `golangci-lint` 0 issues · `make generate` clean (new error code committed).

Issue #2729, opened 2026-08-01
funnel: destination nack index shift acks a destination-failed record as success (silent data loss)


devarismeroxa:
## Summary

`DestinationTask.markBatchRecords` iterates **forward**, calling `b.Nack(from+i, …)` once per errored destination ack, using indices into an `ActiveRecords()` list captured before the loop. `Batch.Nack` propagates across a split run, which can overwrite a `Filter` sibling and so change the active set **mid-loop** — every later index then shifts.

Result: a record the destination explicitly **failed** is acked to the source as a success and never reaches the DLQ.

Found while adversarially reviewing #2727; **pre-existing on `main`**, byte-identical behavior there.

## Reproduction

Batch: a split run `[Ack, Filter, Ack]` (untainted, so never normalized) plus ordinary records `p1`, `p2`. Destination errors on split piece 0 and on `p1`:

```
DLQ WRITE:  pos="p0"  err=dest failed p1     <-- wrong error attached to the DLQ record
SOURCE ACK: ["p0"]
SOURCE ACK: ["p1" "p2"]                      <-- p1 acked as SUCCESS
doTask err = <nil>
```

`p1` failed at the destination. It is acked to the source, never written to the DLQ, and the pipeline reports no error. On restart the source resumes past it.

## Impact

**Silent data loss** (invariants 1 and 3): a record that was never durably written downstream is acknowledged upstream. Also mis-attributes the failure reason on the DLQ record that *is* written, which corrupts operator diagnosis of the original failure.

Currently `Preview.PipelineArchV2` only, but that engine is scheduled to become the default — graduation blocker.

## Fix direction

Same class as #2728: resolve all target indices before mutating, or iterate end→start so a propagation cannot shift indices still to be visited. Note the destination path additionally needs the *error* to stay bound to the record it belongs to.

Ships with the regression test that fails without the fix.

Issue #2728, opened 2026-08-01
funnel: MultiRecord marking loop iterates forward — crashes the process or filters the wrong record


devarismeroxa:
## Summary

`ProcessorTask.markBatchRecords`'s `sdk.MultiRecord` loop iterates **forward** while mixing `b.Filter(from+i)` (for an empty `MultiRecord`) and `b.SplitRecord(from+i, …)` in the same range. Both mutate the active-index mapping that the loop's own indices are derived from, so later iterations address the wrong record.

Two distinct failures, both from **legal** processor output. Found while adversarially reviewing #2727; **pre-existing on `main`** and unrelated to that PR.

## 1. Whole-process crash

Processor returns `[MultiRecord{}, MultiRecord{a,b}]`:

```
panic: runtime error: index out of range [1] with length 1
```

A panic on the data path is unrecoverable — it takes down the entire Conduit process, every pipeline in it, not just the offending one.

## 2. Silent wrong-record filter (data loss)

Processor returns `[MultiRecord{a,b}, MultiRecord{}]`. The record the processor intended to filter (`p1`) stays `Ack` and is delivered; a split piece is silently dropped in its place:

```
flags=[ack filter ack]  positions=["p0" nil "p1"]
```

The filter decision lands on the wrong record — one record is delivered that should not have been, another is dropped that should have been. Silent, no error.

## Why it matters beyond these two shapes

`Batch.SplitRecord` carries a comment stating it is only ever called with an index derived from `ActiveRecords`, and other code now relies on that convention (it is the stated basis for a precondition panic added in the split-run work). The convention is already violated here, two frames up. Whatever the fix, that assumption needs to be made true or the dependent code needs a different basis.

## Fix direction

Iterate the marking loop **end→start** (as the non-`MultiRecord` path already does, precisely so index mutations don't invalidate lower indices still to be processed), or resolve all target indices up front before mutating.

Ships with regression tests for both shapes above.

PR #2733, opened 2026-08-01
fix(funnel): mark records end→start and stop nacks overwriting split-run siblings

devarismeroxa:
## What

Closes **#2728** and **#2729**. Same class of bug — a marking loop iterating **forward** over pre-captured active indices while calling methods that mutate the index mapping — but #2729 turned out to have **two** root causes.

## #2728 — `ProcessorTask.markBatchRecords` (`MultiRecord` case)

The loop mixed `b.Filter(from+i)` and `b.SplitRecord(from+i, …)` in one forward pass. Two failures from **legal** processor output:
- `[MultiRecord{}, MultiRecord{a,b}]` → `panic: runtime error: index out of range [1] with length 1`. A panic on the data path kills the whole process.
- `[MultiRecord{a,b}, MultiRecord{}]` → **the wrong record is filtered**: the intended one stays `Ack` and is delivered, a split piece is silently dropped.

**Fix:** the end→start discipline `Do`'s outer loop already uses. Indices resolve against `activeRecordIndices()`, recomputed per call, so a mutation at *i* only invalidates indices **above** it — processing highest-first means every mutation touches only already-resolved entries.

## #2729 — `DestinationTask.markBatchRecords`, plus a deeper cause

Ordering fixed the same way (defence in depth), but tracing the corruption showed the real bug was in **`Batch.setFlagWithErr`**: a `Nack` propagated its flag *and error* across an entire split run.

That propagation predates `run_ledger.go` (merged in #2731) and is now both redundant and actively harmful:
- `originalBatch()` already drops every non-head member of a run (nil positions), so propagating to siblings **never affected what reached the parent** through that path.
- `runAckNacker`/`splitRun` is now the single place a run's aggregate disposition is decided — `nacked` is sticky, so a run whose head acked and whose tail nacked still resolves to Nack, per-member, with no flag mutation needed.
- It overwrote a sibling that had **already reached its own genuine outcome** — including a `Filter`, whose own doc says its status "can't be changed anymore" — attaching one record's failure reason to a different record's DLQ entry. And it did so via the very index arithmetic it was silently invalidating for later indices in the same call.

**Fix:** propagation removed entirely.

## Failure-mode analysis

| Failure | Before | After |
| --- | --- | --- |
| `[MultiRecord{}, MultiRecord{a,b}]` | process panic | marked correctly |
| `[MultiRecord{a,b}, MultiRecord{}]` | wrong record filtered, piece silently dropped | intended record filtered |
| Destination fails one piece of a split run | sibling's flag+error overwritten; a destination-**failed** record acked as success and never DLQ'd; wrong error on the DLQ entry | each record keeps its own outcome and its own error; the run's Nack verdict comes from `splitRun` |
| `Filter`'d sibling in a nacked run | silently overwritten (contract violation) | untouched |

## Verification — fail-without/pass-with, both issues

**#2728**, reverting only `processor.go`:
```text
panic: runtime error: index out of range [1] with length 1
processor_test.go:281: [{ack} {filter} {ack}] != [{ack} {ack} {filter}]
```
(second line matches the issue's reported `flags=[ack filter ack]`)

**#2729**, reverting `batch.go` + `destination.go`:
```text
run_ledger_test.go:1046: 1 != 2   (only 1 DLQ entry instead of 2 — p1's failure silently swallowed)
```
The two sub-causes were isolated: reverting **only** `batch.go` still fails three lower-level tests, proving the propagation bug is caught independently of the ordering fix.

## Accounting
`TestRunLedger_AccountingConsistent_ThroughMixedMarkingPass` inspects `filterCount`, `splitRun.total` and `terminalCount` through the exact mixed pass from the issue: `filterCount` stays 1 (Nack can no longer overwrite a `Filter` slot), the run reaches 3/3 with no over-count, and `nackErr` stays bound to the head's own error.

## Property test
`TestFullPipeline_Property_MixedProcessorOutputAndDestinationFailure` — real `ProcessorTask` + `DestinationTask`, 300 random iterations × 5 runs (1500 total) under `-race`, mixing `MultiRecord`/`SingleRecord`/`FilterRecord`/`ErrorRecord` with random per-piece destination failures. Green.

**Stated gap:** the property test cannot naturally produce a split run *with a Filter sibling* in one processor pass — the SDK types don't allow one call to both split and filter within a run; that needs two stages. That interaction is covered by three targeted tests instead, all verified fail-without/pass-with.

## Verification run
`go build ./...` clean · `go test -race ./pkg/lifecycle-poc/...` green · full chaos suite green (106s) · `golangci-lint` 0 issues · no error codes changed, so no `make generate` needed.

**Noted, not suppressed:** the pre-existing `TestRunLedger_FanOut_PendingRunOnOneBranch_NoPrematureUnanimity` failed once under heavy concurrent load and passed 3 full-package + 5 isolated reruns. Unmodified by this diff; flagging rather than hiding it.

### DIFF
--- a/pkg/lifecycle-poc/funnel/batch.go
+++ b/pkg/lifecycle-poc/funnel/batch.go
@@ -126,6 +126,31 @@ func (b *Batch) Ack(i int, j ...int) {
 // Note that the indices i and j point to the slice returned by ActiveRecords.
 // If the slice contains filtered records, the filtered records are skipped and
 // not marked as nacked.
+//
+// # #2729: nacking one piece of a split record does not propagate to siblings
+//
+// Nack only ever sets the flag and error of the exact record(s) at the given
+// index/indices - never any other member of a split run it may belong to.
+// This used to be different: setFlagWithErr propagated a Nack's flag and
+// error across an entire split run (see git history), on the theory that
+// Batch.originalBatch's collapse-to-head logic needed every member to agree
+// so the run's failure wasn't lost. That reasoning no longer holds now that
+// run_ledger.go exists: originalBatch already drops every non-head member of
+// a run (their position is nil), so propagating to them never affected what
+// reached the parent through that path in the first place - and
+// runAckNacker/splitRun (run_ledger.go) is the ONE place that decides a run's
+// aggregate disposition, independently of any single member's RecordFlag,
+// via its own terminalCount/nacked bookkeeping.
+//
+// Worse, the propagation was actively wrong: it overwrote a sibling's flag
+// and error even when that sibling had ALREADY reached its own, different,
+// genuine outcome earlier in the same pass (e.g. a Filter decision, which
+// Filter's own doc comment says can never change again once set; or an
+// independent successful destination ack for a different piece of the same
+// run) - misattributing one record's failure reason to a completely
+// different record's DLQ entry, and doing so via index arithmetic
+// (activeRecordIndices) that the propagation itself silently invalidated for
+// any later index still to be processed in the same call. See #2729.
 func (b *Batch) Nack(i int, errs ...error) {
 	b.setFlagWithErr(RecordFlagNack, i, errs)
 	b.tainted = true
@@ -364,9 +389,12 @@ func (b *Batch) setFlagNoErr(f RecordFlag, i int, j ...int) {
 
 // setFlagWithErr sets the flag for the record at index i to f, and sets the
 // error for the record to err. If multiple errors are provided, they are
-// assigned to the records starting at index i. If an error is assigned to a
-// split record, all records in the split record are marked with the same flag
-// and error.
+// assigned to the records starting at index i.
+//
+// This never touches any record other than the one(s) resolved from i/errs -
+// in particular it never reaches across a split run to a sibling record. See
+// Nack's doc comment (#2729) for why: a sibling's disposition is aggregated
+// by run_ledger.go, not by mutating its RecordStatus here.
 func (b *Batch) setFlagWithErr(f RecordFlag, i int, errs []error) {
 	// TODO: we should not have to recalculate the active record indices every time.
 	activeIndices := b.activeRecordIndices()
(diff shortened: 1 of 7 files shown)


devarismeroxa:
## Adversarial review → scope cut to the two `end→start` fixes (`cca5640`)

The review ran the decisive experiment this PR should have — testing each half of the fix independently:

| `batch.go` | `destination.go` | Result |
| --- | --- | --- |
| propagation **ON** (main) | end→start (PR) | **PASS** |
| propagation **OFF** (PR) | forward (main) | **PASS** |
| main | main | FAIL (the bug) |
| PR | PR | PASS |

**Either half alone fixes #2729.** The propagation removal was redundant — and not free.

### Why it's dropped
It breaks a multi-destination pipeline that works on `main`. Processor A splits a record into 3; processor B returns `ErrorRecord` for one piece and `SingleRecord` for the other two — legal, ordinary output, and precisely the RAG shape (chunker splits, embedder fails one chunk, two vector sinks). With propagation the run is uniformly `Nack` → one sub-batch → one DLQ entry. Without it the flags are `[Ack, Nack, Ack]`, `subBatchByFlag` partitions the run into three, and the first hits the fan-out where `validateRunsWholeBeforeFanOut` (added hours earlier in #2731) rejects it:

```text
cannot fan out a batch holding only 1 of the 3 records that record "0" was split into
```

Nothing is acked, so the batch is redelivered and fails identically — a deterministic, permanent failure. No data loss, but a working pipeline dies.

**The reasoning error:** the removal argued `originalBatch()` drops non-head members so propagation "never affected what reached the parent". True of `originalBatch` — I verified it — but it is not the only consumer of a member's `RecordFlag`. `subBatchByFlag` reads per-member flags to decide how the batch is **partitioned**, and the fan-out guard consumes that partition. Propagation was load-bearing for keeping a run whole across a flag partition, which is exactly the property #2731's guard was designed around.

The removal has a real case on its merits — it can overwrite a `Filter` sibling's own outcome, violating `Filter`'s documented immutability — but it needs its own PR with a design note on the `subBatchByFlag` × `validateRunsWholeBeforeFanOut` interaction, not a rider on an index-shift fix.

### Also addressed
- **Added the uncovered shape** (review finding 4): `TestProcessorTask_Do_MultiRecord_PreFilteredBatch_MarksCorrectRecords` — a batch that *already* has filtered records on entry. Both existing #2728 tests start from a fresh batch where active indices coincidentally equal physical ones, and the property test structurally cannot produce it. Verified fail-without: on `main` the wrong records are split (`[record2, record0]`).
- **Narrowed the `destination.go` comment** (finding 5): `Do` calls `markBatchRecords` once per ack chunk with increasing `from`, so the reversal protects indices within **one chunk**, not across the batch. The old wording claimed a blanket guarantee a future maintainer would have relied on.

### Review-confirmed sound (unchanged)
`end→start` correctness verified for both mutators (`Filter` and `SplitRecord` shift only indices *above* the mutation point, so the descending walk is safe with no off-by-one); DLQ write ordering and invariant 4 unaffected; no conflict with #2732.

`go build` clean · `-race` green on `pkg/lifecycle-poc/...` · full chaos suite green (106s) · `golangci-lint` 0 issues.

PR #2734, opened 2026-08-01
feat(lifecycle-poc): N-source topology with shared-sink teardown ordering (3b)

devarismeroxa:
## What

Slice **3b** of the arch-v2 multi-connector epic: **N sources, one `funnel.Worker` per source**, converging on a shared destination. Completes the capability half of arch-v2 graduation alongside 3a (M destinations, #2721).

Roadmap: v0.21 arch-v2 graduation. Design doc: `docs/design-documents/20260801-archv2-multiconnector-nsource.md`.

## Why cross-source ack contamination is structurally impossible

`funnel.Worker` was already per-source and already its own terminal acker (`w.Ack` → `w.Source.Ack`). `doTask`/`doNextTask` are methods on `*Worker`, and `acker` defaults to `w` itself — so even though N workers execute over identical shared `TaskNode` pointers, every call is bound to the goroutine's own worker. A worker can only ever ack its own source. That is enforced by construction, not by a runtime check.

## The Tier-1 hazard found while implementing (not papered over)

`DestinationTask.Do` writes a batch, then reads back exactly that many acks off the destination's **single gRPC stream**, matching each ack to the position it just sent **in strict send order**. With N independent worker goroutines converging on one shared destination, two batches can interleave: worker A writes, worker B writes, then A's `Ack` read consumes B's response — **the wrong record acked to the wrong source**. Concurrent `Send` on one stream is also unsafe at the gRPC layer regardless.

Fixed with a per-shared-root mutex (`TaskNode.MarkSharedBoundary`), acquired for the entire call so the whole shared subtree is serialized — matching v1's `FaninNode` single-consumer behaviour rather than adding a new bottleneck. `TestNSource_ConcurrentWorkers_SerializeSharedDestinationWrites` fails immediately if two goroutines ever hold it concurrently (deterministic, not `-race` luck).

## Shared-sink teardown ordering — the crux

`Worker.Close` used to tear down its whole task chain **including the destination**. With N workers sharing one, the first source to finish would close a destination its siblings are still writing to.

Now each worker's `Close` tears down only its own source and its own DLQ (the shared tail is excluded from its `Tasks()` walk via `MarkSharedBoundary`), and `funnel.Sink.Open`/`Close` run exactly once — `Close` strictly **after** `workersWg.Wait()`, so no worker can be mid-write. Verified by reading, and by `TestNSource_SourceFinishesGracefully_DoesNotTouchSharedSink`.

## Deadlock guard added during rebase (would not have failed — it would have hung)

Rebasing onto #2732 created a live hazard. #2726's fix split `doTask` into a wrapper plus `doTaskAttempt`, with the **retry recursion calling `doTaskAttempt` directly**. Put the shared mutex in `doTaskAttempt` and the first retry inside a shared node re-enters a non-reentrant mutex and **self-deadlocks** — the worker hangs forever holding the processing lock, so the pipeline never drains and SIGTERM never completes (invariant 7).

The lock therefore lives in the `doTask` wrapper, acquired once at the external entry and held across the whole synchronous pass. Nothing covered this interaction, so `TestNSource_SharedBoundary_RetryDoesNotSelfDeadlock` now does — verified by moving the lock into `doTaskAttempt`, which fails it by timeout with the explanation attached instead of silently hanging the suite.

## Failure-mode analysis

| Failure | Behavior |
| --- | --- |
| Two workers write the shared destination concurrently | serialized at the shared root; correct ack↔position pairing |
| A snapshot-only source finishes while another streams | pipeline stays Running; terminal only when all workers exit; shared sink **not** torn down early |
| Fatal error in one source | whole pipeline degrades (matches v1: one tomb fans a Kill to every worker's ctx) |
| Non-fatal error in one source | recovery path, pipeline-wide restart |
| Retry inside the shared subtree | completes; no mutex re-entry |
| `kill -9`, one fast + one slow source | each resumes independently and gaplessly; duplicates allowed |

**Status aggregation generalizes for free** because a worker only calls `rp.t.Kill` on an **actual** error — a source that exhausts gracefully (`io.EOF`) or is asked to Stop returns nil, so the tomb stays alive and siblings keep running. Concurrent kills are safe (tomb records only the first reason).

## Also in this PR
- **Per-source DLQ naming** (`pl.ID+"-"+sourceID+"-dlq"`) — N sources would collide on the old single name. `ProvisionTypeDLQ` is not persisted, so no state migration.
- **Multi-source guard removed.** It was an uncoded `cerrors.New` whose text told users to "disable the experimental feature flag" — advice that inverts once v2 is the default.
- **Flag usage text corrected** (it still claimed "only 1 source"), and the assertion test with it.

## Tests
Funnel-level: ack isolation, no cross-source contamination (idle source's position stays fixed while a sibling runs), serialized shared writes, graceful finish not touching the shared sink, the retry-deadlock guard. Chaos: `TestSIGKILL_NSource_FastAndSlow_GaplessIndependentResume`, `-race -count=10`, non-flaky.

## Verification
`go build ./...` clean · `go test -race ./pkg/lifecycle-poc/...` green · full chaos suite green (110s) · `golangci-lint` 0 issues on all touched packages.

## Stated gap (not hidden)
**No Service-level integration test** for N-source fatal/transient status aggregation — `asserterDestination`'s exact-order assertion doesn't fit an interleaved shared destination without forking it. Covered instead by funnel-level tests plus the end-to-end chaos test, which exercise the real production path under real concurrency and crash. Flagging rather than claiming coverage that isn't there.

**N×M is not blocked** by this code (the same per-root lock makes it structurally safe) but is untested and undocumented as supported — that is 3c.


devarismeroxa:
## Adversarial review: REQUEST CHANGES — converting to draft (CI was 11/11 green)

The lock design itself is **sound** — verified in detail, see the bottom. But two High findings and a coverage hole that hides one of them.

### H1 — CONFIRMED, High. Partial graceful stop: some sources torn down, others still reading, pipeline still `Running`, no signal

`stopRunnablePipeline` loops N workers calling `w.Stop(ctx)` with **one shared, deadline-bound ctx**. `Worker.Stop` fails only when `acquireProcessingLock(ctx)` loses to that ctx — and it fails **before** arming `w.stop`.

3 sources, 1 shared destination:
1. `w0.Stop` arms, tears down source 0 — worker 0 exits.
2. Worker 1 is mid-batch holding its own `processingLock` while blocked on `sharedMu` (worker 2's slow destination write). `w1.Stop` eats the remaining budget → ctx error, **not armed**.
3. `w2.Stop` sees an expired ctx, returns immediately, **not armed**.

End state: source 0 gone, sources 1–2 still reading. `workersWg` never drains → cleanup goroutine never runs → status stays `Running`. No status change, no `FailureEvent`, no metric. The only signal is an error returned to a caller that may be a disconnected HTTP client.

**On `main` this state cannot exist** (one worker either arms or doesn't). The shared sink makes it *more* likely: one slow destination write now holds a sibling's `processingLock` via `sharedMu` and can consume the whole stop budget.

Compounding: `stopAndWaitTimeoutErr`'s `"stop"` phase still says *"the pipeline never began stopping … still running, exactly as before this call — safe to retry"*. Written for the single-worker invariant, **now false for N>1**, and it rides on a coded error `ApplyPlanLive` surfaces to operators.

### H2 — PLAUSIBLE, High. Error inside the shared subtree releases `sharedMu` before the tomb is killed → sibling can ack records nobody wrote

`defer sharedMu.Unlock()` fires on the **error** unwind too. `DestinationTask.Do` returns early on an `Ack` error **leaving unread acks on the shared stream**. And the worker goroutine runs `w.Close(...)` — source teardown, persister flush, a multi-second window — **before** `rp.t.Kill(err)`.

So: worker A writes 10, reads 4 acks, Ack errors, A unwinds and releases the lock with 6 acks still queued, then spends seconds in teardown. Worker B takes the lock, enters `DestinationTask.Do`, and its first `Ack()` returns **A's** leftovers. `validateAcks` compares byte-for-byte against B's positions:
- different bytes → loud failure (fine);
- **identical bytes → validation passes**, and B acks its own records upstream on the strength of A's acks. **Invariant 1 violated.**

Identical position bytes across sources is exactly the N-source hazard — positions are connector-defined and unique only *within* a source. Two file sources, two generators, two offset-based sources collide routinely. Telling detail: `tests/chaos/nsource_child.go` gives each source a deliberately **disjoint** position namespace via `posOffset`, so the only end-to-end N-source test explicitly excludes the colliding case.

### M1 — CONFIRMED. DLQ nack threshold silently loosens by a factor of N
One `funnel.DLQ` per source, each with its own window. `windowSize: 100, windowNackThreshold: 5` used to halt after 5 nacks **pipeline-wide**; with 3 sources it now tolerates 5 *per source*. A user-facing safety guardrail changing meaning with no config change and no docs.

### M2 — CONFIRMED. N DLQ connectors share `pl.DLQ.Settings`
Harmless for `builtin:log`; for a file DLQ it's N instances appending to one path concurrently; for anything taking an exclusive resource it's an Open failure at start.

### M3 / M4 / L1–L4
`io.EOF` worker exit logs **nothing** and conflates "source exhausted" with "plugin stream ended" (no in-tree connector can actually signal exhaustion — the branch's real effect is suppressing a previously-reported error class). A comment carried verbatim into an **inverted** branch now documents the opposite of the code. DLQ ID now embeds the pipeline ID twice and can exceed the 256-char limit, refusing a previously-valid pipeline. Plus: all-sources-exhausted reports `UserStopped`; the destination vanished from the `"pipeline tasks"` debug line; all N DLQs share one metric label set.

### The disclosed gap is bigger than disclosed — and H1 lives in it
I said "no Service-level test for fatal/transient aggregation". In fact there is **no Service-level N-source runtime test at all**: `service_test.go` only asserts `buildRunnablePipeline` wiring, the funnel tests build `Worker`/`Sink` by hand, and `tests/chaos/nsource_child.go` imports only `funnel` — never `lifecycle-poc.Service`. So `runPipeline`'s N-worker choreography, its open-rollback, and **all of `stopRunnablePipeline`** have zero execution coverage. H1 sits exactly there.

### Verified sound (the real result)
Lock design is correct: the retry bypass is airtight (`doTaskAttempt` never re-enters `doTask`), only `buildSharedTail` roots are marked and M branches beneath are not, multi-root locks are held in separate goroutines, and ordering `processingLock → at most one sharedMu → multiAckNacker.mu/DLQ.m` is total and acyclic. The "held for the entire call" claim **holds** — `pool.Wait()` joins before `doNextTask` returns, so no branch goroutine outlives the lock. Teardown ordering is clean on every path including recovery restart and force-stop. And the run ledger, `multiAckNacker.posIndex`, `splitRecords` and DLQ attribution are all per-batch/per-worker — **cross-source position collisions cannot confuse them**. The one place position bytes cross the source boundary is the shared destination's ack stream: that's H2.

Fixing H1, H2, M1–M4 and L1, and closing the `stopRunnablePipeline` coverage gap, before this merges.

devarismeroxa:
## Review findings addressed (`9930d7e`)

### H1 — partial graceful stop
`stopRunnablePipeline` now dispatches every worker's `Stop(ctx)` **concurrently** rather than sequentially, so an earlier worker's slow call can no longer consume a later worker's share of the shared deadline. After `wg.Wait()` the workers are partitioned into armed/unarmed:

- **none armed** → unchanged rollback (genuinely safe to retry);
- **some but not all armed** → **escalate**: `rp.t.Kill(FatalError(CodePartialGracefulStopEscalated))`, which cancels the worker ctx so every unarmed worker unwinds via the existing `context.Canceled` path. `workersWg` is guaranteed to drain and the pipeline lands on `Degraded` — it can never sit reporting `Running` with half its sources torn down.

`stopAndWaitTimeoutErr` special-cases the escalation before the generic `DeadlineExceeded` arm, so the "untouched, safe to retry" text is never shown for it.

**Fail-without:** swapping the sequential loop back in makes `TestServiceLifecycle_NSource_PartialGracefulStop_Escalates` fail at the coded-error assertion **and hang to the test timeout**, with a goroutine dump showing `workersWg.Wait()` blocked forever — the exact stuck shape H1 describes.

### H2 — desynchronized shared ack stream
`TaskNode` gains an `atomic.Bool poisoned`. `doTask` checks it immediately after acquiring `sharedMu` (refusing entry with `CodeSharedDestinationPoisoned`), and sets it **inside the critical section, before the deferred `Unlock`**, whenever a pass through a shared subtree returns an error.

That ordering is the whole point: the store happens-before the *other* goroutine's `Unlock`, and the load happens-after *this* goroutine's `Lock`, so the mutex's own release/acquire pair makes the poison visible-before-entry. It does not depend on how fast `rp.t.Kill`'s context cancellation reaches a sibling already blocked on `sharedMu.Lock()` — which is precisely the race the bug exploited. **No new lock and no new blocking wait**, so the acyclic ordering the review verified is preserved.

**Fail-without:** with plain lock/unlock and no poisoning, `TestNSource_H2_AckStreamErrorPoisonsSharedDestination` fails at `is.True(errB != nil)` — worker B's write silently succeeded using worker A's leftover ack.

### The coverage hole
Added **four Service-level N-source tests** that actually run through `lifecycle-poc.Service` (previously there were none — `stopRunnablePipeline` had zero execution coverage): partial-stop escalation, fatal-in-one-source → Degraded, non-fatal → recovery, and a source finishing while another streams.

### M1 / M2 / M3 — documented, deliberately not redesigned
- **M1 (DLQ threshold now per-source):** kept per-source; a shared window adds cross-worker mutable state without a concrete operator need. Documented, with the explicit guidance that operators divide their threshold by N.
- **M2 (N DLQs share `Settings`):** kept; documented as a failure mode, with the warning to avoid exclusive-resource DLQ targets for N-source pipelines. There is no per-source `Settings` override today.
- **M3 (`io.EOF` exit):** now logs at Info with `source_id`, and the design doc states honestly that **no in-tree connector or SDK path can currently signal exhaustion this way** — the branch's real effect today is suppressing an error class that previously triggered recovery.

### M4 / L1 / L2 / L4
Inverted-branch comment corrected. DLQ name no longer embeds the pipeline ID twice (it was hitting the 256-char limit and could refuse a previously-valid pipeline). All-sources-exhausted → `UserStopped` documented. Destination restored to the `"pipeline tasks"` debug line.

### Not claimed
A genuinely ctx-**ignoring** wedged I/O call still cannot be recovered from. That is pre-existing and applies equally to the single-worker path; this fix does not close it and does not claim to.

`go build` clean · `-race` green on `pkg/lifecycle-poc/...` · full chaos suite green (111s) · `golangci-lint` 0 issues · `make generate` clean (two new error codes committed).

Issue #2534, opened 2026-07-06
Flaky test suite: multiple order/timing-dependent failures make CI green unreliable


devarismeroxa:
While merging #2533 (a `cmd/conduit`-only change), the `test` job failed **three consecutive reruns on three different, unrelated suites**, then passed on the fourth. All failing tests pass locally and none were touched by the PR — the full `-race ./...` suite has several independent flaky tests, so a fully-green run is a coin flip.

## Observed flakes (July 2026)
| Test | Package | Likely cause |
|---|---|---|
| `TestServiceLifecycle_Stop/user_stop:_forceful` | `pkg/lifecycle` | timing on forceful stop — **already tracked in #2521** |
| `TestClient_NotFound` / `TestClient_CacheMiss` / `TestClient_CacheHit` | `pkg/schemaregistry` | shared fake-registry / port or cache state under parallel load |
| `TestEncodeDecode_ExtractAndUploadSchemaStrategy` / `TestEncodeDecode_DownloadStrategy_Avro` | `pkg/plugin/processor/builtin/impl/avro/internal` | same schema-registry fixture contention |
| `TestRandOld` | `pkg/plugin/processor/builtin/internal/diff/lcs` | uses global `math/rand` (`rand.Seed(1)`), which other parallel tests share/perturb — seed no longer deterministic under `t.Parallel()` |

## Why it matters
"The test suite is the product's warranty." A suite that fails ~½ of clean runs trains maintainers to reflexively re-run red CI, which is exactly how a *real* regression slips through. It also burns CI minutes (4× here) and blocks solo-maintainer admin-merge (which correctly refuses to bypass a red check).

## Suggested fixes
- **`TestRandOld`**: give it a local `rand.New(rand.NewSource(seed))` instead of the global source, so parallel tests can't perturb its sequence. Cheap, deterministic.
- **schemaregistry / avro**: isolate the fake-registry per test (unique port / fresh cache), or drop `t.Parallel()` on the affected tests.
- **#2521 lifecycle**: track separately (existing).
- Consider a CI step that runs the suite with `-count=2` on a schedule to surface flakes deliberately, and quarantine known-flaky tests behind a build tag until fixed so they stop gating merges.

Not urgent, but it's eroding the value of the gate. Grouping the untracked ones here; #2521 covers the lifecycle case.

devarismeroxa:
## Progress + remaining backlog

**Fixed and merged:**
[...]

PR #2735, opened 2026-08-01
test(chaos): make the mid-snapshot precondition deterministic and self-diagnosing

devarismeroxa:
## What

Partially addresses **#2534**. Removes a real race in the SIGKILL `mid-snapshot` case's setup and makes its precondition self-diagnosing.

**No assertion is weakened.** This changes how the test establishes its premise, not what it checks.

## The race

`mid-snapshot` exists to test a crash before Conduit has persisted **any** position — the highest-stakes edge case in the DBZ-1 design doc. It established that premise by arithmetic:

> `killAfterReads=30` at 1ms/read is ~30ms in — well before the FIRST automatic flush (~1000ms)

On a loaded CI box, 30 reads can take longer than a second. The persister debounce fires, a position **is** persisted, and the case silently stops testing what it claims to. It then fails for a scheduling reason rather than a correctness one — which is exactly the shape of a flake that trains people to hit re-run.

## Changes

**1. `childConfig.persistDelayMS`** — lets a case override the persister debounce. `mid-snapshot` sets 10 minutes, so the first automatic flush cannot beat the kill regardless of machine speed.

Applied to the **first child only**. The resumed child must run with the real default, or it would never flush either — not the scenario under test, and it would make the resume assertion vacuous. I found that by breaking it exactly that way first: with the override applied to both children, both gap tests failed immediately.

**2. The precondition is now asserted, not assumed.** If a flush ever does beat the kill, the failure says so directly rather than surfacing as a confusing downstream mismatch.

## Honest scope — this is hardening, not a proven fix

I could **not** reproduce the original CI failure locally:

| Attempt | Result |
| --- | --- |
| `-count=10`, this branch | pass |
| `-count=10`, clean `main` | pass |
| CI's exact `-shuffle` seed, full suite | pass |
| `GOMAXPROCS=1` + 8 busy loops, `-count=2` | pass |

So I am **not** claiming this resolves the observed flake. It removes a race that is real and reachable by inspection, and adds instrumentation so the next occurrence identifies itself: if `mid-snapshot` fails again with the precondition intact, the cause is elsewhere and this change will have narrowed it.

## Why bother, given that

`tests/chaos (race, x3)` is a **required** branch-protection check and the strongest correctness signal on the arch-v2 engine. Four distinct load-induced chaos flakes turned up in one day of graduation work (#2534 comment has the full list). Each is individually dismissible; collectively they erode the gate, because the habit they train — red required check → re-run → green → move on — is precisely how a real regression gets waved through.

## Verification
Full chaos suite green (110s) · both SIGKILL gap tests green at `-count=3` · green under CPU contention (`GOMAXPROCS=2` + 6 busy loops) · `golangci-lint` 0 issues.

## Still open in #2534
`TestSchemaDrift_TransportHalf_DLQRouting` (failed in CI twice today on unrelated branches, 20/20 locally) and `TestRunLedger_FanOut_PendingRunOnOneBranch_NoPrematureUnanimity`. Not addressed here — they need their own diagnosis rather than a speculative fix.

### DIFF
--- a/pkg/lifecycle-poc/funnel/run_ledger_test.go
+++ b/pkg/lifecycle-poc/funnel/run_ledger_test.go
@@ -530,6 +530,24 @@ func TestRunLedger_FanOut_PendingRunOnOneBranch_NoPrematureUnanimity(t *testing.
 	// (still gated) or p2 (not reached yet - doTask's loop is still
 	// synchronously blocked resolving p1's retry). Only p0 can have reached
 	// unanimity so far.
+	//
+	// Wait for p0's ack to LAND before asserting, rather than assuming it has
+	// already happened. Branch A votes on its own goroutine, so on a loaded
+	// machine destB can have received all 3 records (the condition above)
+	// while branch A has not yet been scheduled to cast its p0 vote —
+	// ackedPositions() is then still nil and the assertion fails with
+	// "<nil> != [0]". That is the test checking too early, not the engine
+	// acking too early; observed twice in CI (#2534) and never reproducible
+	// locally, including at -count=240 under GOMAXPROCS=1 with 10 busy loops.
+	//
+	// This does NOT weaken the check. The wait only requires that SOME ack has
+	// landed; the assertion that follows is unchanged and still demands the
+	// acked set be EXACTLY [p0], so a premature p1 or p2 (the invariant-1
+	// violation this test exists to catch) still fails it — it would surface
+	// as [p0 p1] rather than being waited away.
+	waitForCondition(t, 5*time.Second, func() bool {
+		return len(src.ackedPositions()) > 0
+	})
 	is.Equal(src.ackedPositions(), []opencdc.Position{p0})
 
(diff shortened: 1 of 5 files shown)


devarismeroxa:
Rebased onto current `main` (`cbe697c`). The branch was cut from a stale main and its diff would have reverted ~3,200 lines of today's merged work (the run-ledger, bounded retry, marking-loop fixes and 3b). Now correctly scoped to 3 files in `tests/chaos/`.

The two CI failures on the previous push were **not** mid-snapshot and not caused by this change: `TestSchemaDrift_TransportHalf_DLQRouting` and `TestRunLedger_FanOut_PendingRunOnOneBranch_NoPrematureUnanimity` — the other two known flakes from #2534. The latter lives in `pkg/lifecycle-poc/funnel`, which this diff does not touch at all.

That is now **three of the four** known chaos flakes firing in a single day. Logged on #2534; they need their own diagnosis rather than a speculative fix.

PR #2741, opened 2026-08-01
test: golden-blob format gate + DLQ window v1/v2 differential (pre-flip)

devarismeroxa:
## What

The two cheapest parts of the upgrade/downgrade gap, both required green **before** arch-v2 flips to default in v0.20. Milliseconds to seconds — no docker, no release binaries — so they gate every PR rather than only releases.

## Gate 1 — golden-blob format test

**The problem it closes.** The persisted state blob is plain JSON with **no version or schema tag**, and nothing sets `DisallowUnknownFields`, so an older binary reading a newer blob **silently drops** unknown fields. Worse, `PrepareSet` copies into an explicit whitelist struct rather than marshalling the instance — any future field is silently un-persisted unless someone remembers to add it there.

**A downgrade test that only asserts "the old binary starts and resumes" cannot fail on that.** It is vacuous for exactly the failure it exists to catch. These goldens are what make the downgrade claim real.

Covers a source instance with `SourceState`, a destination with its positions map, a pipeline with DLQ window config, and a processor. Each asserts HEAD parses the checked-in blob to a semantically identical struct **and re-serializes equivalently**.

Plus `TestStore_PrepareSet_WhitelistCoversAllFields`, which uses reflection to populate every exported field with a distinct value — so it does **not** need hand-updating when a field is added; only `PrepareSet` does, which is the actual trap.

### Mutation evidence (4 independent, each reverted)
| Mutation | Result |
| --- | --- |
| Drop `Plugin` from `PrepareSet`'s whitelist | whitelist test red, diff shows `"Plugin":""` vs `"Plugin":"val-g6"` |
| Strip `Plugin` in connector `encode` | both connector golden tests red |
| Strip `DLQ.WindowNackThreshold` in pipeline `encode` | pipeline golden red — the exact field Gate 2 depends on |
| Strip `Condition` in processor `encode` | processor golden red |

## Gate 2 — DLQ window v1-vs-v2 differential

**The problem.** `WindowSize`/`WindowNackThreshold` are **persisted config**. v1 feeds them to `stream.DLQHandlerNode` which counts per **message**; v2 feeds the same values to `funnel.DLQ` which counts per **batch** (`d.window.Ack/Nack(len(batch.records))`). Equivalence was asserted by construction with no test. If it were off, a pipeline that tolerated its error rate on v0.19 would either start hard-failing after upgrade or start DLQ-ing records v1 would have halted on — invariant-3-adjacent, triggered purely by upgrade, and invisible to any format check.

**Result: equivalent.** Verified across the 5 scripted edge cases ported from the existing v1 window tests, the `WindowSize=0` sentinel, and 30 randomized configs × 300 events with **randomly sub-chunked batch splits** — the real risk, since v2's `store()` loop is equivalent to N sequential single-item calls only while batches are homogeneous (which matches how `worker.go` actually calls it).

Proved non-vacuous: an off-by-one in v2's threshold check produced an immediate divergence report (`message 37 ... v1 accepted=false v2 accepted=true`), then reverted.

## Honest scope
Closes the two cheapest slivers only. Does **not** touch binary-to-binary upgrade, the docker matrix, or whether an old binary can parse a genuinely newer schema — those need two real binaries and land as the release gate. The DLQ differential drives the real `DLQHandlerNode`/`DLQ` types with stub I/O; it does not exercise concurrent multi-worker ack/nack interleaving, which is a separate concern from the batching-vs-per-message equivalence it targets.

## Verification
`go build ./...` clean · `-race` green on all four packages · `golangci-lint` 0 issues on touched packages · total runtime ~1-3s.

### DIFF
--- a/pkg/connector/golden_test.go
+++ b/pkg/connector/golden_test.go
@@ -1,0 +1,289 @@
+// Copyright © 2026 Meroxa, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// Pre-flip gates for the arch-v2 default flip (v0.20). The persisted
+// connector.Instance blob is plain JSON with no version/schema tag anywhere,
+// and Store.decode never sets DisallowUnknownFields - so an older binary
+// reading a newer blob silently drops fields it doesn't recognize instead of
+// erroring. Worse, Store.PrepareSet copies into an explicit whitelist struct
+// (icopy) rather than marshalling the instance directly, so a field added to
+// Instance is silently un-persisted unless someone remembers to also add it
+// to that whitelist.
+//
+// These tests make both failure modes loud instead of silent:
+//
+//   - TestGolden_*_RoundTrip checks in the actual JSON currently persisted
+//     for a source and a destination connector instance and asserts HEAD's
+//     Store can parse it into a semantically identical Instance AND
+//     re-serialize it back to an equivalent blob. If a future change drops
+//     or renames a field, the re-serialized blob loses a key the checked-in
+//     golden still has, and the semantic JSON comparison fails.
+//   - TestStore_PrepareSet_WhitelistCoversAllFields populates every exported
+//     field of Instance via reflection (so it does not need to be
+//     hand-updated when a field is added), round-trips it through the real
+//     Store.Set/db path, and diffs what actually got persisted against a
+//     direct marshal of the same Instance. Any field PrepareSet's icopy
+//     whitelist forgets to copy shows up as a concrete, named diff.
+package connector
+
+import (
+	"context"
+	"os"
+	"reflect"
+	"testing"
+	"time"
+
+	"github.com/conduitio/conduit-commons/database/inmemory"
+	"github.com/conduitio/conduit-commons/opencdc"
+	"github.com/conduitio/conduit/pkg/foundation/log"
+	"github.com/goccy/go-json"
+	"github.com/matryer/is"
+)
+
+// assertJSONSemanticallyEqual parses both blobs into generic
+// map[string]any values and compares them, so differences in key order or
+// whitespace don't matter but a dropped, renamed, or zeroed-out field does.
+func assertJSONSemanticallyEqual(t *testing.T, is *is.I, golden, got []byte) {
+	t.Helper()
+
+	var goldenMap, gotMap map[string]any
+	is.NoErr(json.Unmarshal(golden, &goldenMap))
+	is.NoErr(json.Unmarshal(got, &gotMap))
+
+	if !reflect.DeepEqual(goldenMap, gotMap) {
+		t.Fatalf(
+			"re-serialized instance is not semantically equal to the golden blob "+
+				"(a field was likely dropped or renamed):\ngolden: %s\ngot:    %s",
+			golden, got,
+		)
+	}
+}
+
+func TestGolden_SourceInstance_RoundTrip(t *testing.T) {
+	is := is.New(t)
+	ctx := context.Background()
+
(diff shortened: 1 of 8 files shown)


devarismeroxa:
Independent adversarial review found **three holes, every one verified by mutation, every one leaving CI fully green**. Fixed in d9d8b56. Each fix was re-verified by re-applying the same mutation and confirming it now fails.

### 1. The DLQ differential never checked that nacked records reached the DLQ

`stubDLQHandler.Write` was a no-op and `stubDestination.pending` was drained by `Ack` and never inspected — so the gate compared only the two engines' **accepted counts**.

Replacing v2's `sendToDLQ` call with `len(b.records), nil` (v2 stops writing nacked records anywhere) left the suite **green**. That is not cosmetic: `Worker.Nack` acks `positions[:accepted]` upstream on the strength of that count, so those records are acked and gone. Invariant 3, silent, gate passing.

Both stubs now record what they were handed, and `assertParity` compares the delivered multisets — plus a guard that the comparison isn't vacuously empty.

While fixing it, a genuine behavioral difference surfaced: the engines wrap the failed record differently. v1 sets the DLQ record's `Position` to `msg.ID()` (`<connectorID>/<position>`), v2 to the raw record position. Both nest the original in `Payload.After` via `Record.Map()`, so the comparison reads through to that. **The wrapper-format divergence is real and deserves its own decision before the flip** — a DLQ consumer keying on position sees the format change on upgrade — but it is not what this gate measures, so I did not fold it in here.

### 2. It compared accepted/rejected but never fatality — the exact regression it exists to catch

`cerrors.IsFatalError` is what decides auto-recover vs `StatusDegraded` in both engines' `service.go`. The PR's own motivation is "a pipeline that tolerated its error rate on v1 would start hard-failing after upgrade" — that *is* the fatality bit, and it was the one observable not diffed.

Forcing v2 to return `FatalError` where v1 returns a plain error left the suite **green**. Results are now an `{accepted, fatal}` pair.

### 3. `fillNonZero` silently skipped float/pointer/array/interface kinds

Which made `TestStore_PrepareSet_WhitelistCoversAllFields` blind to new `Instance` fields of exactly those kinds — a zero field matches a zero field.

Adding `BackoffFactor float64` and `MaxRetries *int` **without** adding them to `PrepareSet`'s `icopy` whitelist left the test **green** — precisely the trap the test was written to close. It now synthesizes floats, pointers and arrays, and `t.Fatalf`s on any kind it cannot populate rather than skipping it. The failure names the fields directly (`backoff_factor: 14.5` vs `0`).

Worth recording: the golden round-trip tests do **not** exercise `PrepareSet` at all (deleting `Plugin: instance.Plugin` from `store.go` leaves them green while only the whitelist test goes red), so the whitelist test is the *sole* guard against the icopy trap. It needed to fail closed.

### Also

- Dropped two tautological assertions in `replayV2` — `accepted >= 0 && accepted <= c` over `dlqWindow.store`'s own return domain, and `accepted < c` guarded by a stub that never fails — in favor of the contract that actually constrains the caller: a partial acceptance **must** be reported as an error, because `Worker.Nack` acks exactly the accepted prefix.
- Moved the v1 harness goroutine's `is.NoErr` into `stop()`; `t.FailNow` must run on the test goroutine.

### What the review verified as genuinely sound

The golden gates are non-vacuous — every mutation tried failed correctly: new field on `Instance`, new field on nested `Config`, `ProvisionType` iota reorder, `json:"-"` on `DLQ.WindowNackThreshold`, JSON rename, dropping `Status` from `encodableInstance`, processor `Condition` drop, `Parent.Type` rename. The differential genuinely drives both engines (v1 threshold off-by-one, v1 `Ack()` no-op, v2 `window.Nack(1)`, v2 `window.Ack(1)`, v2 `store` off-by-one all failed correctly) — it is not testing one engine twice. The v2-`Ack`-undercount mutation was caught *only* by `RandomizedBatching`, so the property test is carrying real weight. No skips, build tags, or docker preconditions; nothing load-sensitive.

### Known remaining limits, stated rather than papered over

- Goldens cannot catch representation-compatible type swaps (`opencdc.Position` → `string` round-trips byte-identically through the base64 text).
- `pipeline.encodableInstance` is the same whitelist trap as connector's `icopy` with no coverage test. Lower risk — it embeds `*Instance`, so exported fields are automatic — but the asymmetry is real.
- Record-splitting makes v1/v2 window counts genuinely diverge (v2 dilutes at the post-split rate via `DLQ.Ack` but counts nacks at the pre-split rate via `originalBatch()`); v1 has no equivalent. `replayV2` only builds batches where split == original, so this gate does not cover it.

PR #2742, opened 2026-08-01
docs: design doc + ADR for the arch-v2 run join (defer the fan-out)

devarismeroxa:
Design-only, no production code. Resolves the structural questions **before** implementation — the two previous attempts in this area (#2725, #2727) were both rejected on first adversarial pass for questions left open at design time.

## Problem
A chunking processor splits a record; a rate-limited pre-fan-out processor returns fewer records than it received (`Retry`); `subBatchByFlag` partitions the run across flag groups. The first group reaches `doNextTask`, `validateRunsWholeBeforeFanOut` sees `present < total`, and the pipeline dies fatally with `CodeSplitRunStraddlesFanOut` — ×N blast radius, nothing acked, replays into the same failure.

**That is the RAG template** (chunker → embedder → 2 vector sinks) failing on an ordinary rate limit.

## Decision: defer the fan-out
Buffer inside `doTaskAttempt`'s tainted loop and dispatch once the run is complete, rather than dispatching a partial group and joining votes afterwards.

- No cross-call registry, no new shared mutable state across concurrently-running branches — cannot reintroduce the race #2731's review warned about.
- `validateRunsWholeBeforeFanOut` needs **zero changes** and stays trivially true by construction, which answers "what still catches the genuine bug?" without inventing a successor guard.
- Never arms the `originalBatch()` landmine: it drops nil positions, so a pure-tail group at a fan-out collapses to **zero** positions and any position-byte-indexed join breaks there. Deferring means such a group never reaches a fan-out.
- Cost: within a run, fast members wait for slow ones before reaching a destination. Ack latency is unaffected — the run's position cannot be acked until every member is terminal regardless.

**An ADR is included because this reverses a recorded decision.** #2731's review offered the cross-branch join and it was rejected then as speculative generality. What changed: a concrete flagship use case appeared. The existing ADR is referenced, not edited.

## The design change writing this forced — and why design-first paid
The original sketch buffered whole `subBatchByFlag` groups. **That is unsound.** A malformed or version-mismatched plugin can return a `nil` `ProcessedRecord` in the **middle** of its output — `codes.go`'s own documented trigger for `CodeRetryNotConverging` — and `markBatchRecords` classifies it as `Retry` wherever it lands. So one group can legitimately hold a **complete run followed by an incomplete one**.

Replaced with a strict **prefix scan**: dispatch the longest ready prefix, buffer from the first incomplete span onward, even a later run that is independently complete. That is the identical discipline `multiAckNacker.releaseLocked` already uses and that has already survived review.

## Constraints traced (each killed or nearly killed a prior attempt)
Flag semantics untouched · non-blocking accumulate-and-return (the tainted loop is sequential, so a blocking rendezvous is a **deterministic deadlock**, not a race) · in-source-order release · unanimity/nack-wins preserved · crash safety degrades to at-least-once replay · termination bounded by the existing `maxRetryStall`/`CodeRetryNotConverging` machinery · `Filter` and `Nack` paths unchanged.

## Open question carried forward, not assumed away
Composition with the N-source shared-destination machinery (`sharedBoundary`/`sharedMu`). It composes safely by construction — the fan-out stage is threaded through one worker's own call stack, never attached to the shared `TaskNode` — but **that exact composition is where #2734's H2 concurrency bug was found**. It wants a dedicated adversarial pass and a chaos test before code, not a should-be-fine argument.


devarismeroxa:
Independent adversarial review. The docs are unusually well-cited — ~40 line references spot-checked, nearly all exact — but there is a **data-loss-class hole in the design**, plus several factual errors that matter more than usual because **an ADR is immutable once merged**. Recommending changes before this is used as an implementation spec.

I re-verified the load-bearing claims myself against `main`; marked below.

## Blocker — B1: the readiness predicate is unsatisfiable, and it fails silently

The dispatch rule is "the fan-out stage's accumulated count for that run reaches `run.total`". But two paths take run members terminal **without** passing through `doNextTask` — **verified**, `worker.go`:

- a `RecordFlagNack` group goes straight to `acker.Nack`;
- a group with no active records goes straight to `acker.Ack` (both the sub-batch path and the clean-batch path).

Both credit `splitRun.terminalCount` via `runAckNacker.vote`, but contribute **nothing** to the fan-out stage's count. So the count can never reach `total`. The buffered pieces are never dispatched, never written to any destination, and the run's original position reaches neither `Worker.Ack` nor the DLQ. `doTaskAttempt` returns `nil` — no error, no log, no metric. The next batch's ack then advances `State.Position` past the withheld record, which `run_ledger.go` itself describes as "silently skipping the record on restart."

Reachable with no malformed plugin, entirely inside the motivating RAG shape: chunker splits into 5; embedder shortfalls to `[Ack,Ack][Retry×3]` and the `[Ack,Ack]` group is buffered; on retry the embedder returns 3 `FilterRecord`s (drop sub-threshold chunks) or 3 `ErrorRecord`s (provider 4xx). `Batch.Filter` does not set `tainted`, so the retried batch is clean and acks directly; the error path nacks directly. Net: `terminalCount == 3`, `total == 5`, two records on the floor, original position skipped on restart.

**This converts today's loud failure into exactly the silent skip the guard was built to prevent.**

The trap is already documented verbatim in this doc's own primary reference — `20260801-archv2-split-run-ack-ledger.md`'s Limits section spells out that Nack groups and no-active-record groups bypass `doNextTask`, and records that an earlier draft got this wrong and review caught it. The new doc cites that section without carrying the correction forward. The doc explicitly declines to propose a successor guard; **that declination is the defect.**

Minimum: an assertion at the outermost `doTaskAttempt` frame that the buffer is empty on return, failing loud with a coded error — restoring #2731's behavior for these shapes. Better: notify the fan-out stage on the Nack and direct-Ack paths so it can flush or discard.

## Factual errors about the code (each verified)

- **A1 — `CodeSplitRunStraddlesFanOut` is not a `FatalError`.** The doc says it is, and that it "tombs the whole pipeline". `run_ledger.go` returns a bare `conduiterr.New`; `cerrors.FatalError` is never applied (contrast `worker.go`, which wraps explicitly). Since `service.go` branches on `IsFatalError` for `StatusDegraded` and sends everything else to bounded-backoff **recovery**, today's real behavior is a restart crash-loop to the attempt ceiling, not an immediate tomb. The severity framing that justifies the whole reversal is overstated, and an implementer will write the wrong assertion.
- **A3 — the ADR misattributes H2 to #2731's review.** It was #2734's: `worker.go` says "H2 (adversarial review of #2734)" at three sites, and #2734 merged at 22:25Z, four hours *after* #2731 at 18:14Z. The PR body gets this right; the ADR — the artifact that can never be edited — does not.
- **A2** — "Nack propagates across the whole run atomically" is false. `setFlagWithErr` propagates only when `splitRecords` is non-empty **on the receiving batch**, and the scan is within that batch only. A pure-tail sub-batch from `sub()` has `splitRecords == nil`. This is load-bearing: it is the stated reason for excluding Nack from the readiness accounting, which is where B1 comes from.
- **A4** — the motivating trace's `[Ack,Ack] [Retry] [Ack,Ack]` contradicts its own step 2; per the cited padding-at-the-end behavior the shape is `[Ack,Ack] [Retry,Retry,Retry]`. The three-group shape only arises from the malformed-plugin case argued 90 lines later.
- **A5** — the #2730 citation points at `run_ledger.go:220-232` (that's `vote`'s doc about `released`); the referenced section is `batch.go:221-232`.
- **A6** — "extending `multiAckNacker` would need essentially the same buffering/prefix-scan logic" — it already has it (`releaseLocked`), which this same doc cites as the prior art it is copying. The first rejection reason given is valid; this second one is not.

## Other design gaps

- **B2** — in-source-order release is not enforced *across* `doNextTask` calls. A later call whose batch holds no member of the buffered run sees only ready spans and dispatches, jumping the buffered run in source order. The `releaseLocked` analogy does not carry: that scans one fixed complete position vector under one lock; this scans a different partial batch each call, with the buffer as hidden state the scan never consults.
- **B3/B4** — failure mode 5's premise ("an earlier run always finishes before a later run begins") is false under B1, and failure mode 2's "no gap, no double-ack" holds only if nothing acked past the buffered run, which B1/B2 make reachable.
- **B5** — the reconstruction spec omits `filterCount` (and `tainted`). `sub()` recomputes it and `HasActiveRecords`/`ActiveRecords` depend on it, so a hand-reassembled batch with `filterCount == 0` but Filter-flagged records returns filtered records as **active** — filtered records written to every destination. The doc discusses filtered records travelling through the reassembled batch, so this is live in the very case it addresses.
- **B6** — the buffer bound contradicts the prefix-scan rule: step 3 buffers everything from the first incomplete span to the end of the batch, including unrelated spans, so the real bound is the source batch size, not the run's `total`. For RAG that is embedded vectors pinned in memory for the retry chain.
- **B7/B8** — no observability and no upgrade/rollback section; CLAUDE.md requires both. A gauge for buffered runs/records is arguably mandatory here, since it is the only thing that would make B1 visible.

## Alternatives and honesty

- **C1** — a previously-recorded third option is silently dropped. The ledger doc's Limits section names "hoist the ledger into a per-pass registry, **or have each branch complete against the member count it actually received rather than `run.total`**". The ADR claims "three designs considered" but substitutes "extend `multiAckNacker`" for the second. The per-branch-recount option — plausibly the smallest change, and one that sidesteps every buffering hazard above — is evaluated nowhere.
- **C2** — the ADR asserts the flagship shape "now exists... not a contrived edge case"; the design doc marks the same claim **ASSUMED** and notes the shipped template has one destination. **Verified:** `postgres-pgvector-rag/pipeline.yaml` has exactly one destination, no fan-out. So the reversal of #2731's speculative-generality call currently rests on a hypothetical two-destination extension — the same objection — and the overstatement would live permanently in the immutable artifact. Either land the two-destination template in the same train, or have the ADR say plainly that the shape is anticipated, not observed.
- **C4** — ADR alternative A is titled "a per-pass registry" and rejected, but the adopted design is a value created once per batch-read pass holding accumulated records. The real distinction (pre-fan-out single-goroutine vs cross-branch concurrent) is made in the rejection text, but the heading will read to a future maintainer as rejecting what it adopted.

## Verified clean

No deadlock or livelock — deferring introduces no lock, the fan-out stage is per-goroutine call-stack state, and dispatch still goes through the unchanged `doNextTask`, so the `processingLock → sharedMu → multiAckNacker.mu`/DLQ order is untouched. The prefix-scan counterexample is real and correctly argued (`ProcessorTask.Do`'s end→start walk plus `isSameType`'s nil case genuinely produce a Retry span between Ack spans), so rejecting the trailing-only shortcut is right. `acquireProcessingLock` really is held across the whole synchronous pass. ADR conventions — directory, `YYYYMMDD-slug.md`, section order — match the sibling ADR, and it references rather than edits prior docs.

## Nits

Both files are dated `20260802`; today is 2026-08-01. The test plan has **no case where a run's remainder terminates via Nack or via an all-filtered group** — the two cases that decide whether this design is safe; both belong in it, along with a `filterCount`-carried-forward assertion. And the open question the PR body promises to carry forward (composition with the N-source shared-destination machinery) is not in the doc's Open questions — the PR body is ephemeral, the artifact is not.

**Not merging.** B1 needs a design answer before this becomes an implementation spec, and A1/A3/C2 need correcting before anything lands in an immutable ADR.

Issue #2669, opened 2026-07-23
embed: configureEmbeddedMetrics leaks a registry into pkg/foundation/metrics global bookkeeping on failed Register


devarismeroxa:
## Context

`pkg/conduit/runtime.go`'s `configureEmbeddedMetrics` (used by the `libconduit` embed API, see
`conduit.New`/`WithMetricsRegisterer`) builds a fresh `foundation/metrics/prometheus.Registry`
per-Runtime and calls `metrics.Register(reg)` *before* attempting
`registerer.Register(reg)`/`registerer.Register(statsHandler)`:

```go
func configureEmbeddedMetrics(registerer promclient.Registerer) (*promgrpc.StatsHandler, error) {
	reg := prometheus.NewRegistry(nil)
	metrics.Register(reg) // documented cross-talk limitation, see WithMetricsRegisterer's doc
	if err := registerer.Register(reg); err != nil {
		...
		return nil, conduiterr.Wrap(...)
	}

	statsHandler := promgrpc.ServerStatsHandler()
	if err := registerer.Register(statsHandler); err != nil {
		...
		return nil, conduiterr.Wrap(...)
	}
	return statsHandler, nil
}
```

`metrics.Register(reg)` appends `reg` to `pkg/foundation/metrics`'s process-global bookkeeping
(the `global` slices that every `metrics.NewCounter`-style constructor fans values out to — the
same known cross-talk limitation already documented for the happy path). If either
`registerer.Register` call fails (e.g. a caller-reused `Registerer` with a metric-name
collision — see `TestNew_MetricNameCollision_ReturnsCodedError`), `configureEmbeddedMetrics`
returns an error and `conduit.New` fails, but `reg` was already registered into the process
global and is never unregistered. That registry — and any metric values fanned into it — leaks
for the process lifetime.

## Impact

- A failed `conduit.New` call (metric collision, or any other `Register` failure) leaves a
  dangling registry in `pkg/foundation/metrics`' global bookkeeping.
- In a long-running host process that retries `conduit.New` after a fixable misconfiguration
  (e.g. it swaps `MetricsRegisterer`), each failed attempt adds another leaked registry.
- Not a data-path/correctness issue (Invariants 1-7 unaffected) — this is a resource/observability
  leak scoped to the embed metrics seam.

## Accepted as a v1 ("B1") limitation

Tracked, not blocking the current libconduit embed API PR (#2667) — the existing metrics
cross-talk limitation (two Engines observing each other's metric values,
[...]

PR #2747, opened 2026-08-02
fix: use embedded metrics gatherer

### LINKED ISSUE #2670: embed: /metrics HTTP route serves the default gatherer, not the host registerer
## Context

`pkg/conduit/runtime.go`'s `newHTTPMetricsHandler` (wired to the `/metrics` route in
`serveHTTPAPI`) is:

```go
func (r *Runtime) newHTTPMetricsHandler() http.Handler {
	return promhttp.Handler()
}
```

`promhttp.Handler()` always serves `promclient.DefaultGatherer` (the process-global default
registry), regardless of whether this `Runtime` was constructed via the `libconduit` embed API
with `Options.MetricsRegisterer` set to a caller-supplied `prometheus.Registerer` (see
`conduit.New`, `pkgconduit.WithMetricsRegisterer`, `configureEmbeddedMetrics`).

## Impact

- An embedder that sets `Options.API.Enabled = true` and a custom `MetricsRegisterer` gets an
  HTTP `/metrics` endpoint that does not reflect what was registered into their registerer — it
  serves whatever else in the process happens to be in `promclient.DefaultRegisterer` (nothing,
  in most embed scenarios, since the embed path deliberately never registers into the default
  registry — see `Options.MetricsRegisterer`'s doc: "never `prometheus.DefaultRegisterer`
  implicitly"). In practice this means `/metrics` on an embedded engine currently returns an
  effectively empty or unrelated scrape target instead of Conduit's own metrics.
- This is independent of, but related to, the existing metrics cross-talk known limitation
  (`pkg/conduit/doc.go`'s "Known limitation: metrics cross-talk across engines" /
  `TestTwoEngines_MetricsCrossTalk_KnownLimitation`).

## Accepted as a v1 ("B1") limitation

Tracked, not blocking the current libconduit embed API PR (#2667). The embed API's primary
metrics contract for v1 is "metrics reach the host-supplied `Registerer`" (which is genuinely
true — `configureEmbeddedMetrics` does register into it); the HTTP `/metrics` convenience route
is a CLI-path holdover that hasn't been made registerer-aware yet.

## Suggested fix

`newHTTPMetricsHandler` (or its caller in `serveHTTPAPI`) should build its handler from
`promhttp.HandlerFor(gatherer, opts)` where `gatherer` is the same `prometheus.Registerer`/
`Gatherer` passed via `WithMetricsRegisterer` when set, falling back to
`promclient.DefaultGatherer` only on the CLI path (no `MetricsRegisterer` option). This requires
`Runtime` to retain a reference to the configured `Gatherer`, not just the `*promgrpc.StatsHandler`
it currently keeps.

Guflly:
### Description

Use the gatherer associated with an embedded runtime when serving /metrics.

A supplied registry now serves both host and Conduit metrics. Wrapped registerers can provide their underlying registry through MetricsGatherer, preserving prefixes and constant labels. The Prometheus handler's request metrics remain available.

Fixes #2670

Roadmap: Phase 1, Embedded v1  
Risk: Tier 2

### Compatibility and migration

This intentionally changes API-enabled embeds that pass a registerer which does not also implement prometheus.Gatherer. That configuration previously started but served unrelated default metrics. It now returns CodeInvalidArgument on the first Run or Import.

Set both values when using a wrapped registerer:

    registry := prometheus.NewRegistry()
    registerer := prometheus.WrapRegistererWithPrefix(myapp_, registry)

    conduit.Options{
        MetricsRegisterer: registerer,
        MetricsGatherer:   registry,
    }

API-disabled embeds keep their existing startup behavior. This change is planned for the next minor release; the hard error was chosen instead of carrying a known-wrong scrape target through another release.

### Tests

- Focused embedded metrics tests
- go vet ./pkg/conduit
- Pinned golangci-lint on changed code

The full package suite reaches an existing Windows Badger cleanup lock in TestRuntime; CI covers the Linux run.

### Quick checks

### DIFF
--- a/conduit.go
+++ b/conduit.go
@@ -61,7 +61,9 @@ type Options struct {
 	// *object*, but pkg/foundation/metrics' process-global metric
 	// definitions still fan values out across every registry in the process.
 	// When API is enabled, a registerer that does not also implement Gatherer
-	// requires MetricsGatherer or the first Run or Import returns an error.
+	// cannot back /metrics. Conduit warns and keeps serving the default
+	// Prometheus registry for this release; the next minor turns that into an
+	// error (issue #2800). Set MetricsGatherer to fix it — see that field.
 	MetricsRegisterer promclient.Registerer
 
 	// MetricsGatherer is served by the embedded HTTP API's /metrics endpoint.
@@ -386,14 +388,18 @@ func New(_ context.Context, opts Options) (*Engine, error) {
 		return nil, cerrors.Errorf("invalid config: %w", err)
 	}
 
-	// Fail fast on a metrics pairing that cannot work, using the SAME rule
-	// NewRuntime applies later (pkgconduit.ResolveMetricsGatherer). This check
-	// needs no I/O, so deferring it to the first Run or Import would only mean
-	// the caller wires up an Engine, starts a pipeline, and learns about a
-	// typo'd option then — for the same reason cfg.Validate() runs here rather
-	// than at open time.
-	if _, err := pkgconduit.ResolveMetricsGatherer(opts.MetricsRegisterer, opts.MetricsGatherer, cfg.API.Enabled); err != nil {
+	// Apply the SAME metrics rule NewRuntime applies later
+	// (pkgconduit.ResolveMetricsGatherer), here where the options are already
+	// known and no I/O is needed — for the same reason cfg.Validate() runs here
+	// rather than at open time. A rejected pairing fails New; a degraded one
+	// warns now rather than at first Run, so the operator sees it while they
+	// are still looking at their own wiring code.
+	metricsResolution, err := pkgconduit.ResolveMetricsGatherer(opts.MetricsRegisterer, opts.MetricsGatherer, cfg.API.Enabled)
+	if err != nil {
 		return nil, err
+	}
+	if metricsResolution.Degraded {
+		logger.Warn(pkgconduit.MetricsDegradedWarning)
 	}
 
 	runtimeOpts := []pkgconduit.RuntimeOption{
--- a/conduit.go
+++ b/conduit.go
@@ -60,12 +60,14 @@ type Options struct {
 	// section — two Engines each get an isolated Registerer/Registry
 	// *object*, but pkg/foundation/metrics' process-global metric
 	// definitions still fan values out across every registry in the process.
+	// When API is enabled, a registerer that does not also implement Gatherer
+	// requires MetricsGatherer or the first Run or Import returns an error.
 	MetricsRegisterer promclient.Registerer
 
 	// MetricsGatherer is served by the embedded HTTP API's /metrics endpoint.
 	// Set it when MetricsRegisterer does not also implement Gatherer, such as
 	// when using WrapRegistererWithPrefix or WrapRegistererWith. It must gather
-	// the metrics registered through MetricsRegisterer.
+	// the metrics registered through MetricsRegisterer and cannot be set alone.
 	MetricsGatherer promclient.Gatherer
 
 	// DB configures the embedded storage backend. The zero value (Type =="")
--- a/conduit.go
+++ b/conduit.go
@@ -68,6 +68,15 @@ type Options struct {
 	// Set it when MetricsRegisterer does not also implement Gatherer, such as
 	// when using WrapRegistererWithPrefix or WrapRegistererWith. It must gather
 	// the metrics registered through MetricsRegisterer and cannot be set alone.
+	//
+	// Security: Conduit's HTTP API is unauthenticated (see API.HTTP's CORS
+	// documentation for what that already implies). Everything this gatherer
+	// exposes is therefore readable by anyone who can reach the API address —
+	// including metrics the host application registered for its own purposes,
+	// and every label on them. If that registry carries tenant identifiers,
+	// customer names, internal hostnames, or anything else you would not
+	// publish, bind the API to loopback, front it with an authenticating
+	// proxy, or pass a registry that holds only what you are happy to serve.
 	MetricsGatherer promclient.Gatherer
 
 	// DB configures the embedded storage backend. The zero value (Type =="")
@@ -375,6 +384,16 @@ func New(_ context.Context, opts Options) (*Engine, error) {
 	// database — cfg.Validate() does no I/O of its own.
 	if err := cfg.Validate(); err != nil {
 		return nil, cerrors.Errorf("invalid config: %w", err)
+	}
+
+	// Fail fast on a metrics pairing that cannot work, using the SAME rule
+	// NewRuntime applies later (pkgconduit.ResolveMetricsGatherer). This check
+	// needs no I/O, so deferring it to the first Run or Import would only mean
+	// the caller wires up an Engine, starts a pipeline, and learns about a
+	// typo'd option then — for the same reason cfg.Validate() runs here rather
+	// than at open time.
+	if _, err := pkgconduit.ResolveMetricsGatherer(opts.MetricsRegisterer, opts.MetricsGatherer, cfg.API.Enabled); err != nil {
+		return nil, err
 	}
 
 	runtimeOpts := []pkgconduit.RuntimeOption{
--- a/conduit.go
+++ b/conduit.go
@@ -62,6 +62,12 @@ type Options struct {
 	// definitions still fan values out across every registry in the process.
 	MetricsRegisterer promclient.Registerer
 
+	// MetricsGatherer is served by the embedded HTTP API's /metrics endpoint.
+	// Set it when MetricsRegisterer does not also implement Gatherer, such as
+	// when using WrapRegistererWithPrefix or WrapRegistererWith. It must gather
+	// the metrics registered through MetricsRegisterer.
+	MetricsGatherer promclient.Gatherer
+
 	// DB configures the embedded storage backend. The zero value (Type =="")
 	// defaults to an in-memory store — convenient for examples and tests, but
 	// every pipeline configuration is lost once the Engine stops. A
@@ -374,6 +380,9 @@ func New(_ context.Context, opts Options) (*Engine, error) {
 	}
 	if opts.MetricsRegisterer != nil {
 		runtimeOpts = append(runtimeOpts, pkgconduit.WithMetricsRegisterer(opts.MetricsRegisterer))
+	}
+	if opts.MetricsGatherer != nil {
+		runtimeOpts = append(runtimeOpts, pkgconduit.WithMetricsGatherer(opts.MetricsGatherer))
 	}
 
 	return &Engine{
--- a/conduit_test.go
+++ b/conduit_test.go
@@ -658,14 +658,15 @@ func TestRun_MalformedPipelinesDir_ReturnsError(t *testing.T) {
 	is.Equal(ce.Code, conduiterr.CodeInvalidArgument)
 }
 
-// TestNew_WrappedRegistererWithoutGatherer_FailsFast proves the metrics
-// pairing is validated at construction, not on the first Run or Import.
+// TestNew_WrappedRegistererWithoutGatherer_Degrades proves the pairing rule is
+// applied at construction — but as a WARNING, not a refusal.
 //
-// New already validates config in-memory for exactly this reason, and this
-// check is a type assertion — no I/O. Deferring it would mean an embedder
-// wires up an Engine, starts a pipeline, and only then discovers their
-// registerer cannot back the /metrics endpoint.
-func TestNew_WrappedRegistererWithoutGatherer_FailsFast(t *testing.T) {
+// Such a configuration started in earlier releases. Turning it into a startup
+// failure would kill an operator's application on upgrade over a scrape target
+// that is wrong but harmless, which inverts the cost; the deprecation policy
+// (announce, warn, remove) exists for exactly this. The next minor makes it an
+// error.
+func TestNew_WrappedRegistererWithoutGatherer_Degrades(t *testing.T) {
 	is := is.New(t)
 	registry := promclient.NewRegistry()
 	wrapped := promclient.WrapRegistererWithPrefix("myapp_", registry)
@@ -680,6 +681,22 @@ func TestNew_WrappedRegistererWithoutGatherer_FailsFast(t *testing.T) {
 	opts.API.HTTPAddress = "127.0.0.1:0"
 
 	e, err := conduit.New(context.Background(), opts)
+
+	is.NoErr(err)
+	is.True(e != nil)
+}
+
+// TestNew_GathererWithoutRegisterer_FailsFast: this pairing IS rejected,
+// because MetricsGatherer is new API — there is no existing configuration to
+// break — and a gatherer with nothing registering into it cannot work.
+func TestNew_GathererWithoutRegisterer_FailsFast(t *testing.T) {
+	is := is.New(t)
+
+	e, err := conduit.New(context.Background(), conduit.Options{
+		PipelinesDir:    t.TempDir(),
+		DB:              conduit.DBOptions{Type: "inmemory"},
+		MetricsGatherer: promclient.NewRegistry(),
+	})
 
 	is.True(err != nil)
 	is.True(e == nil)
--- a/conduit_test.go
+++ b/conduit_test.go
@@ -20,6 +20,7 @@ import (
 	"os"
 	"path/filepath"
 	"reflect"
+	"strings"
 	"sync"
 	"testing"
 	"time"
@@ -655,4 +656,61 @@ func TestRun_MalformedPipelinesDir_ReturnsError(t *testing.T) {
 	ce, ok := conduiterr.Get(err)
 	is.True(ok)
 	is.Equal(ce.Code, conduiterr.CodeInvalidArgument)
+}
+
+// TestNew_WrappedRegistererWithoutGatherer_FailsFast proves the metrics
+// pairing is validated at construction, not on the first Run or Import.
+//
+// New already validates config in-memory for exactly this reason, and this
+// check is a type assertion — no I/O. Deferring it would mean an embedder
+// wires up an Engine, starts a pipeline, and only then discovers their
+// registerer cannot back the /metrics endpoint.
+func TestNew_WrappedRegistererWithoutGatherer_FailsFast(t *testing.T) {
+	is := is.New(t)
+	registry := promclient.NewRegistry()
+	wrapped := promclient.WrapRegistererWithPrefix("myapp_", registry)
+
+	opts := conduit.Options{
+		PipelinesDir:      t.TempDir(),
+		DB:                conduit.DBOptions{Type: "inmemory"},
+		MetricsRegisterer: wrapped,
+	}
+	opts.API.Enabled = true
+	opts.API.GRPCAddress = "127.0.0.1:0"
+	opts.API.HTTPAddress = "127.0.0.1:0"
+
+	e, err := conduit.New(context.Background(), opts)
+
+	is.True(err != nil)
+	is.True(e == nil)
+
+	ce, ok := conduiterr.Get(err)
+	is.True(ok)
+	is.Equal(ce.Code, conduiterr.CodeInvalidArgument)
+	// The fix has to name a knob this caller can actually reach: an embedder
+	// sees conduit.Options, never pkg/conduit's WithMetricsGatherer.
+	is.True(strings.Contains(ce.Suggestion, "MetricsGatherer"))
+}
+
+// TestNew_PairedRegistererAndGatherer_Succeeds is the other half: the
+// documented migration (pass the underlying registry alongside the wrapper)
+// constructs cleanly.
+func TestNew_PairedRegistererAndGatherer_Succeeds(t *testing.T) {
+	is := is.New(t)
+	registry := promclient.NewRegistry()
+	wrapped := promclient.WrapRegistererWithPrefix("myapp_", registry)
+
+	opts := conduit.Options{
+		PipelinesDir:      t.TempDir(),
+		DB:                conduit.DBOptions{Type: "inmemory"},
+		MetricsRegisterer: wrapped,
+		MetricsGatherer:   registry,
+	}
+	opts.API.Enabled = true
+	opts.API.GRPCAddress = "127.0.0.1:0"
+	opts.API.HTTPAddress = "127.0.0.1:0"
+
+	e, err := conduit.New(context.Background(), opts)
+	is.NoErr(err)
+	is.True(e != nil)
 }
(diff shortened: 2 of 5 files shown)


devarismeroxa:
Thanks for this — it closes a real gap, and the shape of the fix (prefer the supplied registerer when it can gather, otherwise mirror) is the right instinct. I approved the workflow runs so CI can execute (fork PRs sit in `action_required` until a maintainer does that).

I reviewed it for correctness, security and performance rather than just reading the diff, and verified the findings below by running probes against your branch rather than reasoning about them.

## 1. The mirror path silently loses the wrapper's semantics — verified

This is the one worth fixing before merge.

`promclient.WrapRegistererWithPrefix` / `WrapRegistererWith` return a `Registerer` that is **not** a `Gatherer`, so they take the `else` branch. That branch registers the raw collectors (`reg`, `statsHandler`) into a fresh private registry — bypassing the wrapper entirely. The host therefore scrapes prefixed/labelled metrics while `/metrics` serves the raw ones.

Probe against your branch:

```text
wrapped is NOT a Gatherer -> mirror path
host registry: 1 metric families (1 prefixed, 0 unprefixed)
/metrics body contains 'myapp_': false
host has "myapp_conduit_info" ; /metrics serves it: false
```

So with `WrapRegistererWithPrefix("myapp_", reg)`, the host sees `myapp_conduit_info` and `/metrics` serves `conduit_info`. Same data, two different names, no warning. `WrapRegistererWith(labels, …)` would drop the constant labels the same way.

That partly defeats #2670's intent: `/metrics` still doesn't reflect what the embedder actually registered — it just fails differently than before. It IS an improvement over `DefaultGatherer`, but the silent divergence is a sharp edge.

Options, roughly in order of preference:

- Gather **through** the wrapper. `WrapRegistererWithPrefix` has no exported gatherer counterpart, so this likely means asking the embedder for a `Gatherer` explicitly (a `WithMetricsGatherer` option) rather than inferring one.
- Failing that, **document the limitation** where the old #2670 note used to live in `doc.go`, so the constraint is recorded rather than silently removed — the current diff deletes the caveat and replaces it with a fix that doesn't fully hold for wrapped registerers.

## 2. Test coverage gap

`TestNewHTTPMetricsHandlerUsesEmbeddedGatherer` passes a `promclient.NewRegistry()`, which implements **both** `Registerer` and `Gatherer` — so it only exercises the direct path. The mirror branch, which is the more intricate one and the one with finding 1, has no coverage.

Two cases worth adding:

- a registerer that is **not** a `Gatherer` (`WrapRegistererWithPrefix` is the natural fixture) — which is exactly the test that would have surfaced finding 1;
- the **non-embed** path still serving `promclient.DefaultGatherer`. `metricsGatherer` defaults to it in `NewRuntime`, but nothing pins that, so a later refactor could break the default deployment silently.

## 3. Checked and clean

- **`*Registry` is a valid `promclient.Collector`** — `Describe`/`Collect` at `pkg/foundation/metrics/prometheus/registry.go:174,183` — so `gatherer.Register(reg)` in the mirror path is type-correct. Registering the same collector into two registries is legal.
- **No nil-gatherer panic.** `promhttp.HandlerFor(nil, …)` would panic on gather, but `metricsGatherer` is always set in `NewRuntime`, and the only direct `&Runtime{}` constructions are in `readyz_test.go`, none of which reach `newHTTPMetricsHandler`.
- **Performance: no concern.** The extra registry is built once at startup and only walked at scrape time; nothing lands on the record hot path.
- **Security: a behaviour change worth being deliberate about.** On the direct path, `/metrics` now serves the embedder's *entire* registry, including host-application metrics unrelated to Conduit, on Conduit's HTTP port. That's arguably the correct semantic — it's the registry they handed us — but it's a change from "Conduit's own default registry" to "everything you gave us", and an embedder that shares one registry across their app may not expect it. A line in `WithMetricsRegisterer`'s doc would make it a choice rather than a surprise.
- **Error codes** are right: `CodeInternal` for private-registry failures (our own bookkeeping) vs `CodeInvalidArgument` for the caller-supplied registerer.
- The `doc.go` cleanup correctly keeps the separate #2669 leak-on-failure caveat while removing the #2670 one.

Finding 1 and the mirror-path test are what I'd want before merge; the rest is commentary.

Guflly:
Fixed this. I added an explicit gatherer option so wrapped registerers keep their prefixes and labels, covered the wrapped and default paths, and rebased on main.

devarismeroxa (changes requested):
Reviewed by checking out the branch and probing behavior, not by reading alone. The core fix is **correct and does what #2670 asked** — I verified a plain registry now serves both host-registered and Conduit metrics through `/metrics`:

```text
host metric present: true
conduit metric present: true
```

Thanks for picking this up. Four things to resolve before merge, one of which is blocking.

---

### 1. Blocking — this is a breaking change for embedders, undeclared

The PR is marked Tier 2 with no migration note, but it turns a working configuration into a startup failure.

Verified against `main`:

```text
MAIN: NewRuntime ACCEPTS wrapped registerer (no error)
```

and on this branch the same call returns `CodeInvalidArgument`. So an existing embedder doing:

```go
conduit.New(ctx, conduit.Options{
    MetricsRegisterer: promclient.WrapRegistererWithPrefix("myapp_", reg),
    API: ...Enabled,
})
```

works today (serving the wrong gatherer — that's the bug) and **fails to start** after this. `conduit.Options` is a public embed contract, so per CLAUDE.md this needs a versioning plan and migration notes in the PR description.

I'm not saying the hard error is the wrong call — failing loudly beats silently serving wrong metrics, and I'd lean the same way. But it has to be *declared*, and it's worth weighing against the gentler option: log a warning and keep serving `DefaultGatherer` for one minor version, then error. Your call, but the PR should state which and why.

### 2. Blocking — nil gatherer panics `/metrics`

`API.Enabled == false` plus a non-Gatherer registerer skips the error and leaves `r.metricsGatherer == nil`. `promhttp.HandlerFor(nil, ...)` then dereferences it:

```text
metricsGatherer is nil: true
PANIC serving /metrics: runtime error: invalid memory address or nil pointer dereference
```

`TestNewRuntimeAllowsWrappedRegistererWhenHTTPDisabled` constructs exactly this state and asserts only that `NewRuntime` succeeds — so the test currently *documents* the landmine rather than catching it.

The `cfg.API.Enabled` gate couples a constructor-time decision to a config field that's really about a different subsystem. Simplest fix: always resolve a non-nil gatherer (fall back to `DefaultGatherer`, or a `promclient.NewRegistry()`), so the field can never be nil regardless of API state.

### 3. The description doesn't match the implementation

> Otherwise Conduit's collectors are mirrored into a private registry for the HTTP handler.

There is no mirroring in the diff — `grep -n "mirror" pkg/conduit/runtime.go` finds nothing relevant. The code returns an error instead. It looks like the approach changed (`fix: preserve wrapped metrics semantics`) without the description following. Worth fixing, because a reviewer going by the description would approve materially different behavior — mirroring and hard-erroring have opposite compatibility consequences.

### 4. Test gaps

The tests cover the wrapped-registerer paths well, but miss the case the issue is actually about:

- **No test that a plain (Gatherer-implementing) registry serves host metrics** — the primary #2670 scenario. I wrote one to verify it (output at the top); it belongs in the PR.
- **`TestNewRuntimeUsesDefaultGatherer` sets `API.Enabled = false`.** The default-gatherer path matters most when the API *is* enabled, so this asserts the field in the configuration where it's least load-bearing.
- **No test for `WithMetricsGatherer` without `WithMetricsRegisterer`** — that error branch is the first check in the new block and is unexercised.
- **No test pins the nil-gatherer behavior** from (2).

### 5. Minor — undocumented behavior change

`promhttp.Handler()` → `promhttp.HandlerFor(...)` drops the handler's own instrumentation. Verified:

```text
promhttp self-metrics present: false
```

`promhttp_metric_handler_requests_total` and friends disappear from `/metrics` for **all** users, including plain `conduit run`. That may well be fine, but it's a silent change to a scraped surface. If intended, say so in the description; if not, `promhttp.InstrumentMetricHandler` restores it.

---

Nothing here is a criticism of the approach — resolving the gatherer at construction and serving it explicitly is the right shape. (1) and (2) are the ones I'd want fixed before merge; (3)–(5) are cheap.

Happy to push the tests from my probe onto the branch if that's easier than rewriting them.

Guflly:
Addressed in 0cf3db92 and rebased on main. API-disabled wrapped registerers now keep a non-nil gatherer; the API-enabled hard error and migration are documented in the description. I added the plain-registry, default-enabled, gatherer-without-registerer, and nil-safety cases, and kept the promhttp self-metrics. Focused tests and vet pass.

devarispbrown:
Thanks for this one — and for sticking with it through three rounds. You found the right root cause and fixed it in the right place: threading the gatherer through the Runtime rather than papering over `promhttp.Handler()` at the call site is exactly how this should work. The `doc.go` cleanup that removes the now-fixed limitation while keeping #2669 is a detail a lot of PRs miss.

I rebased it locally onto current `main` (it's ~21 commits behind) and ran it: build clean, `pkg/conduit` and the root package green, `-race -count=2` green, `golangci-lint` clean. So the code is healthy. A few things below before it lands — the first one is a genuine "this can take down the embedding app" problem, the rest are smaller.

Heads up on CI: this is a fork PR, so every workflow run is sitting in `action_required` and has never actually executed. That's on our side, not yours — a maintainer has to approve the runs. I'll get them approved so you can see real results.

---

## 1. `InstrumentMetricHandler` can panic the host process

This is the one that blocks merge. `promhttp.InstrumentMetricHandler` panics on any registration error that isn't `AlreadyRegisteredError` (client_golang v1.24.1, `promhttp/http.go`):

```go
if err := reg.Register(cnt); err != nil {
    are := &prometheus.AlreadyRegisteredError{}
    if errors.As(err, are) {
        cnt = are.ExistingCollector.(*prometheus.CounterVec)
    } else {
        panic(err)   // <- here
    }
}
```

Before this PR that registerer was `promclient.DefaultRegisterer` — ours, predictable. Now it's the **host application's** registry, and a host that already registered something named `promhttp_metric_handler_requests_total` with different labels hits the `panic` branch. `newHTTPMetricsHandler` is called from `serveHTTPAPI` (`runtime.go:1026`), which `Run` calls synchronously at `runtime.go:686`, and there's no `recover` anywhere in that path — so the panic unwinds straight into the embedding application and kills it.

I reproduced it:

```go
func TestNewHTTPMetricsHandler_ConflictingHostCollector_DoesNotPanic(t *testing.T) {
    is := is.New(t)
    reg := promclient.NewRegistry()

    // A host collector that collides with promhttp's own, with different labels.
    conflicting := promclient.NewCounterVec(
        promclient.CounterOpts{Name: "promhttp_metric_handler_requests_total", Help: "host's own"},
        []string{"method"}, // promhttp uses "code"
    )
    is.NoErr(reg.Register(conflicting))

    runtime, err := NewRuntime(newMetricsRuntimeConfig(t), WithMetricsRegisterer(reg))
    is.NoErr(err)
    defer runtime.DB.Close()

    handler := runtime.newHTTPMetricsHandler() // panics today

    recorder := httptest.NewRecorder()
    handler.ServeHTTP(recorder, httptest.NewRequestWithContext(
        context.Background(), http.MethodGet, "/metrics", nil))

    // Degraded (no handler instrumentation) but still serving.
    is.Equal(recorder.Code, http.StatusOK)
    is.True(strings.Contains(recorder.Body.String(), "conduit_info"))
}
```

On this branch that fails with:

```
a previously registered descriptor with the same fully-qualified name as
Desc{fqName: "promhttp_metric_handler_requests_total", ...} has different label names
or a different help string
```

Suggested fix — degrade instead of dying. Losing scrape-counter instrumentation is a fine trade for not crashing someone's application:

```go
func (r *Runtime) newHTTPMetricsHandler() http.Handler {
    handler := promhttp.HandlerFor(r.metricsGatherer, promhttp.HandlerOpts{
        ErrorHandling: promhttp.ContinueOnError,
        ErrorLog:      promLogger{logger: r.logger},
    })
    return instrumentMetricHandler(r.logger, r.metricsRegisterer, handler)
}

// instrumentMetricHandler is promhttp.InstrumentMetricHandler with its panic
// contained. That function panics on any registration error that is not
// AlreadyRegisteredError, and r.metricsRegisterer can be supplied by the host
// application — so a name collision in the HOST's registry would otherwise take
// the embedding process down. Serving /metrics without the handler's own request
// counters is a much better outcome than crashing.
func instrumentMetricHandler(logger log.CtxLogger, reg promclient.Registerer, handler http.Handler) (h http.Handler) {
    defer func() {
        if p := recover(); p != nil {
            logger.Warn(context.Background()).
                Msgf("serving /metrics without handler instrumentation: %v", p)
            h = handler
        }
    }()
    return promhttp.InstrumentMetricHandler(reg, handler)
}
```

(If you'd rather avoid `recover`, registering the two collectors yourself and tolerating `AlreadyRegisteredError` works too — the recover version just stays correct if promhttp adds another panic path later.)

## 2. One broken host collector shouldn't blank Conduit's metrics

Related, and folded into the snippet above. `promhttp.HandlerOpts{}` defaults to `HTTPErrorOnError` with no `ErrorLog`. Now that host collectors are in the gather set, a single failing host collector turns the whole scrape into a silent 500 — Conduit's own metrics included, with nothing in the log saying why. `ContinueOnError` plus the runtime logger keeps us observable when host metrics misbehave:

```go
// promLogger adapts the runtime logger to promhttp.Logger so a gather error
// says something instead of vanishing into a 500.
type promLogger struct{ logger log.CtxLogger }

func (l promLogger) Println(v ...any) {
    l.logger.Warn(context.Background()).Msg(strings.TrimSpace(fmt.Sprintln(v...)))
}
```

## 3. Please document the exposure

Our HTTP API has no authentication — `runtime.go:1053-1062` warns about it, and we refuse wildcard CORS on a non-loopback bind for exactly that reason. This PR makes `/metrics` serve the host's **entire** registry, which is the point of the feature, but it also means a host's metric labels (tenant IDs, user IDs, internal hostnames) become readable by anyone who can reach the port. Worth saying out loud where people will see it:

```go
// MetricsGatherer is served by the embedded HTTP API's /metrics endpoint.
// ...
//
// Security: the HTTP API is unauthenticated, so everything this gatherer
// exposes — including metrics the host application registered for its own use —
// is readable by anyone who can reach the API address. Bind to loopback or
// front it with an authenticating proxy if that registry carries sensitive
// labels.
MetricsGatherer promclient.Gatherer
```

## 4. Validation should happen in `New`, not on first `Run`

`conduit.New` already validates config in-memory precisely so misconfiguration surfaces at construction (`conduit.go:372-378`, "gives New its documented fail-fast behavior without needing to open the database"). The new metrics check is a type assertion and a bool — no I/O — but it's deferred to the first `Run`/`Import`, so someone wires up an Engine, runs a pipeline, and only then learns their registerer was wrong.

Since both paths need the identical rule, one shared helper beats two copies that can drift:

```go
// pkg/conduit — exported so the embed package applies the SAME rule at New()
// that NewRuntime applies later; two copies of this would drift.
func ResolveMetricsGatherer(reg promclient.Registerer, gatherer promclient.Gatherer, apiEnabled bool) (promclient.Gatherer, error) {
    // ... the logic currently inline in NewRuntime
}
```

then in `conduit.New`, right after `cfg.Validate()`:

```go
if _, err := pkgconduit.ResolveMetricsGatherer(opts.MetricsRegisterer, opts.MetricsGatherer, cfg.API.Enabled); err != nil {
    return nil, err
}
```

## 5. Error messages: add a `Suggestion`, and name the symbol the caller can actually reach

Our convention is that every user-facing error carries a code *and* a suggested fix. Also worth noting: someone using the embed package sees `conduit.Options`, never `WithMetricsGatherer` — that's `pkg/conduit`'s API, so the current message points them at a symbol they can't call.

```go
e := conduiterr.New(conduiterr.CodeInvalidArgument,
    "metrics registerer does not implement prometheus.Gatherer, so /metrics cannot serve it")
e.Suggestion = "pass the underlying registry as well: conduit.Options{MetricsRegisterer: wrapped, MetricsGatherer: registry} " +
    "(or pkg/conduit's WithMetricsGatherer)"
return nil, e
```

## 6. One thing that's our call, not yours

The hard error is a breaking change: a configuration that used to start now fails. Our policy for public contracts is announce → warn → remove over at least two minors, and the commit type drives the changelog, so `fix:` would file this under bug fixes with no breaking marker.

I lean toward **warn-and-degrade for one release** — keep starting, log loudly, keep serving `DefaultGatherer`, then error in the next minor. The thing being corrected is a wrong scrape target, not a data-safety issue, so breaking startup on upgrade feels harsher than the problem warrants. Your PR description argues the opposite case well ("the hard error was chosen instead of carrying a known-wrong scrape target through another release"), and I don't think that's unreasonable either.

Let me confirm with DeVaris and come back to you — please don't rework it on my say-so yet. If we keep the hard error, the commit just needs to be `feat!:` with a `BREAKING CHANGE:` footer and a migration note in the release notes.

## 7. Tests worth adding

The existing six are well-chosen — the wrapped-registerer prefix assertion (`myapp_conduit_info`) is a nice touch. Four more would close the gaps:

- the panic repro above (#1),
- a gather error from a deliberately broken host collector still returns Conduit's own metrics (#2),
- the embedded gatherer does **not** serve `DefaultGatherer` contents, which is the isolation claim the whole seam rests on,
- `New` rejects a bad registerer immediately, without opening a database (#4).

On `TestNewRuntimeAllowsWrappedRegistererWhenHTTPDisabled`: it asserts a 200 from a handler that's never mounted when the API is disabled, so it's really documenting an internal state rather than user-visible behavior. Consider asserting `runtime.metricsGatherer == promclient.DefaultGatherer` and dropping the HTTP round-trip, or dropping the test — your call.

---

Sorry for the slow turnaround on this. Since "Allow edits by maintainers" is on and you've already put three rounds into this, I'd rather hand you working code than a homework list: I'm going to push the rebase plus items 1, 2, 4 and 5 and the tests as commits on your branch, keeping you as the author. Item 3 (the doc note) comes with them; item 6 stays open until DeVaris weighs in.

Take a look when they land — if you disagree with any of it, revert or rework freely, it's your PR. And if you'd rather have done them yourself, say so and I'll back the commits out; some people would rather write their own fix than receive one, and that's a completely fair preference.

devarismeroxa:
Pushed those as one commit on your branch (`1325231`), authored to you as co-author since it's your fix — the gatherer threading is unchanged, these only harden the edges it exposes.

What's in it:

- **`instrumentMetricHandler`** wraps `promhttp.InstrumentMetricHandler` and catches the panic, falling back to the uninstrumented handler with a logged warning. Perturbation-checked: pull the `recover` out and the new test *panics* instead of failing, which is the production symptom exactly.
- **`ContinueOnError` + the runtime logger** on `HandlerOpts`, so one broken host collector can't blank Conduit's own metrics into a silent 500.
- **`ResolveMetricsGatherer`**, exported from `pkg/conduit`, holding the pairing rule once. `conduit.New` now calls it too, so a bad pairing fails at construction instead of on first `Run`. I went with one shared function rather than two copies specifically because two copies of a rule like this drift, and the symptom would be an Engine that constructs fine and then refuses to run.
- **`Suggestion` on both errors**, naming `conduit.Options.MetricsGatherer` — the knob an embedder can actually reach.
- **The security note** on `Options.MetricsGatherer` about the unauthenticated API.
- **Four tests**: the panic case, the broken-collector case, the isolation claim (an embedded gatherer must not serve the default registry — that one had no coverage and it's the assertion the whole seam rests on), and `New` failing fast.

Local run on your branch: `go test . ./pkg/conduit/... -count=1` green, `golangci-lint` (CI-exact v2.12.2) 0 issues.

Two things still open, neither needing anything from you right now:

1. **The breaking-change question (item 6).** Still with DeVaris. If we go with warn-and-degrade for a release, I'll write that change — no reason for you to do it twice.
2. **CI.** I'm getting the workflow runs approved so you can see real results instead of my word for it.

I deliberately didn't rebase, so your commits are untouched and nothing you have locally is invalidated — a `git pull` should be clean. And the standing offer holds: if you'd rather have written any of this yourself, say so and I'll drop the commit. Some people would rather own their fix end to end, which is completely fair.

Genuinely nice work on this — the root cause here was not obvious, and the `doc.go` cleanup that keeps #2669 while removing the fixed limitation is the kind of detail that usually gets missed.

devarismeroxa:
DeVaris made the call on item 6: **warn-and-degrade for one release**, then error. Pushed as `e205e35`, so you don't have to rework it.

What changed from your version: the API-enabled, non-Gatherer case no longer refuses to start. It logs a warning and keeps serving the default registry — exactly the pre-existing behavior — and #2800 tracks turning it into an error in the next minor. Both the godoc and the warning text carry that deadline.

Your reasoning in the PR description wasn't wrong, for the record: *"the hard error was chosen instead of carrying a known-wrong scrape target through another release"* is a real argument, and if this were a data-safety issue it would have won. What tipped it is that the failure is an observability defect — the metrics are wrong, nothing is at risk — while the cost of the error lands on someone whose application stops booting after a routine Conduit upgrade. Our deprecation policy is announce → warn → remove over at least two minors, and this is the shape of change it was written for.

One consequence worth calling out: the warning is now the *entire* mitigation for a release, which makes it load-bearing rather than decorative. So it names the fix, names the deadline, is emitted from both entry points via one shared constant, and has a test asserting an operator actually receives it. A silent degrade would leave someone staring at a dashboard full of unrelated metrics with nothing anywhere explaining why.

Your `TestNewRuntimeRequiresGathererForWrappedRegisterer` became `TestNewRuntimeDegradesForWrappedRegisterer` — same case, inverted expectation. That's the only test of yours I changed, and only because the behavior it pinned is the behavior we decided to reverse.

The gatherer-without-registerer case stays a hard error, since `MetricsGatherer` is new API in this PR — no existing configuration can break — and it can't work under any reading.

Side effect you'll like: with the break gone, `fix:` is now the correct commit type, so the changelog concern goes away too.

Local: `go test . ./pkg/conduit/... -race` green, `golangci-lint` clean. CI is running on your branch now that the workflows are approved. If it comes back green I'll merge it — and thank you for the patience through four rounds and a policy reversal that wasn't yours to make.

devarismeroxa:
Merged as `11e4d30`. Thank you — this was a good fix and you stuck with it through four rounds, a panic nobody had noticed, and a policy reversal that wasn't yours to make.

What you actually shipped, for the record:

- `/metrics` on an embedded Runtime now serves the embedder's own gatherer instead of the process-global default registry, which closes #2670 — the bug you found and correctly root-caused.
- A supplied registry serves host metrics *and* Conduit's, so an embedder gets one scrape endpoint rather than two.
- Wrapped registerers (`WrapRegistererWithPrefix` and friends) work properly via the new `MetricsGatherer`, with prefixes and constant labels preserved.
- `#2800` tracks turning the degraded case into a hard error next minor — your original position, on the deprecation policy's schedule.

If you want another: **#2669** is the sibling of the bug you just fixed, in the same file. A failed metrics registration during `ensureRuntime` leaks a registry into `pkg/foundation/metrics`' process-global bookkeeping. Same area, same shape of reasoning, and you now know this code better than most people do. No obligation — but you'd be starting from a running start.

Either way, thanks for the contribution. Open a PR any time; I'll get the CI approvals unblocked faster next time now that I know to watch for them.

Issue #2746, opened 2026-08-02
WaitPipeline returns the pre-recovery error while recovery is in its backoff wait


devarismeroxa:
## Summary

`TestServiceLifecycle_NSource_TransientErrorOneSource_Recovers` (`pkg/lifecycle-poc/service_test.go`) is intermittently red, and in at least one CI run it did not fail — it **hung for the full 10-minute test timeout**, taking the whole `test` job down on `main`.

## Evidence

CI, `main` @ `d6b33de`, run `30732975204`:

```text
panic: test timed out after 10m0s
	running tests:
		TestServiceLifecycle_NSource_TransientErrorOneSource_Recovers (9m59s)

FAIL	github.com/conduitio/conduit/pkg/lifecycle-poc	600.203s
```

Locally reproducible as a fast failure rather than a hang:

```bash
go test ./pkg/lifecycle-poc/ -run 'TestServiceLifecycle_NSource' -race -count=40 -timeout 900s
```

roughly **2 failures per 40-count sweep** of the suite, each `--- FAIL ... (0.01s)`.

## The trigger is cross-test interaction

Run alone it is solid:

```bash
go test ./pkg/lifecycle-poc/ -run 'TestServiceLifecycle_NSource_TransientErrorOneSource_Recovers' -race -count=60
```

→ 60/60 pass. It only fails when the sibling `NSource` tests run alongside it, which points at shared state or a leaked goroutine/pipeline from a neighbouring test rather than at this test's own logic.

Two other tests in the same suite have also been seen failing in the same sweeps: `..._OneSourceFinishesGracefully_StaysRunningUntilAllExit` and `..._FatalErrorOneSource_DegradesWholePipeline`. The first is explained by the `io.EOF`-while-stopping bug fixed in #2745; the other two are not.

## Why this matters beyond flakiness

- `tests/chaos (race, x3)` is a **required** branch-protection context. A test that can hang for 10 minutes rather than failing fast turns a red build into a timed-out build, which is much harder to attribute and much slower to surface.
- It is in the arch-v2 N-source recovery path — exactly the area being graduated for the v0.20 default flip. A recovery test that intermittently doesn't recover is either a test bug or a real recovery bug, and we should know which before flipping.
- Per this repo's own process note, chasing red rather than re-running is how the last several real bugs were found. A 10-minute hang is the strongest possible signal not to paper over it.

## Suspicion, not a conclusion

[...]

PR #2748, opened 2026-08-02
bench(archv2): benchmark methodology + retraction of the v1-vs-v2 numbers

devarismeroxa:
**Tier 2.** Closes the benchi half of the arch-v2 graduation gate.

## The gate

Pre-committed before any number existed: **a >10% regression on default config means the flip does not ship in v0.20.** Agreeing the threshold up front was the point — the number arrives with a meaning already attached instead of being argued about afterwards.

## Result — cleared. v2 is faster by single-digit percent

Three full runs, 60s steady state, 59 samples per engine per run.

| run | v1 median msg/s | v2 median msg/s | delta |
| --- | --- | --- | --- |
| 1 | 23,547 | 25,589 | **+8.7%** |
| 2 | 24,628 | 25,244 | **+2.5%** |
| 3 | 23,697 | 25,425 | **+7.3%** |

- v2 faster in **3/3** runs; delta median +7.3%, range +2.5% to +8.7%.
- Pooled (n=177 per engine): v1 23,758 · v2 25,402 · **+6.9%**.
- One-sided permutation test, 20,000 resamples: **p < 0.0001**.
- CPU: v2 ~133% vs v1 ~151%. Memory: v2 ~128 MB vs v1 ~122 MB.

**One run was not enough, and the repeats were not ceremony.** Single-run sd is 8–12% of the median. In run 1 alone the distributions overlapped so heavily that v1's *best* sample beat v2's median — reporting that run by itself would have been indistinguishable from noise. Read the magnitude as "single-digit percent," not as +6.9% precisely.

## Two things this corrects

**The regression I predicted didn't appear.** `sdk.batch.size`/`batch.delay` default to 0, so the SDK calls `readFn(ctx, 1)` and arch-v2 reads batches of **one**, while the fan-out ADR's ~6.3× was measured on a mocked 1000-record batch. I expected that to hurt v2 on default config. It doesn't — batch-of-one is real but costs nothing measurable on this pipeline.

**It is nowhere near 6.3×.** A real pipeline on default config shows single-digit percent. That ADR figure should not be quoted as what users get, and this PR is the number I'd stand behind instead.

## Scope — stated, not implied

- One source, one destination, `builtin:generator` → `builtin:log`, no external infrastructure. This is the **default-config** gate; the **N×M shape remains unmeasured**.
- Developer laptop under Docker Desktop, not dedicated hardware. Absolute numbers are environment-specific. The claim is the v1-vs-v2 *ratio*, measured on the same machine, same image, back to back, differing only in `--preview.pipeline-arch-v2`.
- Metrics **on** for the v2 run — the flip ships with them on, so measuring with them off would describe something users don't get.

## What this does and doesn't unblock

Unblocks the throughput objection to the v0.20 flip. Does **not** clear the rest of the pre-flip list: N×M shapes, the run-join implementation, the upgrade/downgrade suite, and #2746 (an N-source recovery test that intermittently fails and once hung for the full 10-minute CI timeout) are all still open.

## Reproducing

```bash
docker build -t conduit-bench:archv2-parity .
docker network create benchi
cd benchi/archv2-parity
script -q /dev/null benchi -config bench.yml -out ./results/<name>
```

Two mechanical gotchas hit while getting this to run, both now handled in-repo: benchi does a `docker compose pull` before each tool which fails on a locally-built image (`pull_policy: never`), and it requires a TTY (`script -q /dev/null`).


devarismeroxa:
## Update: N×M added — and it reverses the sign

`036ed53` adds `benchi/archv2-nxm/`: 2 sources × 2 destinations, the shape the 1×1 benchmark structurally cannot see. This is where the engines differ — v1 fans in/out via `FaninNode`/`FanoutNode`, arch-v2 runs one `Worker` per source and serializes them into a shared destination subtree behind a per-shared-root mutex.

**Seven runs. v2 slower in 6 of 7.**

| | 1×1 (`archv2-parity`) | 2×2 (`archv2-nxm`) |
| --- | --- | --- |
| pooled delta | **+6.9%** (v2 faster) | **−3.4%** (v2 slower) |
| significance | p < 0.0001 | p = 0.0019 |
| 95% CI | — | [−5.9%, −0.1%] |

**Both inside the >10% gate, so throughput does not block the flip.** But the direction flip is the real finding: arch-v2 wins on the linear path and gives some of it back under shared-sink contention. Recorded as a genuine regression rather than rounded to "no change".

### Why seven runs, and a correction

I reported run 1's **−8.7%** as the N×M answer before repeats existed. That was wrong. The first three runs gave −9.0%, **+16.5%**, −6.4% — pooled −4.0% at p = 0.08, i.e. **inconclusive**. Four more were needed before anything could be claimed.

Run 2 (+16.5%) is **kept**, not discarded: its v1 baseline came in at 12,058 against ~15,300–16,000 everywhere else with sd 18.6%, so the host disturbed that run. Dropping an outlier because it contradicts the conclusion is how benchmark numbers stop meaning anything. With seven runs it no longer drives the result.

N×M is materially noisier than 1×1 (sd 9–21% vs 8–12%) — expected for two workers contending on a shared subtree, and precisely why a sub-10% effect needs repetition here. Worth knowing for anyone re-running this: three runs is not enough at this shape.

### Scope

2×2 only. Wider fan-outs would load the shared-root mutex harder and are not covered. Same environment caveat as the 1×1 config — laptop under Docker Desktop, so the claim is the ratio, not the absolute numbers.

devarismeroxa:
## Update: the N×M regression is a batch-size artifact, not a design cost

`8619f3d` adds `benchi/archv2-nxm-batched/`, testing a specific hypothesis rather than just re-measuring.

**Hypothesis:** arch-v2 serializes N workers into a shared destination subtree behind a per-shared-root mutex. At the shipped defaults the SDK reads batches of **one** record, so that mutex is taken and released **once per record per destination**. v1 has no equivalent — it moves records through channels between fan-in/fan-out nodes. If that's the cause, batching should amortize the lock and the regression should vanish.

**It vanishes, and inverts.** Same 2×2 pipeline, only change is `sdk.batch.size: 100`:

| N×M 2×2 | v1 median | v2 median | delta |
| --- | --- | --- | --- |
| default (batch of 1) | 15,258 | 14,742 | **−3.4%** |
| batched (size 100) | 15,191 | 24,624 | **+62.1%** |

Per-run: +60.3%, +74.6%, +60.9%. Permutation test **p < 0.0001**.

**The decisive detail is that v1 does not move** — 15,258 unbatched vs 15,191 batched, flat. v1 gets nothing from batching. arch-v2 goes 14,742 → 24,624, **+67% on the same pipeline**.

So the regression is not a property of arch-v2's design. It is the cost of the one configuration that maximally penalizes it — per-record lock traffic with zero amortization — and that configuration is the current default.

### Full picture

| shape | config | delta (v2 vs v1) |
| --- | --- | --- |
| 1×1 | default | **+6.9%** |
| 2×2 | default | **−3.4%** |
| 2×2 | batched | **+62.1%** |

arch-v2 is never meaningfully slower, and is dramatically faster the moment records arrive in batches.

### Options for closing the default-config gap

Recorded, not decided:

1. **Amortize the shared-root lock inside the worker.** The mutex is acquired per `doTask` pass; if a pass covers more records the per-record cost falls with no user-visible change. This is the one that helps users who never touch batch settings.
2. **Change the shipped `sdk.batch.size` default.** Highest leverage, lowest code risk — but SDK-wide, affecting every connector and both engines, so it needs its own decision and compatibility analysis.
3. **Document it.** Weakest: defaults are what most pipelines actually run.

Option 1 is the natural follow-up to this measurement.

### Scope

`batch.size: 100` was chosen as a plainly-batched value, not tuned — the curve between 1 and 100 is unmeasured. Three runs suffice here because the effect is large; the −3.4% default result needed seven precisely because it was small. 2×2 only; wider fan-outs load the mutex harder.

devarismeroxa:
## Retitled and rescoped: this PR now retracts its own numbers

The original title claimed the gate was cleared. It wasn't — the measurement was invalid. `5e6ce14` adds `benchi/METHODOLOGY.md` and stamps a retraction banner on all three config READMEs so the figures can't be cited.

**Every engine-comparison number from this branch is withdrawn:** 1×1 +6.9%, 2×2 −3.4%, 2×2-batched +62.1%, and the later 1×1 −28.8% / 2×2-batched +196%.

### Defect 1 — the metric isn't comparable between engines

benchi's `msg-rate-per-second` comes from Conduit's own metrics, and the engines don't instrument the same event:

- **v1** observes inside an **ack handler** (`pkg/lifecycle/stream/metrics.go`) — one observation per acked record.
- **arch-v2** calls `ConnectorMetricsImpl.Observe` from **both** `SourceTask.Do` (on read) **and** `DestinationTask.Do` (on ack).

Against ground truth (records actually written to a file), same pipeline: ground truth **6,342 rec/s** where benchi reported **29,537** — ~5× off, with the v1-vs-v2 delta ~5× off too. Those comparisons measured instrumentation, not throughput.

### Defect 2 — the effect was smaller than the noise floor

No A/A control was ever run. When one finally was — v1 against **itself** — 30s runs reported **+13.0%** and **−5.0%**. Every delta claimed at that duration was smaller than the harness's error against a known-zero difference.

### What the corrected method establishes

Ground truth, 60s runs, warmup discarded, alternating singles, n=6:

| | median rec/s | sd |
| --- | --- | --- |
| v1 | 338,200 | 9.8% |
| arch-v2 | 252,954 | **20.2%** |

Median delta **−25.2%**, permutation **p = 0.16** — large effect, **not statistically established** at this n. Two open threads: arch-v2's variance is double v1's and looks **bimodal** (mostly ~250k, once 358k — v1 territory), and the environment degraded across the session (v1's own sd rose ~3% → 9.8%, consistent with thermal drift).

### Recommendation

Don't settle the v0.20 gate on this machine. A >10% gate needs an A/A floor that stays well under 10% for the whole run. Either dedicated/CI hardware with an A/A control in the same session, or an in-process Go benchmark with no Docker and no destination I/O.

**Rule going forward: no gate decision without an A/A floor quoted beside the result.** A number without one isn't evidence.

The configs are kept — they're reusable once the platform question is settled. Only the results are retracted.

PR #2855, opened 2026-09-05
fix(cli): conduit init must generate a config conduit run accepts

devarismeroxa:
## What

`conduit init` wrote a `conduit.yaml` that `conduit run` then refused to start with. The documented getting-started walkthrough — `conduit init` → `conduit pipelines init` → `conduit run` — failed on **v0.19.0 and on today's main**, from a clean directory:

```
$ conduit init && conduit pipelines init && conduit run
error: failed to set up conduit runtime:
    github.com/conduitio/conduit/pkg/conduit.(*Entrypoint).Serve
        .../pkg/conduit/entrypoint.go:48
  - invalid config:
  - "api.http.cors.allowed-origins" config value is invalid:
    github.com/conduitio/conduit/pkg/conduit.invalidConfigFieldErr
```

Found by an end-to-end developer-experience audit, not by CI — no test walks the path a first-time user walks.

## Root cause

`processConfigStruct` (`init.go`) rendered every field with `fmt.Sprintf("%v", …)`. On an empty `[]string` that produces `[]`, which the YAML encoder writes as the **quoted scalar** `'[]'`. viper decodes that back as `[]string{"[]"}` — a one-element allowlist — and `Validate()` rejects it.

Three fields are affected, all `[]string`:

| Field | Effect |
|---|---|
| `api.http.cors.allowed-origins` | hard startup failure — the only one validated |
| `processors.egress.allow` | engine-level egress allowlist, malformed baseline |
| `processors.egress.secret-refs` | engine-level egress secret grants, malformed baseline |

The egress pair is not currently exploitable — the ceiling is gated by `processors.egress.enabled: false` — but an operator who starts from the generated file starts from a malformed allowlist.

## Fix

`YAMLTree` gains `InsertSeq`, which writes a real `SequenceNode`: `[]` in flow style when empty, a block list when populated. `Insert`'s path-walking is extracted into a shared `insertNode`; the scalar path is byte-for-byte unchanged. `processConfigStruct` routes `reflect.Slice` fields through `InsertSeq`.

Generated output, before and after:

```yaml
# before                          # after
allowed-origins: '[]'             allowed-origins: []
allow: '[]'                       allow: []
secret-refs: '[]'                 secret-refs: []
```

Per-key usage comments are retained, so the generated file keeps its documentation role. A populated list renders as a block sequence (verified).

## The test that would have caught it

`TestInit_GeneratedConfigIsAcceptedByRun` runs `init`, then feeds the generated `conduit.yaml` back through `ecdysis.ParseConfig` — the exact viper path `RunCommand.Config()` uses — and calls `Validate()`.

The three existing `init` tests assert the file **exists**; none asserted it **works**. That was the gap.

**Verified to fail without the fix:**
```
init_test.go:173: 1 != 0
--- FAIL: TestInit_GeneratedConfigIsAcceptedByRun
```
(`AllowedOrigins` decoded as one element containing `"[]"` — the diagnosis, stated by the test.)

`cmd/conduit/internal/yaml_test.go` is new: direct coverage for the empty-sequence, non-empty-sequence, and scalar paths.

## Verification

- `go test -race ./cmd/conduit/internal/... ./cmd/conduit/root/initialize/... ./cmd/conduit/root/pipelines/...` — green
- `golangci-lint run` (pinned v2.12.2 via `make install-tools`) on touched packages — 0 issues
- Manual, from a clean temp dir with a binary built off this branch: `init` → `pipelines init` → `run` → **12 records flowed**, no config error

## Adversarial self-review

- **Nil vs empty slice** — `reflect.Value.Len()` is 0 for both; both take the empty-sequence path. Covered.
- **`[]byte` config fields** — would render as a sequence of numbers. Considered and declined: no `[]byte` field exists, all three slices are `[]string` and enumerated in the test. Adding speculative handling would be a YAGNI violation.
- **Pre-existing, not fixed here** — `insertNode` walking into an existing `ScalarNode` on a path collision (`a.b` inserted, then `a.b.c`) would append to a scalar and corrupt the node. Unreachable with the current config struct (no colliding `long` tags) and unchanged by this PR. Flagging, not fixing.
- **Blast radius** — `YAMLTree` has exactly one non-test consumer, `init.go`. Confirmed by grep.
- **Unrelated** — `gofumpt` flags `cmd/conduit/internal/deploy/{deploy,service}.go` on `main` already; untouched here.

## Risk tier

**Tier 2** — CLI surface. No data-path change, no config-schema change (this fixes a malformed *value*, not the schema), no serialized-format change.

## Roadmap

v0.20 ship blocker: the documented first-run path.

https://claude.ai/code/session_016xE861dwb3MLgqEdWRECLY

### DIFF
--- a/cmd/conduit/internal/yaml.go
+++ b/cmd/conduit/internal/yaml.go
@@ -34,8 +34,39 @@ func NewYAMLTree() *YAMLTree {
 	}
 }
 
-// Insert adds a path with a value to the tree
+// Insert adds a path with a scalar value to the tree.
 func (t *YAMLTree) Insert(path, value, comment string) {
+	t.insertNode(path, &yaml.Node{
+		Kind:  yaml.ScalarNode,
+		Value: value,
+	}, comment)
+}
+
+// InsertSeq adds a path holding a YAML sequence of scalars.
+//
+// A sequence must be inserted as a real SequenceNode, never as a scalar: a
+// Go slice formatted through fmt (%v) yields "[]" or "[a b]", which the
+// encoder then emits as the *quoted string* '[]' — a value the config loader
+// rejects. That produced a conduit.yaml `conduit run` refused to start with,
+// so slices go through this method and never through Insert.
+func (t *YAMLTree) InsertSeq(path string, values []string, comment string) {
+	seq := &yaml.Node{Kind: yaml.SequenceNode}
+	if len(values) == 0 {
+		// Flow style so an empty sequence renders as `key: []` rather than a
+		// dangling key with a nil value.
+		seq.Style = yaml.FlowStyle
+	}
+	for _, v := range values {
+		seq.Content = append(seq.Content, &yaml.Node{
+			Kind:  yaml.ScalarNode,
+			Value: v,
+		})
+	}
+	t.insertNode(path, seq, comment)
+}
+
+// insertNode walks/creates the mapping path and attaches valueNode at the leaf.
+func (t *YAMLTree) insertNode(path string, valueNode *yaml.Node, comment string) {
 	parts := strings.Split(path, ".")
 	current := t.Root
 
@@ -50,13 +81,13 @@ func (t *YAMLTree) Insert(path, value, comment string) {
 			keyNode.HeadComment = "# " + comment
 		}
 
-		var valueNode *yaml.Node
+		var node *yaml.Node
 		found := false
 
 		// Search for existing key
 		for j := 0; j < len(current.Content); j += 2 {
 			if current.Content[j].Value == part {
-				valueNode = current.Content[j+1]
+				node = current.Content[j+1]
 				found = true
 				break
 			}
@@ -64,18 +95,15 @@ func (t *YAMLTree) Insert(path, value, comment string) {
 
 		if !found {
 			if isLast {
-				valueNode = &yaml.Node{
-					Kind:  yaml.ScalarNode,
-					Value: value,
-				}
+				node = valueNode
 			} else {
-				valueNode = &yaml.Node{
+				node = &yaml.Node{
 					Kind: yaml.MappingNode,
 				}
 			}
-			current.Content = append(current.Content, keyNode, valueNode)
+			current.Content = append(current.Content, keyNode, node)
 		}
 
-		current = valueNode
+		current = node
 	}
 }
--- a/cmd/conduit/internal/yaml.go
+++ b/cmd/conduit/internal/yaml.go
@@ -58,7 +58,13 @@ func (t *YAMLTree) InsertSeq(path string, values []string, comment string) {
 	}
 	for _, v := range values {
 		seq.Content = append(seq.Content, &yaml.Node{
-			Kind:  yaml.ScalarNode,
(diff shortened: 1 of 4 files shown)


devarismeroxa:
## Adversarial review round 1 — all three findings addressed

Fresh-context review found a real defect. Fixed in `65eba3a`.

### 1. `InsertSeq` elements were untagged (BLOCKING in effect, though unreachable today)

YAML quotes only what syntax forces, so a plain-but-resolvable element is emitted bare and decodes as another type. Measured on the encoder + decoder this PR relies on — **six elements in, four out**:

```
in:  ["true","false","null","~","123","0755"]
out: ["true","false","123","0755"]
```

`"null"` and `"~"` resolved to null and **vanished from the `[]string`**. (Sharper than the review predicted — they don't blank, they drop.)

Elements now carry `Tag: "!!str"`. Not reachable from `conduit init` today since all three slice defaults are empty, but `InsertSeq` is a general method and the next populated default would have hit it silently.

The review's sharpest point: `TestInsertSeq_NonEmptyRoundTrips` existed *specifically* to cover this branch and picked `api.openai.com` / `https://api.voyageai.com:443` — values structurally incapable of exposing the defect. That's how it survived the first pass. The test now uses the hostile set above and **fails without the tag**.

### 2. The regression test restated `RunCommand.Config()` instead of calling it

It omitted `ExcludedFlags`, and passing a bare `cobra.Command` bound **zero** pflags where real `run` binds ~40 — including a `StringSlice` for `api.http.cors.allowed-origins`. Inert today, but a test that *approximates* the path it guards goes green while `conduit run` breaks. That is the precise failure mode this PR exists to close.

It now builds the real `run.RunCommand`, parses `--config.path` against its real flag set, and calls its own `Config()`. No restatement left to drift.

### 3. The test inherited `CONDUIT_*` from the shell

`ecdysis` binds viper with `AutomaticEnv`, and `CONDUIT_CONFIG_PATH` replaces the config path outright — a stray export made the canary validate a different file. `CONDUIT_*` is now cleared for the test's duration.

### Not fixed, deliberately

`Insert`'s scalar path is untagged in the same way (review finding 4). Tagging it `!!str` would quote every bool, int and duration in the generated `conduit.yaml`. All 23 string defaults were checked and none is ambiguous. Left alone rather than risk the file this PR fixes.

The pre-existing `insertNode` path-collision hazard also stands unfixed and unreachable, as originally flagged.

### Re-verified after the amendments

- Regression test still **fails** against `origin/main`'s `yaml.go` + `init.go` (`1 != 0`), passes here — re-proved after the rewire, not carried over from before it
- `go test -race` green across `cmd/conduit/internal/...` and `cmd/conduit/root/initialize/...`
- `golangci-lint` 2.12.2 — 0 issues; `gofumpt` clean on all touched files
- Binary built from this branch: `init` → `pipelines init` → `run` → **12 records, 0 config errors**, generated file shows `allowed-origins: []`, `allow: []`, `secret-refs: []`

Also confirmed independently by the reviewer: the `diff` of `conduit init` output between `main` and this branch is **exactly the 3 claimed lines** — a stronger check on the `insertNode` refactor than the unit tests are.

